PR #53, opened 2021-09-10
add support for running acceptance tests with sandbox/playground tokens

displague:
This PR investigates what it would look like to support API sandbox / playground tokens:

image

These can be found in the developer app dashboard: _developer.equinix.com/user/{your id}/apps?refresh_pg_token=true_

I'm not sure if the sandbox provides the necessary state consistency to allow for the acceptance tests to pass.

If _some_ acceptance tests can pass with the playground, we could introduce levels of acceptance testing, gated by environment variables, that cover what is possible.

To test:

```
export EQUINIX_API_ENDPOINT=https://playgroundapi.equinix.com
export EQUINIX_API_TOKEN=..........

make testacc
```

---

I ran into `unsupported state format version: expected "0.1", got "0.2"` and had to address that by updating the version of the TF SDK:

https://github.com/equinix/terraform-provider-metal/issues/155#issuecomment-883265481

---

## TODO

Several other acceptance test variables will need to be explored, such as `TF_ACC_NETWORK_DEVICE_METRO`. It may not be possible to set mock values for all of these (compatible with playground). 

These can be found with `git grep TF_ACC_`.

### DIFF
--- a/equinix/provider.go
+++ b/equinix/provider.go
@@ -20,6 +20,7 @@ const (
 	endpointEnvVar      = "EQUINIX_API_ENDPOINT"
 	clientIDEnvVar      = "EQUINIX_API_CLIENTID"
 	clientSecretEnvVar  = "EQUINIX_API_CLIENTSECRET"
+	clientTokenEnvVar   = "EQUINIX_API_TOKEN"
 	clientTimeoutEnvVar = "EQUINIX_API_TIMEOUT"
 )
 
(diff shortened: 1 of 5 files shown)


equinix/provider.go
```diff
@@ -44,18 +45,25 @@ func Provider() *schema.Provider {
...
+				DefaultFunc: schema.EnvDefaultFunc(clientSecretEnvVar, nil),
+				// ValidateFunc: validation.StringIsNotEmpty,
+				Description: "API Consumer secret available under My Apps section in developer portal",
+			},
+			"token": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc(clientTokenEnvVar, nil),
+				// ValidateFunc: validation.StringIsNotEmpty,
+				Description: "API token from the developer sandbox",
```
ocobleseqx, line 66: It shall be valid also for production environment, previously retrieving a token with the endpoint https://api.equinix.com/oauth2/v1/token . I wonder if we should update /docs to explain this attribute and its possibilities

equinix/provider.go
```diff
@@ -44,18 +45,25 @@ func Provider() *schema.Provider {
...
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc(clientSecretEnvVar, nil),
+				// ValidateFunc: validation.StringIsNotEmpty,
+				Description: "API Consumer secret available under My Apps section in developer portal",
+			},
+			"token": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc(clientTokenEnvVar, nil),
+				// ValidateFunc: validation.StringIsNotEmpty,
```
ocobleseqx, line 65: I would add conflict checking to avoid confusion over which one takes precedence if both (client-pair/token) are set:

`ConflictsWith: []string{"client_id", "client_secret"},
`

displague: ConflictsWith makes sense in the scope of a provider block:
```hcl
provider "equinix" {
   client_id ="..."
   client_secret = "..."
   token = "..."
}
```

But we may also want to consider users supplying secrets via environment.
I'm not sure if we should reject `EQUINIX_TOKEN` if `EQUINIX_CLIENT_ID` or `EQUINIX_CLIENT_SECRET` are set.

Perhaps `ConflictsWith` will not restrict our ability to handle environment variables.
I'd like to compare notes with other Terraform providers to see what they've done in this situation.

ocobleseqx: As you mention, my proposal could be a problem with environment variables and with external credential files (possible alternative in the future).

After checking the official providers, none of them add restrictions at schema definition, and all conditions are managed in the Config builder function ([ex GCP Provider](https://github.com/hashicorp/terraform-provider-google/blob/f98a7c02b43c4113253a547a2d1ad213156a45a1/google/config.go#L1101)) or underlying sdk. However, they do define the order of precedence in the attribute [documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#:~:text=application%2Ddefault%20login.-,access_token,-%2D%20(Optional)%20A%20temporary)

> If you specify both with environment variables, Terraform uses the access_token instead of the credentials field

displague: I've updated the provider documentation, I think this takes care of both concerns.

ocobleseqx (approved):

PR #59, opened 2021-11-11
WAN/SSH interface selection

yvikrant:
* Added ability to select WAN/SSH interface
* Bug fix - Handled retrigger of device because ofLicense token

### DIFF
--- a/equinix/resource_network_device.go
+++ b/equinix/resource_network_device.go
@@ -50,6 +50,7 @@ var networkDeviceSchemaNames = map[string]string{
 	"InterfaceCount":      "interface_count",
 	"CoreCount":           "core_count",
 	"IsSelfManaged":       "self_managed",
+	"WanInterfaceId":      "wan_interface_id",
 	"Interfaces":          "interface",
 	"VendorConfiguration": "vendor_configuration",
 	"UserPublicKey":       "ssh_key",
@@ -90,6 +91,7 @@ var networkDeviceDescriptions = map[string]string{
 	"InterfaceCount":      "Number of network interfaces on a device. If not specified, default number for a given device type will be used",
 	"CoreCount":           "Number of CPU cores used by device",
 	"IsSelfManaged":       "Boolean value that determines device management mode: self-managed or subscription (default)",
+	"WanInterfaceId":      "device interface id picked for WAN",
 	"Interfaces":          "List of device interfaces",
 	"VendorConfiguration": "Map of vendor specific configuration parameters for a device",
 	"UserPublicKey":       "Definition of SSH key that will be provisioned on a device",
@@ -350,6 +352,14 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 			ForceNew:    true,
 			Description: networkDeviceDescriptions["IsSelfManaged"],
 		},
+		networkDeviceSchemaNames["WanInterfaceId"]: {
+			Type:         schema.TypeString,
+			Optional:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.StringIsNotEmpty,
+			Description:  networkDeviceDescriptions["WanInterfaceId"],
+		},
+
 		networkDeviceSchemaNames["Interfaces"]: {
 			Type:     schema.TypeList,
 			Computed: true,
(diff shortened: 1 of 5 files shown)


displague:
When adding new attributes, those attributes should be reflected in the docs.
https://github.com/equinix/terraform-provider-equinix/blob/master/docs/resources/network_device.md

Generally, I would also expect this new field to be added to a network_device datasource, but we don't have one of those in this provider (yet).

displague:
We should revise an existing test or add a new test related to this field: https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_network_device_test.go

equinix/resource_network_device.go
```diff
@@ -899,9 +920,6 @@ func updateNetworkDeviceResource(primary *ne.Device, secondary *ne.Device, d *sc
 	if err := d.Set(networkDeviceSchemaNames["IsBYOL"], primary.IsBYOL); err != nil {
 		return fmt.Errorf("error reading IsBYOL: %s", err)
 	}
-	if err := d.Set(networkDeviceSchemaNames["LicenseToken"], primary.LicenseToken); err != nil {
```
displague, line 902: Since `LicenseToken` is `ForceNew`, it would never be used in Update requests (`PUT`/`PATCH`).

https://github.com/equinix/terraform-provider-equinix/pull/59/files#diff-92efd33eb689af1aff178688e7f2c8db713369942a9537f78bf4bcff7cf6e33aR244-R251

We can see that it is not used in UpdateRequests here: https://github.com/equinix/terraform-provider-equinix/pull/59/files#diff-92efd33eb689af1aff178688e7f2c8db713369942a9537f78bf4bcff7cf6e33aR1118-R1134

By removing `LicenseToken` here, we are asserting that the API returned value for LicenseToken on `GET` requests should be ignored. Is that the case? Is this field no longer returned after the network device is created with the token?

yvikrant: The license token was never returned from the API. This was forcing device recreation because it was sent to empty.

yvikrant:
@displague . Updated test cases and documentation. Please validate

equinix/resource_network_device_test.go
```diff
@@ -136,7 +139,7 @@ func TestNetworkDevice_updateResourceData(t *testing.T) {
 	assert.Equal(t, ne.StringValue(inputPrimary.PackageCode), d.Get(networkDeviceSchemaNames["PackageCode"]), "PackageCode matches")
 	assert.Equal(t, ne.StringValue(inputPrimary.Version), d.Get(networkDeviceSchemaNames["Version"]), "Version matches")
 	assert.Equal(t, ne.BoolValue(inputPrimary.IsBYOL), d.Get(networkDeviceSchemaNames["IsBYOL"]), "IsBYOL matches")
-	assert.Equal(t, ne.StringValue(inputPrimary.LicenseToken), d.Get(networkDeviceSchemaNames["LicenseToken"]), "LicenseToken matches")
+	assert.Empty(t, d.Get(networkDeviceSchemaNames["LicenseToken"]), "LicenseToken is empty")
```
displague, line 142: I'm puzzled by this test - should we expect the LicenseToken to be unchanged from the creation state?

displague: These tests are unusual Terraform provider tests.  Usually Terraform tests are constructed with HCL stanzas being applied on top of each and tested after they are applied.

 I think I see why this is, we are not starting with the state used in TestNetworkDevice_createFromResourceData, we are starting from an empty (possibly invalid) state - https://github.com/equinix/terraform-provider-equinix/pull/59/files#diff-b562c6e4659cba3932687df05efa8fc8149a3412bfe2bb8a506bcb787250dc7dR124

displague (approved):

equinix/resource_network_device_test.go
yvikrant: Yes, License Token is not returned by the GET API as we don't store that information. The token value should remain the same as that value provided during the creation process.

displague: > The token value should remain the same as that value provided during the creation process.

And in this test, the initial value is `""`, right?

Issue #26, opened 2020-11-06
ECX L2 Get connection details by UUID Resource


pfeiffermj:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

Right now the equinix_ecx_l2_connection resource provides the mechanism to POST a new L2 connection and the equinix_ecx_l2_connection_accepter resource provides the ability to PATCH the L2 connection. However, what is missing to have a full integration with AWS TF is a resource to GET connection details by UUID (/ecx/v3/l2/connections/{uuid}). 

The reason this is important is in the AWS I need to pass in the **connection_id** and **vlan** attributes to **resource "aws_dx_private_virtual_interface"** to create a private VIF. These values can be pulled from the GET operation mentioned above by **"key": "awsConnectionId"** and **"zSideVlanSTag"**. 

The net result of what is illustrated below is the DX can be fully provisioned between Equinix and AWS without any kind of manual intervention or having to write custom code to call the API to get the desired behavior.

### New or Affected Resource(s)

* (new) equinix_ecx_l2_connection_details

### Potential Terraform Configuration

```tf
resource "equinix_ecx_l2_connection" "connect" {
  count               = length(var.region_list)
  name                = format("%s%s", var.name_prefix, "${element(var.region_list, count.index)}")
  profile_uuid        = var.profile_uuid
  device_uuid         = "${element(var.device_uuids, count.index)}"
  device_interface_id = var.device_interface
  speed               = var.speed
  speed_unit          = var.speed_unit
  notifications       = var.email_list
  seller_metro_code   = "${element(var.seller_metro_codes, count.index)}"
  seller_region       = "${element(var.region_list, count.index)}"
  authorization_key   = "${element(var.pairing_keys, count.index)}"
}

resource "equinix_ecx_l2_connection_accepter" "accepter" {
  // creates dependency on equinix_ecx_l2_connection
  connection_id = equinix_ecx_l2_connection.connect.uuid 
  access_key = "YYY"
  secret_key = "XXX"
}

resource "equinix_ecx_l2_connection_details" "details" {
[...]

Issue #22, opened 2020-10-29
Network Edge: awsConnectionId attribute is not available


abudilov:
### Affected resources
ecx_l2_connection

### Output

```
module.aws.aws_dx_private_virtual_interface.vif-2: Creating...
module.aws.aws_dx_private_virtual_interface.vif-1: Creating...

Error: error creating Direct Connect private virtual interface: DirectConnectClientException: ConnectionID 57ff76d2-a875-464a-b5c7-afc03c7278d7 has an invalid format.



Error: error creating Direct Connect private virtual interface: DirectConnectClientException: ConnectionID b6fc1012-888f-4888-bc92-388c03a138d0 has an invalid format
```
### Enhancement request

When creating an AWS Virtual Interface within the Direct Connect service, a "Connection ID" attribute must be used. In the absence of AWS Data Source for 'aws_dx_connection' resource, there's no way to fetch this attribute from AWS.

Equinix Terraform provider allows creating an "equinix_ecx_l2_connection" resource and relies on ECX API to create the Direct Connect connection. ECX API L2 connection object includes actionDetails list that includes the AWS connection_id:
```
"actionDetails": [
        {
            "actionType": "EQUINIX_EXECUTE_ACTION",
            "operationId": "CONFIRM_CONNECTION",
            "actionMessage": "Please provide the following credentials so that we can confirm the connection on your behalf. You also have the option to do this operation on AWS Console",
            "actionRequiredData": [
                {
                    "key": "accessKey",
                    "label": "Amazon Access Key",
                    "value": "",
                    "editable": true,
                    "validationPattern": "{RegexForString50chars}"
                },
                {
                    "key": "secretKey",
                    "label": "Amazon Secret Key",
                    "value": "",
                    "editable": true,
                    "validationPattern": "{RegexForString24chars}"
                },
                {
                    "key": "awsConnectionId",
                    "label": "AWS Hosted Connection Id",
                    "value": "dxcon-fg8fakqw",
                    "editable": false,
                    "validationPattern": ""
                }
            ]
        }
    ],
```
#### Requesting terraform provider to expose this API attribute

mikouaj:
@abudilov my proposal is to add `aws_connection_id` attribute to `equinix_ecx_l2_connection_accepter` resource output.

Example:
```
resource "equinix_ecx_l2_connection_accepter" "aws-dc" {
  connection_id = equinix_ecx_l2_connection.aws-dc.id
  access_key    = var.aws_access_key
[...]

PR #64, opened 2021-11-29
Deprecate equinix_ecx_l2_connection_accepter

### LINKED ISSUE #62: remove aws_dx_connection_confirmation resource
I've been looking into the `equinix_ecx_l2_connection_accepter` resource and from what I can tell, this resource should be removed entirely in favor of AWS provider resources that provide the same service.

[aws_dx_connection_confirmation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dx_connection_confirmation) appears to provide the same functionality as the [equinix_ecx_l2_connection_accepter resource](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/ecx_l2_connection_accepter).

By managing AWS resources within the Equinix Provider, the Equinix provider takes on AWS credential handling concerns. These concerns are entirely avoided when AWS resources are used.  Also, importing AWS packages into the provider complicates the Go dependencies of this project.

The resource should be removed with examples and documentation updated to reflect that the AWS resource should be used.  In order to remove this resource from the provider, since the provider is > v1, a v2 will be needed following a v1.x that includes a deprecation message.  See https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-data-source-or-resource-removal for more details.

_Originally posted by @displague in https://github.com/equinix/terraform-provider-equinix/issues/56#issuecomment-981235919_

displague:
This PR disabled the `equinix_ecx_l2_connection_accepter` resource with soft errors. Users are instructed via Terraform deprecation warnings, provision errors, and resource documentation to adopt the `aws_dx_connection_confirmation` resource.

From the updated  `equinix_ecx_l2_connection_accepter`  documentation:

> Use the [`aws_dx_connection_confirmation`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dx_connection_confirmation) resource from the [AWS provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs). The documentation below applies to version v1.3 and below of the Equinix provider. Later versions of the Equinix provider will return an error, requiring usage of the [AWS provider v3.62.0+](https://github.com/hashicorp/terraform-provider-aws/blob/v3.62.0/CHANGELOG.md#3620-october-08-2021) feature.

The issue that led to the resource that provides this feature in the AWS provider is https://github.com/hashicorp/terraform-provider-aws/issues/9366.

Fixes #62


displague:
An additional motivator for this PR is that the Amazon SDK adds Go dependencies that make this provider cumbersome to import in other providers (Crossplane, Pulumi).

examples/connectivity/aws-connection/main.tf
```diff
@@ -31,14 +31,18 @@ resource "equinix_ecx_l2_connection" "example" {
...
 }
 
-resource "equinix_ecx_l2_connection_accepter" "example" {
-  connection_id = equinix_ecx_l2_connection.example.id
-  access_key    = var.aws_access_key
-  secret_key    = var.aws_secret_key
+resource "aws_dx_connection" "example" {
+  name      = "tf-aws-dot1q"
+  bandwidth = "50Mbps"
+  location  = var.aws_location_code
```
displague, line 37: It would be nice if there was a way to match the location code with an Equinix location for the purposes of these examples. I could not find one so I took the path of least resistance and suggested that the user define this value in a variable.

displague: I'm also creating the connection in this example.  It seems we have the option to instead use an existing connection via https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/dx_connection. For the purposes of the example, I think either approach is fine. (assuming this code works - I was not able to verify this. I'm hoping @ocobleseqx  may be able to.)

examples/connectivity/aws-connection/main.tf
```diff
@@ -31,14 +31,18 @@ resource "equinix_ecx_l2_connection" "example" {
   authorization_key = var.aws_account_id
 }
 
-resource "equinix_ecx_l2_connection_accepter" "example" {
-  connection_id = equinix_ecx_l2_connection.example.id
-  access_key    = var.aws_access_key
-  secret_key    = var.aws_secret_key
+resource "aws_dx_connection" "example" {
```
ocobleseqx, line 34: 'equinix_ecx_l2_connection' resource creates the connection in the AWS side on your behalf and therefore the 'aws_dx_connection' resource is not required

displague: If we don't need an dx_ec2_connection resource, how should we get the connection I'd that AWS needs in their APIs and Terraform resources?

examples/connectivity/aws-connection/main.tf
ocobleseqx: Since the connection in AWS is configured through the 'equinix_exc_l2_connection' resource the connection id should be referenceable in that resource. This was already discussed here #22 .

AWS dx_connection datasource could be an alternative but there may be an issue, AWS allows creating multiple direct connect connections with the same name but the aws_dx_connection datasource doesn't allow more filters than [name](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/directconnect/connection_data_source.go#L52)

ocobleseqx (changes requested):

examples/connectivity/aws-connection/main.tf
displague: There is a convention to use a singular name or a plural name in data sources depending on whether the resource name returns a map or array of results or a single result. When a resource is expected to return a single result, the provider will typically throw an error if more than one results match the selection criteria.

I don't know if this AWS follows that convention, but I think if this approach works at all, it is sufficient for the example. If it helps to avoid an incorrect example, we could simplify further with `var.aws_connection_id` instead of creating or looking up the AWS DX Connection.

displague: From #22
> When creating an AWS Virtual Interface within the Direct Connect service, a "Connection ID" attribute must be used. In the absence of AWS Data Source for 'aws_dx_connection' resource, there's no way to fetch this attribute from AWS.

We now have https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/dx_connection available.

> Equinix Terraform provider allows creating an "equinix_ecx_l2_connection" resource and relies on ECX API to create the Direct Connect connection. ECX API L2 connection object includes actionDetails list that includes the AWS connection_id:

It seems we may need to revisit https://github.com/equinix/terraform-provider-equinix/issues/26.  

@pfeiffermj proposed `connection_id     = 
  equinix_ecx_l2_connection_details.details.actionDetails.actionRequiredData.ELEMENT.awsConnectionId.value` which @abudilov also suggested in https://github.com/equinix/terraform-provider-equinix/issues/22.

Issues 22 and 26 discuss an `actionDetails` element of the ECX API which contains a list of maps with field values to populate AWS credentials and retrieve AWS connection details.  While improved access to this attribute (Terraform providers should match the upstream API), this API pattern has been revoked by the successor API, ECX APIv3:

>  V3 removes the need for a service provider to approve a deleted connection and immediately deletes the connection and stops billing when this action is performed by a user

More details are provided on https://developer.equinix.com/docs/ecx-v3-apis-migration and https://developer.equinix.com/docs/ecx-connect-aws.

> As of January 5, 2020, Equinix stopped supporting Equinix Fabric V2 APIs, and after July 5, 2020, all previous versions of Equinix Fabric APIs have been deprecated.

It seems we may require a more comprehensive overhaul.

For now, I believe it would be sufficient to find a way to discover the AWS Connection ID.

ocobleseqx (approved):

ocobleseqx (dismissed):

ocobleseqx (approved):

PR #71, opened 2021-12-02
Fix connectivity example to establish an azure connection

### LINKED ISSUE #70: Incorrect Azure connection example - Wrong resources to retrieve required authorization key
The example to create a connection to Azure is using the resource 'azurerm_express_route_circuit_authorization' to pick up the service key required in the 'equinix_ecx_l2_connection' resource, but that azurerm resource is intended for a different use case. It must take the service key from the 'azurerm_express_route_circuit' resource

https://github.com/equinix/terraform-provider-equinix/blob/16159de36ef1c120c591ffd98ad3be0fb85b985f/examples/connectivity/azure-connection/main.tf#L57

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.

### Affected Resource(s)

* equinix_XXXXX

### Terraform Configuration Files

```tf
# Copy-paste your Terraform configurations here.
#
# For large Terraform configs, please use a service like Dropbox and share a link to the ZIP file.
#
# If reproducing the bug involves modifying the config file (e.g., apply a config,
# change a value, apply the config again, see the bug), then please include both:
# * the version of the config before the change, and
# * the version of the config after the change.
```

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

ocobleseqx:
This fixes #70

### DIFF
--- a/examples/connectivity/azure-connection/main.tf
+++ b/examples/connectivity/azure-connection/main.tf
@@ -30,7 +30,7 @@ resource "azurerm_express_route_circuit" "demo" {
   resource_group_name   = azurerm_resource_group.demo.name
   location              = azurerm_resource_group.demo.location
   service_provider_name = "Equinix"
-  peering_location      = "Silicon Valley"
+  peering_location      = var.azure_peering_location
   bandwidth_in_mbps     = 50
   sku {
     tier   = "Premium"
@@ -39,11 +39,6 @@ resource "azurerm_express_route_circuit" "demo" {
   allow_classic_operations = false
 }
 
-resource "azurerm_express_route_circuit_authorization" "demo" {
-  name                       = "TFDemoExpressRouteAuth"
-  express_route_circuit_name = azurerm_express_route_circuit.demo.name
-  resource_group_name        = azurerm_resource_group.demo.name
-}
 
 resource "equinix_ecx_l2_connection" "azure-dot1q-pub" {
   name              = "tf-azure-dot1q-pub-pri"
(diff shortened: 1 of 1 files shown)


displague:
I've looked at this and I find it hard to reason about without access to the underlying resources (nor E2E testing).  Could you share a demo of this with me?

The existing variable names, attributes, and resources make sense to me. I'm not sure how the authorization_key would be replaced with a service_key.

* https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/express_route_circuit_authorization#attributes-reference
* https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/express_route_circuit

Alternatively, the following would convince me:
* terraform output captures showing that this was successfully provisioned
* screenshots showing that the two UIs are in a good state
* screenshot or output showing a working network link

These screenshots that you add here, along with supportive text, could be added to the README.md later.

ocobleseqx:
The circuit authorization is an ExpressRoute (ER) feature to connect virtual networks that are in separate tenants/enrollments to a single ExpressRoute circuit not in the same tenant/enrollment - https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-linkvnet-arm#circuit-owner-operations

To create a new ER circuit you can see the process in the ER circuit properties section - https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#view-the-circuits-and-properties

> Refer to the service key for your circuit and provide it to the service provider to complete the provisioning process. The service key is specific to your circuit

The confusion comes from that in Equinix Fabric "authorization key" is used as a generic name for all service providers and coincidentally in there is an authorization key for a different matter in Azure. 

To demonstrate that it works as expected you can compare the status in the attached screenshot with the described here as the desired status after the provider (Equinix) enables the connection on your behalf -  https://docs.microsoft.com/en-us/azure/expressroute/expressroute-howto-circuit-portal-resource-manager#send-the-service-key-to-your-connectivity-provider-for-provisioning

> The circuit changes to the following state when the connectivity provider is currently enabling it for you:
> 
> Provider status: Provisioning
> Circuit status: Enabled
> 
> To use the ExpressRoute circuit, it must be in the following state:
> 
> Provider status: Provisioned
> Circuit status: Enabled

image

image

image

PR #105, opened 2022-02-28
NFV-18043: add cluster availability in device creation API

rling-equinix:
Cluster is introduced in device creation API. Some new device and device license pending states are introduced along with it.
Device provisioning time has been updated to 90 minutes considering the cluster scenarios.
Go module has been cleaned up by go mod tidy and ne-go has been upgraded to v1.4.0
reference: https://developer.equinix.com/docs/launch-cluster-device
This terraform availability is part of April release scope for NE.

### DIFF
--- a/docs/resources/equinix_network_device.md
+++ b/docs/resources/equinix_network_device.md
@@ -137,7 +137,7 @@ specified, default number for a given device type will be used
 * `wan_interafce_id` - (Optional) Specify the WAN/SSH interface id. If not
   specified, default WAN/SSH interface for a given device type will be used
 * `vendor_configuration` - (Optional) Map of vendor specific configuration parameters
-for a device
+for a device (controller1, activationKey, managementType, siteId, systemIpAddress)
 * `ssh-key` - (Optional) Definition of SSH key that will be provisioned
 on a device (max one key)
 * `secondary_device` - (Optional) Definition of secondary device for redundant
@@ -160,8 +160,8 @@ secondary device
 will receive notifications about secondary device
 * `additional_bandwidth` - (Optional) Additional Internet
 bandwidth, in Mbps, for a secondary device
-* `vendor_configuration` - (Optional) map of vendor specific
-configuration parameters for a secondary device
+* `vendor_configuration` - (Optional) Map of vendor specific
+configuration parameters for a secondary device (controller1, activationKey, managementType, siteId, systemIpAddress)
 * `acl_template_id` - Identifier of an ACL template that will
 be applied on a secondary device
 * `ssh-key` - (Optional) up to one definition of SSH key that will be provisioned
--- a/docs/resources/equinix_network_device.md
+++ b/docs/resources/equinix_network_device.md
@@ -76,6 +76,7 @@ resource "equinix_network_device" "panw-cluster" {
   term_length     = 6
   account_number  = data.equinix_network_account.sv.number
   version         = "10.1.3"
+  interface_count = 10
   core_count      = 2
   ssh_key {
     username = "test"
--- a/docs/resources/equinix_network_device.md
+++ b/docs/resources/equinix_network_device.md
@@ -58,6 +58,47 @@ resource "equinix_network_device" "csr1000v-ha" {
   }
 }
 ```
+```hcl
+# Create self configured PANW cluster with BYOL license
+
+data "equinix_network_account" "sv" {
+  metro_code = "SV"
+}
+
+resource "equinix_network_device" "panw-cluster" {
+  name            = "tf-panw"
+  metro_code      = data.equinix_network_account.sv.metro_code
+  type_code       = "PA-VM"
+  self_managed    = true
+  byol            = true
+  package_code    = "VM100"
+  notifications   = ["john@equinix.com", "marry@equinix.com", "fred@equinix.com"]
+  term_length     = 6
+  account_number  = data.equinix_network_account.sv.number
+  version         = "10.1.3"
+  core_count      = 2
+  ssh_key {
+    username = "test"
+    key_name = "test-key"
+  }
+  acl_template_id = "0bff6e05-f0e7-44cd-804a-25b92b835f8b"
+  cluster_details {
+    cluster_name    = "tf-panw-cluster"
+    node0 {
+      vendor_configuration {
+        hostname = "panw-node0"
+      }
+      license_token = "licenseToken"
+    }
+    node1 {
+      vendor_configuration {
+        hostname = "panw-node1"
+      }
+      license_token = "licenseToken"
+    }
+  }
+}
+```
 
 ## Argument Reference
 
@@ -94,12 +135,13 @@ that will be allocated to the device (in addition to default 15Mbps)
 specified, default number for a given device type will be used
 * `wan_interafce_id` - (Optional) Specify the WAN/SSH interface id. If not
   specified, default WAN/SSH interface for a given device type will be used
-* `vendor_configuration` - (Optional) map of vendor specific configuration parameters
+* `vendor_configuration` - (Optional) Map of vendor specific configuration parameters
 for a device
-* `ssh-key` - (Optional) definition of SSH key that will be provisioned
+* `ssh-key` - (Optional) Definition of SSH key that will be provisioned
 on a device (max one key)
 * `secondary_device` - (Optional) Definition of secondary device for redundant
 device configurations
+* `cluster_details` - (Optional) An object that has the cluster details
 
 The `secondary_device` block supports the following arguments:
 
@@ -129,6 +171,33 @@ The `ssh_key` block supports the following arguments:
 * `username` - (Required) username associated with given key
 * `name` - (Required) reference by name to previously provisioned public SSH key
 
+The `cluster_details` block supports the following arguments:
+
+* `cluster_name` - (Required) The name of the cluster device
+* `node0` - (Required) An object that has node0 details
+* `node1` - (Required) An object that has node1 details
+
+The `node0` block supports the following arguments:
+
+* `vendor_configuration` - (Optional) An object that has fields relevant to the vendor of the cluster device
+* `license_file_id` - (Optional) License file id. This is necessary for Fortinet and Juniper clusters
+* `license_token` - (Optional) License token. This is necessary for Palo Alto clusters
+
+The `node1` block supports the following arguments:
+
+* `vendor_configuration` - (Optional) An object that has fields relevant to the vendor of the cluster device
+* `license_file_id` - (Optional) License file id. This is necessary for Fortinet and Juniper clusters
+* `license_token` - (Optional) License token. This is necessary for Palo Alto clusters
+
+The `vendor_configuration` block supports the following arguments:
+
+* `hostname` - (Optional) Hostname. This is necessary for Palo Alto, Juniper, and Fortinet clusters
+* `admin_password` - (Optional) The administrative password of the device. You can use it to log in to the console. This field is not available for all device types
+* `controller1` - (Optional) System IP Address. Mandatory for the Fortinet SDWAN cluster device
+* `activation_key` - (Optional) Activation key. This is required for Velocloud clusters
+* `controller_fqdn` - (Optional) Controller fqdn. This is required for Velocloud clusters
+* `root_password` - (Optional) The CLI password of the device. This field is relevant only for the Velocloud SDWAN cluster
+
 ## Attributes Reference
 
 * `uuid` - Device unique identifier
@@ -137,6 +206,8 @@ The `ssh_key` block supports the following arguments:
   * PROVISIONING
   * WAITING_FOR_PRIMARY
   * WAITING_FOR_SECONDARY
+  * WAITING_FOR_REPLICA_CLUSTER_NODES 
+  * CLUSTER_SETUP_IN_PROGRESS 
   * FAILED
   * PROVISIONED
   * DEPROVISIONING
@@ -145,6 +216,7 @@ The `ssh_key` block supports the following arguments:
   * APPLYING_LICENSE
   * REGISTERED
   * APPLIED
+  * WAITING_FOR_CLUSTER_SETUP
   * REGISTRATION_FAILED
 * `license_file_id` - Unique identifier of applied license file
 * `ibx` - Device location Equinix Business Exchange name
@@ -166,15 +238,17 @@ primary or secondary
   * `interface.#.type` - interface type
 * `asn` - Autonomous system number
 * `zone_code` - Device location zone code
+* `cluster_id` - The id of the cluster
+* `num_of_nodes` - The number of nodes in the cluster
 
 ## Timeouts
 
 This resource provides the following [Timeouts configuration](https://www.terraform.io/language/resources/syntax#operation-timeouts)
 options:
 
-* create - Default is 60 minutes
-* update - Default is 10 minutes
-* delete - Default is 10 minutes
+* create - Default is 90 minutes
+* update - Default is 30 minutes
+* delete - Default is 30 minutes
 
 ## Import
 
--- a/equinix/resource_network_device.go
+++ b/equinix/resource_network_device.go
@@ -137,41 +137,35 @@ var neDeviceUserKeyDescriptions = map[string]string{
 }
 
 var neDeviceClusterSchemaNames = map[string]string{
-	"ClusterId":          "cluster_id",
-	"ClusterName":        "cluster_name",
-	"NumOfNodes":         "num_of_nodes",
-	"ClusterNodeDetails": "cluster_node_details",
-	"Nodes":              "nodes",
+	"ClusterId":   "cluster_id",
+	"ClusterName": "cluster_name",
+	"NumOfNodes":  "num_of_nodes",
+	"Node0":       "node0",
+	"Node1":       "node1",
 }
 
 var neDeviceClusterDescriptions = map[string]string{
-	"ClusterId":          "The id of the cluster",
-	"ClusterName":        "The name of the cluster device",
-	"NumOfNodes":         "The number of nodes in the cluster",
-	"ClusterNodeDetails": "An object that has the cluster node details",
-	"Nodes":              "An array of multiple node objects",
+	"ClusterId":   "The id of the cluster",
+	"ClusterName": "The name of the cluster device",
+	"NumOfNodes":  "The number of nodes in the cluster",
+	"Node0":       "An object that has node0 details",
+	"Node1":       "An object that has node1 details",
 }
 
 var neDeviceClusterNodeSchemaNames = map[string]string{
-	"Node0":               "node0",
-	"Node1":               "node1",
 	"VendorConfiguration": "vendor_configuration",
 	"LicenseFileId":       "license_file_id",
 	"LicenseToken":        "license_token",
 	"UUID":                "uuid",
 	"Name":                "name",
-	"Node":                "node",
 }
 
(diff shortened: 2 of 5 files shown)


equinix/resource_network_device.go
```diff
@@ -635,6 +743,70 @@ func createNetworkDeviceUserKeySchema() map[string]*schema.Schema {
...
+		neDeviceClusterSchemaNames["AdminPassword"]: {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Sensitive:   true,
+			Description: neDeviceClusterDescriptions["AdminPassword"],
+		},
+		neDeviceClusterSchemaNames["VendorConfiguration"]: {
+			Type:     schema.TypeMap,
+			Computed: true,
+			Elem: &schema.Schema{
```
ocobleseqx, line 801: https://developer.equinix.com/docs/launch-cluster-device
> {vendorConfig{hostName, activationKey, controllerFqdn, rootPassword, adminPassword, controller1}

It seems there is a predefined list of valid parameters. It would be better to use a TypeSet so that users know what values they can include

equinix/resource_network_device.go
```diff
@@ -635,6 +743,70 @@ func createNetworkDeviceUserKeySchema() map[string]*schema.Schema {
...
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: neDeviceClusterDescriptions["Name"],
+		},
+		neDeviceClusterSchemaNames["Node"]: {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: neDeviceClusterDescriptions["Node"],
+		},
+		neDeviceClusterSchemaNames["AdminPassword"]: {
```
ocobleseqx, line 792: Isn't `AdminPassword` part of the `VendorConfiguration` block?

go.mod
```diff
@@ -18,63 +18,10 @@ require (
 )
 
 require (
```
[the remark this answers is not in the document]
displague: I was screen-sharing with @rling-equinix when this change was made. I'm not sure why all of these indirect were removed. I'm suspecting the version of go used between CI (1.17.7) and the local install.

rling-equinix:
> I added some comments that should be addressed before merging.
> 
> Can you share an example of how a resource for a cluster definition should look like? I wonder if we should add some "ConflictsWith" validations, I found some parameters like hostName/licenseFileId/LicenseToken that now can be defined at different levels.
> 
> It would also necessary update the documentation to describe this new option but I can help with that if you share an example.

This is one example resource for panw cluster.
resource "equinix_network_device" "panw-cluster" {
  name            = "cluster-name"
  metro_code      = "SV"
  type_code       = "PA-VM"
  self_managed    = true
  byol            = true
  package_code    = "VM100"
  notifications   = ["test@eq.com"]
  term_length     = 1
  account_number  = "some_number"
  version         = "10.1.3"
  interface_count = 10
  core_count      = 2
  ssh_key {
    username = "test"
    key_name = "test_key"
  }
  cluster_details {
    cluster_name    = "test-panw-cluster-name"
    cluster_node_details {
      node0 {
        vendor_configuration = {
          hostname = "panw-host0"
        }
        license_token = "some_token"
      }
      node1 {
        vendor_configuration = {
          hostname = "panw-host1"
        }
        license_token = "some_token"
      }
    }
  }
}

equinix/resource_network_device.go
```diff
@@ -132,6 +136,40 @@ var neDeviceUserKeyDescriptions = map[string]string{
 	"KeyName":  "Reference by name to previously provisioned public SSH key",
 }
 
+var neDeviceClusterSchemaNames = map[string]string{
```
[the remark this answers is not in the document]
rling-equinix: fixed

equinix/resource_network_device.go
```diff
@@ -132,6 +136,40 @@ var neDeviceUserKeyDescriptions = map[string]string{
...
+	"LicenseToken":        "license_token",
+	"Nodes":               "nodes",
+	"UUID":                "uuid",
+	"Name":                "name",
+	"Node":                "node",
+	"AdminPassword":       "admin_password",
+	"VendorConfiguration": "vendor_configuration",
+}
+
+var neDeviceClusterDescriptions = map[string]string{
```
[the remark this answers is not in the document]
rling-equinix: fixed

equinix/resource_network_device.go
```diff
@@ -570,6 +614,70 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
...
+						Computed:    true,
+						Description: neDeviceClusterDescriptions["ClusterId"],
+					},
+					neDeviceClusterSchemaNames["ClusterName"]: {
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringLenBetween(3, 50),
+						Description:  neDeviceClusterDescriptions["ClusterName"],
+					},
+					neDeviceClusterSchemaNames["NumOfNodes"]: {
```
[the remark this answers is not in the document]
rling-equinix: I remove this field from the request. It should be computed in the response.

go.mod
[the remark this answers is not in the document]
rling-equinix: I updated my local install to go 1.17.7 and that go module issues have been fixed.

equinix/resource_network_device.go
rling-equinix: This is an existing field under device schema as well and for now we won't change it because it's already consumed by the customer.

equinix/resource_network_device.go
rling-equinix: Thanks for pointing that out. I have removed that from our terraform schema. It looks like we have API duplication for this field. Will fix that in parallel.

equinix/resource_network_device.go
```diff
@@ -570,6 +618,69 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 				},
 			},
 		},
+		networkDeviceSchemaNames["ClusterDetails"]: {
+			Type:        schema.TypeList,
+			Optional:    true,
+			ForceNew:    true,
```
displague, line 624: nitpick: I don't recall if ForceNew here will allow a clustername change without destroying the device first. We can introduce this new parameter with ForceNew and remove that flag later if it is a problem.

displague: There are no "update" API  methods for cluster details, so `ForceNew` is in fact appropriate here.

equinix/resource_network_device.go
```diff
@@ -570,6 +618,69 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
...
+						MaxItems:    1,
+						Description: neDeviceClusterDescriptions["ClusterNodeDetails"],
+						Elem: &schema.Resource{
+							Schema: map[string]*schema.Schema{
+								neDeviceClusterNodeSchemaNames["Node0"]: {
+									Type:     schema.TypeList,
+									Required: true,
+									MaxItems: 1,
+									Elem: &schema.Resource{
+										Schema: createClusterNodeDetailRequestSchema(),
```
displague, line 657: Let's merge the `ClusterNodeDetails ` and `Nodes` into one object.

The `createClusterNodeDetailRequestSchema` and `createNetworkDeviceClusterNodeSchema` schemas would be combined.

This will have the advantage of the vendor config block being defined and accessed through one attribute. 

I understand why we implemented this with two fields in this PR initially. We generally want to have Terraform match the API format. In this case, the request and response offer the same object with a different format, one as a named map entity, and the other as an array member. We can quell this difference for a simpler Terraform interface.

displague: The net result would be:
```hcl
resource "equinix_network_device" "foo" {
   cluster_details {
     cluster_node_details {
        node0 { ... }
        node1 { ... }
     }
   }
}
```

Users would access the hostname as:  `equinix_network_device.foo.cluster_details.cluster_node_details[0].node0[0].vendor_config.hostname`

Perhaps we should use the response key (`node`) instead of the request key (`cluster_node_details`).

equinix/resource_network_device.go
displague: I think the change to the user is something that can be introduced in a semantic version "minor" release. 

While this is technically a breaking change, the user fix is simple (remove the `=` sign). 

@rling-equinix points out that the access pattern will also change, the map would be accessed with an index: `vendor_config.field` becomes `vendor_config[0].field` or `vendor_config.0.field`. 

The user benefit is that the structure will be validated rather than be an anonymous list of strings. The contents of the fields in the structure can also be validated and strictly typed. We can also mark the sensitive vendor configs as such.

displague: https://www.terraform.io/plugin/sdkv2/best-practices/versioning#example-minor-number-increments

> Increasing the minor number suggests that new features have been added but that existing functionality remains broadly compatible.

However:
> Example Major Number Increments: ...
> * Changing attribute type where the new type is functionally incompatible (including but not limited to changing TypeSet to TypeList and TypeList to TypeSet)

Perhaps this change should be deferred on the EXISTING parameter. 

For this PR, we could introduce this as a list of schema map.  While the two identically named parameters would have a slightly different interface, we'd get the benefits described above.   What are your thoughts, @rling-equinix @ocobleseqx?

equinix/resource_network_device.go
```diff
@@ -991,6 +1169,15 @@ func updateNetworkDeviceResource(primary *ne.Device, secondary *ne.Device, d *sc
 			return fmt.Errorf("error reading Secondary: %s", err)
 		}
 	}
+	if primary.ClusterDetails != nil {
+		if v, ok := d.GetOk(networkDeviceSchemaNames["ClusterDetails"]); ok {
+			clusterDetailsFromSchema := expandNetworkDeviceClusterDetails(v.([]interface{}))
+			primary.ClusterDetails.ClusterNodeDetails = clusterDetailsFromSchema.ClusterNodeDetails
+		}
+		if err := d.Set(networkDeviceSchemaNames["ClusterDetails"], flattenNetworkDeviceClusterDetails(primary.ClusterDetails)); err != nil {
```
displague, line 1177: To combine the Node and ClusterNodeDetails, we'll have to merge `primary.ClusterDetails` and the `nodes` response fields.

displague (commented):
In the interest of getting these changes into some early user testing, I'd be willing to merge this and provide an alpha release. But - we would need to address the outstanding schema change requests (if they hold up to further discussion) before cutting the next release. That could potentially block other releases unless we revert this PR, which I wouldn't want to do.

displague (changes requested):
I'd really like to see an E2E test added which validates that the fields that have been added are accessed.

Echoing @ocobleseqx , whenever we change or add fields or resources, we have to update the docs.  This is manual today. There is hope that we can automate this in the future using the `Description` fields. 

See these issues for more info on generating docs in the future:

https://github.com/equinix/terraform-provider-metal/issues/13 and
https://github.com/equinix/terraform-provider-metal/pull/132

equinix/resource_network_device.go
displague: After some exploration and discussion with @rling-equinix on what the merged request and response HCL object should look like, we've arrived at a few different approaches that we could take here that would simplify the structure and access pattern:

```hcl
resource "equinix_network_device" "foo" {
    cluster_detail {
        cluster_name = "foo"

        # flatten the list of cluster_node_details, cluster_node_details only holds nodes, so hoist the nodes to the higher level
        # also considered naming them primary and secondary but this would be confused with the existing HA fields.
        node0 {} 
        node1 {}
        # user can access the COMPUTED node attributes with:
        # equinix_network_device.foo.cluster_detail.0.node0.0.name

        # Nodes with node elements
        nodes { # or name this cluster_node_details
            node0 {}
            node1 {}
        }
        # user can access the COMPUTED node attributes with:
        # equinix_network_device.foo.cluster_detail.0.nodes.0.node0.0.name

        # Nodes as array (singular name, order matters)
        node {
            vendor_config {}
        }

        node {
            vendor_config {}
        }
        # user can access the COMPUTED node attributes with:
        # equinix_network_device.foo.cluster_detail.0.nodes.0.name
    }
}
```

ocobleseqx: Has there been any agreement here? Will it be merged? @rling-equinix

ocobleseqx: After discussion with @rling-equinix @displague, we find in the option below the best solution to ensure that users always define both required nodes.

>         # flatten the list of cluster_node_details, cluster_node_details only holds nodes, so hoist the nodes to the higher level
>         # also considered naming them primary and secondary but this would be confused with the existing HA fields.
>         node0 {} 
>         node1 {}

equinix/resource_network_device.go
```diff
@@ -371,6 +431,7 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 		networkDeviceSchemaNames["VendorConfiguration"]: {
 			Type:     schema.TypeMap,
 			Optional: true,
+			Computed: true,
```
displague, line 434: This was added to accommodate additional vendor configuration details that are included in the API response.

displague, line 434: Noting that this was added to accommodate additional vendor configuration details that are included in the API response.

equinix/resource_network_device.go
```diff
@@ -539,6 +600,7 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 					networkDeviceSchemaNames["VendorConfiguration"]: {
 						Type:     schema.TypeMap,
 						Optional: true,
+						Computed: true,
```
displague, line 603: Let's add a list of the map keys to `networkDeviceDescriptions["VendorConfiguration"]` since users may be confused that these map keys are not identical to the new node schema names.

We will address that inconsistency in the future, but that will be a breaking change.  As discussed, we will defer that until after this PR and feature are released.

equinix/resource_network_device_acc_test.go
```diff
@@ -823,6 +826,64 @@ func TestAccNetworkDevice_CGENIX_HA_Self_BYOL(t *testing.T) {
...
+					testAccNeDeviceAttributes(&primary, context),
+					testAccNeDeviceStatusAttributes(&primary, ne.DeviceStateProvisioned, ne.DeviceLicenseStateApplied),
+					testAccNeDeviceClusterAttributes(deviceResourceName),
+					testAccNeDeviceClusterNodeAttributes(primary.ClusterDetails, context),
+					testAccNetworkACLTemplateExists(clusterAclResourceName, &primaryAcl),
+					resource.TestCheckResourceAttrSet(deviceResourceName, "acl_template_id"),
+					testAccNeDeviceACL(clusterAclResourceName, &primary),
+				),
+			},
+		},
```
displague, line 883: Let's add an Import step.

It should look like this:

```suggestion
                        {
                                ResourceName:      deviceResourceName,
                                ImportState:       true,
                                ImportStateVerify: true,
                        },
		},
```

displague (changes requested):
This looks great! 🚀
I've left a few small notes, no major blockers.

equinix/resource_network_device_acc_test.go
displague: Here's some background on Import - https://www.terraform.io/plugin/sdkv2/resources/import

equinix/resource_network_device.go
rling-equinix: We can definitely change the schema of vendorConfig for cluster, however, for top level vendorConfig under device shema, let's defer that changes.

equinix/resource_network_device_acc_test.go
```diff
@@ -1162,7 +1280,7 @@ resource "equinix_network_device" "%{device-resourceName}" {
 	}
 	if _, ok := ctx["acl-resourceName"]; ok {
 		config += nprintf(`
-  acl_template_id       = equinix_network_acl_template.%{acl-resourceName}.id`, ctx)
```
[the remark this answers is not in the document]
rling-equinix: Fixed.

equinix/resource_network_device_acc_test.go
```diff
@@ -1298,6 +1416,97 @@ resource "equinix_network_device" "%{device-resourceName}" {
...
+    cluster_name                 = "%{device-cluster_name}"`, ctx)
+		config += `
+	node0 {`
+		if _, ok := ctx["device-node0_license_file_id"]; ok {
+			config += nprintf(`
+	  license_file_id   = "%{device-node0_license_file_id}"`, ctx)
+		}
+		if _, ok := ctx["device-node0_license_token"]; ok {
+			config += nprintf(`
+      license_token     = "%{device-node0_license_token}"`, ctx)
```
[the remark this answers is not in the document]
rling-equinix: Fixed.

equinix/resource_network_device_acc_test.go
rling-equinix: Fixed.

docs/resources/equinix_network_device.md
```diff
@@ -94,12 +136,13 @@ that will be allocated to the device (in addition to default 15Mbps)
 specified, default number for a given device type will be used
 * `wan_interafce_id` - (Optional) Specify the WAN/SSH interface id. If not
   specified, default WAN/SSH interface for a given device type will be used
-* `vendor_configuration` - (Optional) map of vendor specific configuration parameters
+* `vendor_configuration` - (Optional) Map of vendor specific configuration parameters
```
displague, line 139:
```suggestion
* `vendor_configuration` - (Optional) Map of vendor specific configuration parameters  (controller1, activationKey, managementType, siteId, systemIpAddress)
```

equinix/resource_network_device.go
```diff
@@ -159,8 +213,14 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 			Description: networkDeviceDescriptions["UUID"],
 		},
 		networkDeviceSchemaNames["Name"]: {
-			Type:         schema.TypeString,
-			Required:     true,
+			Type:     schema.TypeString,
+			Required: true,
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
displague, line 218: Help me understand this. 

Is the remote adding `-Node0` (always 0?) to the user-supplied name? Was this a one-time change or is this applied on every user-requested name change?

Can the name be synced to what the user requested? If so, why wouldn't we update the name instead of suppressing the change?

rling-equinix: > Is the remote adding `-Node0` (always 0?) to the user-supplied name? Was this a one-time change or is this applied on every user-requested name change?

Since we use Node 0 device to manage the cluster, name field will always be populated with user-supplied name + "-Node0" in this case. It's a one time change for the name.  

> Can the name be synced to what the user requested? If so, why wouldn't we update the name instead of suppressing the change?

Noticed that, name is a required field here in the request. Also, each time they provision the cluster, a name change will be given for

displague: Ok. Per our conversation, we can follow-up with a change that will introduce:
* `name_prefix` would be added at the `equinix_network_edge_device` level, required with "cluster_details". This field will have to be set on Ready, by trimming, for the benefit of imports.
* `name` would be computed, conflicts with "name_prefix", optional, exactlyoneof "name" or "name_prefix" 

What's interesting about this is that a user can specify `name_prefix` under all conditions.

We may also want to move the `name_prefix` field into `cluster_details`. If we do that, we may want to use a different field name, like `device_name_prefix`.

displague: Issue opened here https://github.com/equinix/terraform-provider-equinix/issues/110

displague (approved):

ocobleseqx (approved):
Excellent work @rling-equinix! 💯  Thanks for your contribution

PR #111, opened 2022-03-18
support for equinix_ecx_l2_connection > secondary_connection > service_token

ocobleseqx:
secondaryServiceToken is required for redundant connection. At this moment a service token can only be issued for a single port per connection, and unlike directly specifying an Equinix port, with service token it is not possible to reuse same one for both primary and secondary.

- Updated ecx-go version to [v2.2.0](https://github.com/equinix/ecx-go/releases/tag/v2.2.0) to support secondaryServiceToken
- Updated mock test to check secondary_connection.0.service_token attribute 
- Updated docs


ocobleseqx:
I need to attach an example and a succeeded terraform apply result

ocobleseqx:
Example of a successful apply result using two service token to establish an interconnection to Azure

image

image

displague (approved):

displague (commented):
The PR description offers insights that are not present in the README. 

Perhaps this could be used to improve the Secondary Connection heading in the connection doc.  I noticed that we don't link out to any of the product documentation for the Fabric features. I find links to the backing API and product guides helpful when reviewing Terraform resource documentation.

A guide on Interconnection could discuss this topic in greater depth and could link to the various modules that you've made available on Equinix-labs and Equinix repos.

PR #115, opened 2022-03-25
NFV-18247: add mgmt acl in device creation and update flow

rling-equinix:
The pr will allow the customer to add mgmt acl in the device creation flow. It will also allow them updating the mgmt acl for the device in place. 

This PR is dependent on ne-go changes 
https://github.com/equinix/ne-go/pull/12

### DIFF
--- a/equinix/resource_network_device.go
+++ b/equinix/resource_network_device.go
@@ -1476,21 +1476,25 @@ func fillNetworkDeviceUpdateRequest(updateReq ne.DeviceUpdateRequest, changes ma
 
 func getNetworkDeviceStateChangeConfigs(c ne.Client, deviceID string, timeout time.Duration, changes map[string]interface{}) []*resource.StateChangeConf {
 	configs := make([]*resource.StateChangeConf, 0, len(changes))
-	for change, changeValue := range changes {
-		switch change {
-		case networkDeviceSchemaNames["ACLTemplateUUID"], networkDeviceSchemaNames["MgmtAclTemplateUuid"]:
-			aclTempID, ok := changeValue.(string)
-			if !ok || aclTempID == "" {
-				break
-			}
+	if changeValue, found := changes[networkDeviceSchemaNames["ACLTemplateUUID"]]; found {
+		aclTemplateUuid, ok := changeValue.(string)
+		if ok && aclTemplateUuid != "" {
 			configs = append(configs,
 				createNetworkDeviceACLStatusWaitConfiguration(c.GetDeviceACLDetails, deviceID, 1*time.Second, timeout),
 			)
-		case networkDeviceSchemaNames["AdditionalBandwidth"]:
+		}
+	} else if changeValue, found := changes[networkDeviceSchemaNames["MgmtAclTemplateUuid"]]; found {
+		mgmtAclTemplateUuid, ok := changeValue.(string)
+		if ok && mgmtAclTemplateUuid != "" {
 			configs = append(configs,
-				createNetworkDeviceAdditionalBandwidthStatusWaitConfiguration(c.GetDeviceAdditionalBandwidthDetails, deviceID, 1*time.Second, timeout),
+				createNetworkDeviceACLStatusWaitConfiguration(c.GetDeviceACLDetails, deviceID, 1*time.Second, timeout),
 			)
 		}
+	}
+	if _, found := changes[networkDeviceSchemaNames["AdditionalBandwidth"]]; found {
+		configs = append(configs,
+			createNetworkDeviceAdditionalBandwidthStatusWaitConfiguration(c.GetDeviceAdditionalBandwidthDetails, deviceID, 1*time.Second, timeout),
+		)
 	}
 	return configs
 }
--- a/equinix/resource_network_device.go
+++ b/equinix/resource_network_device.go
@@ -545,6 +545,12 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 						ValidateFunc: validation.StringIsNotEmpty,
 						Description:  networkDeviceDescriptions["ACLTemplateUUID"],
 					},
+					networkDeviceSchemaNames["MgmtAclTemplateUuid"]: {
+						Type:         schema.TypeString,
+						Optional:     true,
+						ValidateFunc: validation.StringIsNotEmpty,
+						Description:  networkDeviceDescriptions["MgmtAclTemplateUuid"],
+					},
 					networkDeviceSchemaNames["SSHIPAddress"]: {
 						Type:        schema.TypeString,
 						Computed:    true,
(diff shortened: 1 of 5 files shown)


equinix/resource_network_device.go
```diff
@@ -1467,7 +1478,7 @@ func getNetworkDeviceStateChangeConfigs(c ne.Client, deviceID string, timeout ti
 	configs := make([]*resource.StateChangeConf, 0, len(changes))
 	for change, changeValue := range changes {
 		switch change {
-		case networkDeviceSchemaNames["ACLTemplateUUID"]:
+		case networkDeviceSchemaNames["ACLTemplateUUID"], networkDeviceSchemaNames["MgmtAclTemplateUuid"]:
 			aclTempID, ok := changeValue.(string)
 			if !ok || aclTempID == "" {
 				break
```
displague, line 1484: Should this be `continue`?

displague: For example, what would happen if the user had:
```hcl
resource "equinix_network_device" "foo" {
  acl_template_uuid = "<uuid>"
  mgmt_acl_template_uuid = "<uuid>"
  ...
}
```
and then changed to:
```
resource "equinix_network_device" "foo" {
  mgmt_acl_template_uuid = "<uuid>"
  ...
}
```

I'm wondering if we end up with `aclTempID == ""` and we `break` out of the loop without updating the other changes (including `AdditionalBandwidth`).

displague: @rling-equinix you mentioned that https://github.com/equinix/terraform-provider-equinix/blob/63f33e2a542a6e8530c11a41e2f8ffd13a434178/equinix/resource_network_device.go#L961-L986 will always log the warning because the ACL can not be reset to `""`. 

Should we open an issue to track that or address it here?

equinix/resource_network_device_acc_test.go
```diff
@@ -875,16 +879,18 @@ func TestAccNetworkDevice_PaloAlto_Cluster_Self_BYOL(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet(deviceResourceName, "mgmt_acl_template_uuid"),
 					testAccNeDeviceACL(clusterAclResourceName, &primary),
 				),
 			},
 			{
 				ResourceName:            deviceResourceName,
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"cluster_details.0.node0.0.license_token", "cluster_details.0.node1.0.license_token"},
+				ImportStateVerifyIgnore: []string{"cluster_details.0.node0.0.license_token", "cluster_details.0.node1.0.license_token", "mgmt_acl_template_uuid"},
```
displague, line 893: The `mgmt_acl_template_uuid` field is not returned by the API, therefore it is not `d.Set` or supported for import. 👍 

@rling-equinix should we mention that these three fields are not supported for import? 
https://github.com/equinix/terraform-provider-equinix/pull/115/files#diff-d3be87f03d53e46b1cb409535110711814c46716d6a3ec94c3167f842de3b796R260

displague: If a user does not define `mgmt_acl_template_uuid` in their HCL `equinix_network_device` resource config and they `terraform import` a remote resource (not created with Terraform, for example), the desired state of "no management ACL" will not be applied to the remote resource.

This would also be the case for Network Edge devices that have had the Management ACL added in the UI.  The Terraform desired state will not be enforced.

displague: Nothing to change here - just pointing this out.

displague: To remove a Mgmt ACL, a user would have to replace the existing Mgmt ACL with the UUID referring to a template with an empty `InboundRules` list.

equinix/resource_network_device_acc_test.go
```diff
@@ -1522,27 +1536,56 @@ resource "equinix_network_device" "%{device-resourceName}" {
...
-     src_port = "any"
-     dst_port = "any"
+    subnet   = "192.0.0.0/24"
+    protocol = "IP"
+    src_port = "any"
+    dst_port = "any"
+  }
+}`, ctx)
+	}
+	if _, ok := ctx["mgmtAcl-secondary_name"]; ok {
```
displague, line 1579: We are not using this configuration template in the tests today. Should we keep it here or remove it until we add a test that exercises it? Can we test it in this PR?

equinix/resource_network_device_acc_test.go
displague: Perhaps we should be calling `GetDeviceACLDetails` at https://github.com/equinix/terraform-provider-equinix/blob/63f33e2a542a6e8530c11a41e2f8ffd13a434178/equinix/resource_network_device.go#L902

We would need to update `ne-go` to include all the fields of the `DeviceACLDetails` because today the only field exposed in ne-go is `Status`.

Let's create a new issue for this.

displague: #116

equinix/resource_network_device.go
```diff
@@ -1467,7 +1478,7 @@ func getNetworkDeviceStateChangeConfigs(c ne.Client, deviceID string, timeout ti
 	configs := make([]*resource.StateChangeConf, 0, len(changes))
 	for change, changeValue := range changes {
```
displague, line 1479: Should we be using a `for` loop here? We know that we are only concerned with three fields: `MgmtAclTemplateUuid`, `ACLTemplateUUID`, and `AdditionalBandwidth`.
 
Per our conversation, it sounds like there is a dependency that AdditionalBandwidth can only be applied if an ACL is defined. Therefor, we should be applying these changes in a predefined order, rather than the order the random order the `changes` map is iterated in.

displague (changes requested):
This generally looks good. 
The key changes I'd like to see are:
- updating the ne-go casing for the new attributes
- fixing or verifying what seems to be bad logic in getNetworkDeviceStateChangeConfigs (some preexisting)

Other concerns have been opened in new issues or are left for open for comment. Thanks!

equinix/resource_network_device.go
displague: something like: 

```go
if change, found := changes[networkDeviceSchemaNames["ACLTemplateUUID"]]; found {
...
}
if change, found := changes[networkDeviceSchemaNames["MgmtAclTemplateUuid"]]; found {
...
}
if change, found := changes[networkDeviceSchemaNames["AdditionalBandwidth"]]; found {
...
}
```

displague: This would process the fields in order, assuring that an ACL is applied before the Bandwidth is applied.

equinix/resource_network_device.go
rling-equinix: We should open an issue to track that. Maybe in the future, we will support deleting ACL from the device.

equinix/resource_network_device_acc_test.go
rling-equinix: Let's keep it just in case we forget it in the future. The logic and function for this secondary device acl is exactly same as single device. They call the same function to exercise the change.

equinix/resource_network_device.go
displague: addressed in 0952637

displague (approved):

equinix/resource_network_device.go
rling-equinix: https://github.com/equinix/terraform-provider-equinix/issues/118

PR #119, opened 2022-04-01
add data_source_network_device.go

wrgeorge1983:
This adds a data source for network devices so an existing network device can be referenced.

### DIFF
--- a/equinix/data_source_network_device.go
+++ b/equinix/data_source_network_device.go
@@ -1,0 +1,906 @@
+package equinix
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/equinix/ne-go"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var deviceStates = []string{ // Not sure if other states should be included
+	// ne.DeviceStateClusterSetUpInProgress,
+	// ne.DeviceStateDeprovisioned,
+	// ne.DeviceStateDeprovisioning,
+	// ne.DeviceStateFailed,
+	// ne.DeviceStateInitializing,
+	ne.DeviceStateProvisioned,
+	// ne.DeviceStateProvisioning,
+	// ne.DeviceStateWaitingClusterNodes,
+	// ne.DeviceStateWaitingPrimary,
+	// ne.DeviceStateWaitingSecondary,
+}
+
+var ecxNetworkDeviceSchemaNames = map[string]string{
+	"UUID":                "uuid",
+	"Name":                "name",
+	"TypeCode":            "type_code",
+	"Status":              "status",
+	"MetroCode":           "metro_code",
+	"IBX":                 "ibx",
+	"Region":              "region",
+	"Throughput":          "throughput",
+	"ThroughputUnit":      "throughput_unit",
+	"HostName":            "hostname",
+	"PackageCode":         "package_code",
+	"Version":             "version",
+	"IsBYOL":              "byol",
+	"LicenseToken":        "license_token",
+	"LicenseFile":         "license_file",
+	"LicenseFileID":       "license_file_id",
+	"LicenseStatus":       "license_status",
+	"ACLTemplateUUID":     "acl_template_id",
+	"MgmtAclTemplateUuid": "mgmt_acl_template_uuid",
+	"SSHIPAddress":        "ssh_ip_address",
+	"SSHIPFqdn":           "ssh_ip_fqdn",
+	"AccountNumber":       "account_number",
+	"Notifications":       "notifications",
+	"PurchaseOrderNumber": "purchase_order_number",
+	"RedundancyType":      "redundancy_type",
+	"RedundantUUID":       "redundant_id",
+	"TermLength":          "term_length",
+	"AdditionalBandwidth": "additional_bandwidth",
+	"OrderReference":      "order_reference",
+	"InterfaceCount":      "interface_count",
+	"CoreCount":           "core_count",
+	"IsSelfManaged":       "self_managed",
+	"WanInterfaceId":      "wan_interface_id",
+	"Interfaces":          "interface",
+	"VendorConfiguration": "vendor_configuration",
+	"UserPublicKey":       "ssh_key",
+	"ASN":                 "asn",
+	"ZoneCode":            "zone_code",
+	"Secondary":           "secondary_device",
+	"ClusterDetails":      "cluster_details",
+}
+
+var ecxNetworkDeviceDescriptions = map[string]string{
+	"UUID":                "Device unique identifier",
+	"Name":                "Device name",
+	"TypeCode":            "Device type code",
+	"Status":              "Device provisioning status",
+	"MetroCode":           "Device location metro code",
+	"IBX":                 "Device location Equinix Business Exchange name",
+	"Region":              "Device location region",
+	"Throughput":          "Device license throughput",
+	"ThroughputUnit":      "Device license throughput unit (Mbps or Gbps)",
+	"HostName":            "Device hostname prefix",
+	"PackageCode":         "Device software package code",
+	"Version":             "Device software software version",
+	"IsBYOL":              "Boolean value that determines device licensing mode: bring your own license or subscription (default)",
+	"LicenseToken":        "License Token applicable for some device types in BYOL licensing mode",
+	"LicenseFile":         "Path to the license file that will be uploaded and applied on a device, applicable for some device types in BYOL licensing mode",
+	"LicenseFileID":       "Unique identifier of applied license file",
+	"LicenseStatus":       "Device license registration status",
+	"ACLTemplateUUID":     "Unique identifier of applied ACL template",
+	"MgmtAclTemplateUuid": "Unique identifier of applied MGMT ACL template",
+	"SSHIPAddress":        "IP address of SSH enabled interface on the device",
+	"SSHIPFqdn":           "FQDN of SSH enabled interface on the device",
+	"AccountNumber":       "Device billing account number",
+	"Notifications":       "List of email addresses that will receive device status notifications",
+	"PurchaseOrderNumber": "Purchase order number associated with a device order",
+	"RedundancyType":      "Device redundancy type applicable for HA devices, either primary or secondary",
+	"RedundantUUID":       "Unique identifier for a redundant device, applicable for HA device",
+	"TermLength":          "Device term length",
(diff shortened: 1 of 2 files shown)


equinix/data_source_network_device.go
```diff
@@ -0,0 +1,906 @@
...
+	// ne.DeviceStateFailed,
+	// ne.DeviceStateInitializing,
+	ne.DeviceStateProvisioned,
+	// ne.DeviceStateProvisioning,
+	// ne.DeviceStateWaitingClusterNodes,
+	// ne.DeviceStateWaitingPrimary,
+	// ne.DeviceStateWaitingSecondary,
+}
+
+var ecxNetworkDeviceSchemaNames = map[string]string{
```
wrgeorge1983, line 26: `SchemaNames` and `SchemaDescriptions` vars are copied whole-sale from the Network Device Resource.  Could have referenced them directly but not sure if that's a 'bad idea', or not idiomatic etc.

equinix/data_source_network_device.go
```diff
@@ -0,0 +1,906 @@
...
+var ecxDeviceVendorConfigDescriptions = map[string]string{
+	"Hostname":       "Hostname. This is necessary for Palo Alto, Juniper, and Fortinet clusters",
+	"AdminPassword":  "The administrative password of the device. You can use it to log in to the console. This field is not available for all device types",
+	"Controller1":    "System IP Address. Mandatory for the Fortinet SDWAN cluster device",
+	"ActivationKey":  "Activation key. This is required for Velocloud clusters",
+	"ControllerFqdn": "Controller fqdn. This is required for Velocloud clusters",
+	"RootPassword":   "The CLI password of the device. This field is relevant only for the Velocloud SDWAN cluster",
+}
+
+func createDataSourceNetworkDeviceSchema() map[string]*schema.Schema {
```
wrgeorge1983, line 194: Schema definitions were also copied and renamed from the Network Device Resource, but also I had to modify all or nearly all of the elements to be `Computed: true`, remove validators, etc.  Again not sure if this exactly the right way to do this, but Terraform definitely didn't like elements with 'Required: true' or validators and similar fields specified.

displague:
Thanks for opening this @wrgeorge1983! When do you expect to open this PR for review?
One early point I can offer is that new Resources and Datasource PRs must be accompanied by the corresponding docs changes. This involves both Description fields (as you have provided) and the markdown representation of that resource. This is unfortunately not generated today.

wrgeorge1983:
of course!  I actually forgot this was still a draft! but yes, docs make sense.  I'll put something together using existing sources as example.

wrgeorge1983:
@displague curious if you have perspective or feedback on
* currently only returning devices in 'PROVISIONED' state on the basis that I wasn't sure if other states were sensible.  i.e. if it's in "deprovisioning" it exists *now*, but won't exist soon.  Other states might indicate the device exists, but might not be able to accept connections yet, etc.  If there's other states that should be accepted they can be uncommented on lines 13-24.
* tons of duplication between this and the resource (schema definitions in particular).  I suspect there's a "better" approach to this where the resource reverences the data source or vice versa, but not experienced enough with TF providers to say for certain.

equinix/data_source_network_device.go
rling-equinix: Since the virtual device is a product of Network Edge (NE), could you rename the schemaName and schemaDescription to for example neDeviceSchemaNames and neDeviceDescriptions to avoid the confusion?

docs/data-sources/equinix_network_device.md
```diff
@@ -0,0 +1,71 @@
...
+* `name` - (Optional) Name of an existing Equinix Network Edge device
+
+NOTE: Exactly one of either `uuid` or `name` must be specified.
+
+## Attributes Reference
+
+* `uuid` - Device unique identifier
+* `status` - Device provisioning status
+  * INITIALIZING
+  * PROVISIONING  (**NOTE: Currenly data source will only return devices in this state**)
```
rling-equinix, line 35: This note should be put beside PROVISIONED category.

equinix/data_source_network_device.go
```diff
@@ -23,174 +23,6 @@ var deviceStates = []string{ // Not sure if other states should be included
 	// ne.DeviceStateWaitingSecondary,
 }
 
-var ecxNetworkDeviceSchemaNames = map[string]string{
```
wrgeorge1983: Since all of this was 100% identical to the definitions in the resource, I redirected all references to those since they're in the same namespace.  In fact many were already cross-referenced and I hadn't noticed - proof that duplicate definitions add confusion.  I took a look around the repo but couldn't find a reasonable analog for this circumstance for inspiration.  I can revert this is you feel strongly duplication is preferred here, but will need to name them something else (again, shared namespace)

displague:
@wrgeorge1983 
> currently only returning devices in 'PROVISIONED' state on the basis that I wasn't sure if other states were sensible. i.e. if it's in "deprovisioning" it exists now, but won't exist soon. Other states might indicate the device exists, but might not be able to accept connections yet, etc. If there's other states that should be accepted they can be uncommented on lines 13-24.

When in doubt, make it an option.  With resources, we nearly always only want to return the resource when it is ready for consumption, fully provisioned. With data sources, the user may have needs that don't require a fully provisioned device.
The data source could expose status as a filter, allowing the user to define the acceptable status values. This could be an array of strings or a single desired state.  I think the array of strings would be most versatile.  A default value of "PROVISIONED" may help for the general use case.

> tons of duplication between this and the resource (schema definitions in particular). I suspect there's a "better" approach to this where the resource reverences the data source or vice versa, but not experienced enough with TF providers to say for certain.

One approach that we take in this provider is to leverage schema building functions that return the base schema and then layer additional schema to suit the resource (or data source).

`git grep func.*Schema`, and take a look at `equinix/resource_metal_reserved_ip_block.go` and how `metalIPResourceComputedFields` is used, for example.

rling-equinix:
@displague @wrgeorge1983 
I think in most scenarios, we cannot reuse that schema because almost all the fields are computed in the data source we try to create here. Any suggestions for that?

equinix/data_source_network_device.go
```diff
@@ -0,0 +1,759 @@
...
+	return map[string]*schema.Schema{
+		neDeviceSchemaNames["UUID"]: {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: neDeviceDescriptions["UUID"],
+		},
+		neDeviceSchemaNames["Name"]: {
+			Type:     schema.TypeString,
+			Optional: true,
```
rling-equinix, line 36: We don't need optional flag any more in case of data source. I think for most of the schema defined here, we cannot reuse from resource_network_device.go.

equinix/data_source_network_device.go
```diff
@@ -0,0 +1,759 @@
...
+		ReadContext: dataSourceNetworkDeviceRead,
+		Description: "Use this data source to get details of Equinix Network Edge network device with a given Name or UUID",
+		Schema:      createDataSourceNetworkDeviceSchema(),
+	}
+}
+
+func getDeviceByName(deviceName string, conf *Config) (*ne.Device, error) {
+	var devices []ne.Device
+	err := error(nil)
+	devices, err = conf.ne.GetDevices(neDeviceStates)
```
rling-equinix, line 573: We provide searchText="deviceName or device uuid" as query parameters in this GET devices API to search by name or uuid. 
It looks like we forget to put it in the documentation. Will update our public documentation.

equinix/data_source_network_device.go
```diff
@@ -0,0 +1,759 @@
...
+			secondary.LicenseFile = secondaryFromSchema.LicenseFile
+		}
+		if err := d.Set(neDeviceSchemaNames["Secondary"], flattenNetworkDeviceSecondary(secondary)); err != nil {
+			return fmt.Errorf("error reading Secondary: %s", err)
+		}
+	}
+	if primary.ClusterDetails != nil {
+		if v, ok := d.GetOk(neDeviceSchemaNames["ClusterDetails"]); ok {
+			clusterDetailsFromSchema := expandNetworkDeviceClusterDetails(v.([]interface{}))
+			primary.ClusterDetails.Node0.LicenseFileId = clusterDetailsFromSchema.Node0.LicenseFileId
```
rling-equinix, line 749: This section might not be required in the data source because in the API response, we never provide license token and license file id back.

equinix/data_source_network_device.go
```diff
@@ -0,0 +1,759 @@
...
+	if err := d.Set(neDeviceSchemaNames["ASN"], primary.ASN); err != nil {
+		return fmt.Errorf("error reading ASN: %s", err)
+	}
+	if err := d.Set(neDeviceSchemaNames["ZoneCode"], primary.ZoneCode); err != nil {
+		return fmt.Errorf("error reading ZoneCode: %s", err)
+	}
+	if secondary != nil {
+		if v, ok := d.GetOk(neDeviceSchemaNames["Secondary"]); ok {
+			secondaryFromSchema := expandNetworkDeviceSecondary(v.([]interface{}))
+			secondary.LicenseFile = secondaryFromSchema.LicenseFile
```
rling-equinix, line 740: Similarly, license file will not be returned by the API. We might not need this section as well.

rling-equinix:
> @displague great feedback. I'm taking a crack at it, though will likely be a few days. Feel free to pester me if you don't see something by monday.

Piggyback to what Marques suggest, we also provide status as query parameters to fetch the devices. We use comma to delimit multiple status. An example would be 
```
https://api.equinix.com/ne/v1/devices?offset=0&limit=20&metroCode=DC&status=PROVISIONED, PROVISIONING
```

equinix/data_source_network_device.go
wrgeorge1983: the 'optional' here is to permit the user to specify the UUID *or* the device name (but not both).  Right now this validated in the code itself (in `dataSourceNetworkDeviceRead()`).  I saw elsewhere that there is a schema attribute "ExactlyOneOf" that expresses this clearly, so I'll use this instead, as that will be more clear and explicit.

equinix/data_source_network_device.go
wrgeorge1983: > Piggyback to what Marques suggest, we also provide status as query parameters to fetch the devices. We use comma to delimit multiple statuses. 

Sure.  to be clear I'm not implementing the API calls directly from the ground up, instead simply using the existing functions 
 from the `ne-go` package:

```go
type Client interface {
       ...
	GetDevice(uuid string) (*Device, error)
	GetDevices(statuses []string) ([]Device, error)
```

So my lookups here will just inherit the correctness of this package and the underlying API, and any improvements done there.

wrgeorge1983:
There's been a bunch of back/forth, so just want to clarify that I *believe* I've addressed concerns and requests so far.  If there's something I missed please let me know

displague (approved):
Thanks @wrgeorge1983 for the detailed datasource addition and documentation!

Issue #122, opened 2022-04-06
can't just use auth_token to use metal things


cprivitere:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform Version

1.1.7

### Affected Resource(s)

* equinix_metal_*

### Terraform Configuration Files

```tf
terraform {
  required_providers {
    equinix = {
      source = "equinix/equinix"
      version = "1.6.0-alpha.2"
    }
  }
}

provider "equinix" {
  # Configuration options
  auth_token = var.auth_token
}

variable auth_token {
type = string
}
variable project_id {
type = string
}

resource "equinix_metal_device" "vmware" {
  hostname         = "vmware-a"
  plan             = "m3.large.x86"
  metro            = "da"
  operating_system = "vmware_esxi_7_0U2a"
  billing_cycle    = "hourly"
  project_id       = var.project_id
  tags             = ["vmware"]
}
```

### Expected Behavior

I should be able to provision equinix_metal resources.

### Actual Behavior

I get the following error:
```
❯ terraform plan
╷
│ Error: one of 'token' or pair 'client_id' - 'client_secret' must be set
│ 
│   with provider["registry.terraform.io/equinix/equinix"],
│   on main.tf line 10, in provider "equinix":
│   10: provider "equinix" {
│ 
```
It seems to require either token or client_id + client_secret, even though neither of those are used for equinix_metal provisioning.

### Steps to Reproduce

1. `terraform apply`

displague:
The fix for this has not been fully released. It is available in v1.6.0-alpha.3

btlogy:
I can confirm this also fix the need of having both METAL_AUTH_TOKEN and EQUINIX_API_TOKEN variables set in the environment (the 2nd being any non-empty string).

Thank you  @displague

Issue #124, opened 2022-04-08
Capture and standardize errors from all upstream APIs


ocobleseqx:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

There are already defined some functions to capture errors from Equinix Metal API. 

https://github.com/equinix/terraform-provider-equinix/blob/6dabc41bb2e66e5a61bb064914003628603be079/equinix/errors.go#L17

This issue aims to extend that functionality to all clients consumed and standardize the format in which errors are displayed. For example, when a 401 occurs in Equinix Metal it shows "Error: GET <<endpoint>>: 401  Invalid authentication token" and in Equinix Fabric it is "Equinix REST API error: Message: "HTTP operation failed: Get <<endpoint>>: oauth2: error when acquiring token: code: S1006, message Login Failure, please try again." Capturing this errors will let us expose same message and point out to the provider configuration documentation.

PR #125, opened 2022-04-08
Fix provider required credentials

ocobleseqx:
The required credentials for the provider configuration were not being validated correctly and there was a dependency on the credentials arguments required for Equinix Fabric and Network Edge.

Now it is possible to define credentials for Equinx Metal individually. It has been improved the error message to explain possible combinations and link to documentation. This fixes https://github.com/equinix/terraform-provider-equinix/issues/122 however the error message that will be returned if someone tries to interact with one of the services without the required credentials relies on the upstream API without consistency across all services exposed in the provider. I've opened the issue https://github.com/equinix/terraform-provider-equinix/issues/124 to capture and standardize errors from all upstream APIs.


equinix/config.go
```diff
@@ -28,6 +28,23 @@ const (
 	consumerToken = "aZ9GmqHTPtxevvFq9SK3Pi2yr9YCbRzduCSXF2SNem5sjB91mDq7Th3ZwTtRqMWZ"
 	metalBasePath = "/metal/v1/"
 	uaEnvVar      = "TF_APPEND_USER_AGENT"
+	resourceProviderRegistrationError = `the provider needs to be configured with the proper credentials before it
+can be used.
+
+One of pair "client_id" - "client_secret" or "token" must be set in the provider
+configuration to interact with Equinix Fabric and Network Edge services, and
+"auth_token" to interact with Equinix Metal. All them can also be configured using
```
displague, line 36:
```suggestion
"auth_token" to interact with Equinix Metal. These can also be configured using
```

equinix/provider.go
```diff
@@ -171,31 +168,20 @@ func Provider() *schema.Provider {
 
 func configureProvider(ctx context.Context, d *schema.ResourceData, p *schema.Provider) (interface{}, diag.Diagnostics) {
 	mrws := d.Get("max_retry_wait_seconds").(int)
+    rt := d.Get("request_timeout").(int)
```
displague, line 171: the indentation on this line appears off

equinix/provider.go
```diff
@@ -171,31 +168,20 @@ func Provider() *schema.Provider {
...
-	}
-	if v, ok := d.GetOk("client_secret"); ok {
-		config.ClientSecret = v.(string)
+		AuthToken:      d.Get("auth_token").(string),
+		BaseURL:        d.Get("endpoint").(string),
+		ClientID:       d.Get("client_id").(string),
+		ClientSecret:   d.Get("client_secret").(string),
+		Token:          d.Get("token").(string),
+		RequestTimeout: time.Duration(rt) * time.Second,
+		PageSize:       d.Get("response_max_page_size").(int),
```
displague, line 180: It looks like `PageSize` this can still be `nil`. Should we define a default?

displague: Hmm. That said, the acceptance tests didn't have any problem with this.

ocobleseqx: That Get function calls [ValueOrZero](https://github.com/hashicorp/terraform-plugin-sdk/blob/v1.17.2/helper/schema/field_reader.go#L37) internally to ensure a consistent non-nil return value. In case of `int` it returns `0`

displague (approved):
lgtm

Issue #134, opened 2022-05-06
r/equinix_metal_organization add required field address


ocobleseqx:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

![image](https://user-images.githubusercontent.com/68897552/167197195-16b61376-563b-40c1-bb82-8f997b414bb5.png)

It must be implemented in upstream go SDK 

![image](https://user-images.githubusercontent.com/68897552/167197337-d5ea53b7-0d88-450e-9fe6-9dba30fa13c5.png)

### Affected Resource(s)

* equinix_metal_organization

ocobleseqx:
Address was already implemented in the go SDK https://github.com/packethost/packngo/issues/332#issuecomment-1120838297

displague:
now available in https://github.com/packethost/packngo/releases/tag/v0.24.0

PR #137, opened 2022-05-10
Metal organization address field

ocobleseqx:
This PR adds field address (mandatory now to create an Equinix Metal organization) in both resource and data source `equinix_metal_organization`. 

- Some `optional` tags and validations for nil values have been added to the data source to be able to return previous organizations without address.
- Acceptance test updated. 
- Documentation updated.

fix https://github.com/equinix/terraform-provider-equinix/issues/134


displague (approved):
Great to see this test passing again with the added Organization fields!

PR #141, opened 2022-05-13
add metal_device support for user_ssh_key_ids

displague:
Fixes https://github.com/equinix/terraform-provider-equinix/issues/140

### DIFF
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -619,7 +619,7 @@ resource "equinix_metal_ssh_key" "test" {
 	public_key = "%s"
 }
 
-resource "equinix_metal_ssh_project_key" "test" {
+resource "equinix_metal_project_ssh_key" "test" {
 	project_id = equinix_metal_project.test.id
 	name = "tfacc-project-key-%s"
 	public_key = "%s"
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -85,22 +85,27 @@ func TestAccMetalDevice_facilityList(t *testing.T) {
 	})
 }
 
-func TestAccMetalDevice_networkPortsOrder(t *testing.T) {
-	var device packngo.Device
+func TestAccMetalDevice_sshConfig(t *testing.T) {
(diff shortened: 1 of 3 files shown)


equinix/resource_metal_device_acc_test.go
```diff
@@ -85,22 +85,27 @@ func TestAccMetalDevice_facilityList(t *testing.T) {
 	})
 }
 
-func TestAccMetalDevice_networkPortsOrder(t *testing.T) {
-	var device packngo.Device
+func TestAccMetalDevice_sshConfig(t *testing.T) {
```
displague, line 88: I replaced `TestAccMetalDevice_networkPortsOrder` with `TestAccMetalDevice_sshConfig` and moved the two tests "networkPortsOrder" had to the "_basic" test. The HCL was the same.

Once concern of this is that `networkPortsOrder` is currently failing because device ports are no longer returned in the order that we expected when the test was written.

I think we have two options here:
* update the documentation to assert order is not predictable. `for` expressions should be used to match ports by name.
* fix the order that so that the legacy behavior is preserved. This could also mean extending the pattern into undefined areas.  Should we offer ports in eth0, eth1, bond0, eth2, eth3, bond1 order? What happens if more devices are aggregated into a single bond in the future? What happens when more ports or bonds are added? Is it wise to depart from the format the API is returning? Is the API adopting an ordering model of [ethN..., bondN...]?

With that stated, I think the NetworkPortsOrder test should ultimately be removed and the documentation should be updated (with appropriate CHANGELOG messaging and versioning).

displague: I've not updated the documentation for Metal Device network port order because that update is out of scope for this PR.

My current thinking is that the ports should be a `TypeSet` for a few reasons:
* this will prevent jitter if/when the API returns the ports in a different order
* there is no "good" order that the provider can prescribe
* this will force `for` expression consumption because the map index of `TypeSet` is somewhat unpredictable

ocobleseqx (dismissed):
lgtm.

New tests `TestAccMetalDevice_sshConfig` cant take advantage of #215 to select a valid plan/metro but this can be done in a new PR

ocobleseqx (approved):

Issue #66, opened 2021-11-29
Support additional Equinix APIs


displague:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

Equinix provides several APIs not represented in this Terraform provider.

Today, Equinix APIs are offered in the following format:

https://api.equinix.com/{service}/{version}/

Compatibility between versions is in no way guaranteed, these versions typically represent a major API revision in [semver](https://semver.org) parlance.  This is most evident in the differences between ecx/v2 and ecx/v3 service revisions.

Terraform providers "bridge Terraform Core and their respective target APIs". It does not benefit Terraform users or Terraform provider authors to invent translation layers and features that do not exist in the API.  Terraform providers should not offer magic. Doing so creates domain logic tightly controlled in the Terraform provider. This project, the Terraform provider, does not have the authority to dictate how the Equinix APIs work. Changes in the upstream APIs can upheave any of the underpinnings of magic interfaces (as we have seen with the ecx_l2_connection_accepter resource #64).

The most faithful representation of the Equinix API in Terraform, therefor, must include versioned resources:

`equinix_ecxv2_{endpoint}`
`equinix_ecxv3_{endpoint}`
`equinix_fabricv4_{endpoint}` https://developer.equinix.com/docs/fabric-apis-v4
`equinix_metalv1_{endpoint}` (currently offered by the https://registry.terraform.io/providers/equinix/metal/latest provider)
This approach will remove the fragility present in today's resource structure.

There is precedent for this format in other providers, including AWS: https://registry.terraform.io/providers/hashicorp/aws/latest/docs.  In the AWS provider, versioning does not appear planned for in Terraform resource names so we see confusing patterns, including: `aws_foov2`, `aws_foo_v2`, and `aws_foo` vs `aws_otherfoo`.  The provider will benefit from adopting a pattern early and predictably.

Alternatives considered:

* _`ecx_{endpoint} (independent ECX provider, v2 and v3_): Too many redundant provider configuration blocks would be needed. Any opportunity to share API connection (rate limit throttling) handling would be lost.
[...]

PR #142, opened 2022-05-17
move all resources to internal/

### LINKED ISSUE #106: adopt internal/ layout structure
The number of files in `equinix/` has been expanding with recent additions. The developer experience can be improved by compartmentalizing resources into an alternate structure. This practice has become more popular in the past year or so.

Proposed structure:

* `equinix/provider.go`
   By maintaining the "provider" package, downstream tooling such as Pulumi, Kubeform, Crossplane, and the OpenShift IPI driver (under development, paused) may continue to import the project
* `internal/services/metal/device/schema.go`
   This would combine all schemas into a single file.
   This is a pattern found in Kubebuilder `{resourcename}_types.go` files for a single resource.  In Kubebuilder all resources reside in their own file, packaged by API version.  
   Crossplane follows a similar pattern with an extra service path (https://github.com/crossplane-contrib/provider-tf-equinix-metal/tree/main/apis/connection/v1alpha1). An interesting distinction about Crossplane's pattern is that the schema symbols are exported. This may be a requirement for some downstream tooling (I haven't researched or identified this).
* `internal/services/metal/device/resource.go`
* `internal/services/metal/device/datasource.go`
* `internal/services/networkedge/device/resource.go`
* other common files should flow naturally by refactoring the single resource files today. Each resource could, for example, include flatteners, expanders, converters, and other helpers.
* `internal/migrationtool/`
* A matching `_test.go` for each file

Considerations:

* removing exported symbols from `equinix/` is a breaking change from a Go and user perspective. I believe we can safely assert from the aforementioned consumers, that most if not all Go consumers are interested in the `Provider` type, and nothing else.
* Linode's provider puts the HCL templates used during E2E testing into a separate folder and file for each template. A template builder pattern allows for a functional, chaining, approach to modifying the test template.

Inspiration:
* https://github.com/linode/terraform-provider-linode/tree/main/linode (the Equinix provider offers multiple service trees, in the proposed paths above, the folder is one level deeper to distinguish services)
* https://github.com/hashicorp/terraform-provider-aws/tree/main/internal/service (this level of break-up seems unnecessary for the size of the Equinix provider.)
* https://github.com/hashicorp/terraform-provider-azurerm/tree/main/internal/services  (The `internal/services` pattern could be useful in the Equinix provider to keep service implementation code away from, migration tool code, for example.)

displague:
This is a major refactor that I wasn't able to break up meaningfully.

Goals:
* avoid exporting anything that can break Go API compatibility between releases
* separate resources into easier to template and reproduce patterns
* identify and remove unused code
* prepare for the Terraform plugin Framework patterns (#365 )
* reduce hidden codependencies 
* isolate packages to introduce alternate Go SDKs (generated ones)
* OWNERS files that map projects to the appropriate product API teams

Status:
* `go build` works.
* Tests need additionally refactoring to avoid cyclic dependency loops. I find this between internal/tfacc and internal/provider. I think the internal/provider may need to be moved in-part, back to equinix/.
  This will be easier to catch in the future because smaller changes will be responsible for triggering the loop.
In a few places where two resources shared common
* Additional cleanup that could happen would be to remove package name studdering:
  ```
  ecxL2SellerProfileAdditionalInfosSchemaNames -> (ecx/l2sellerprofile) additionalInfosSchemaNames
  FlattenECXL2SellerProfileAdditionalInfos -> (ecx/l2sellerprofile) FlattenAdditionalInfos
  ```

Somewhat inspired by https://github.com/equinix/terraform-provider-equinix/issues/66
Fixes #106


displague:
For the problems plumbing provider, I intend to investigate the ProviderFactories approach, https://github.com/hashicorp/terraform-provider-aws/issues/24500

Issue #144, opened 2022-05-23
Fix data source metal connection - error retrieving existing connection without defined speed


ocobleseqx:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

Error found[ running acceptance tests](https://github.com/equinix/terraform-provider-equinix/runs/6427198863?check_suite_focus=true#step:5:380
):

> Error: 0 is not allowed speed value. Allowed values: [50000000 200000000 500000000 1000000000 2000000000 5000000000 10000000000]

Creating a connection through the portal doesn't allow specify speed. For that reason, no speed must be also allowed.

This error have been fixed in the legacy [terraform metal provider ](https://github.com/equinix/terraform-provider-metal/pull/223) along with other issues that must be implemented here as well.

PR #206, opened 2022-06-21
r/ds equinix_metal_connection fixed and sync behavoir with terraform-provider-metal

ocobleseqx:
Sync https://github.com/equinix/terraform-provider-metal/pull/223

- Always return project_id specified in the schema (to cover dedicated connection without project), and later for shared connections check project assigned to the virtual circuit.
- Allow retrieve existing connections created through the portal with speed value 0.
- Sort ports by role , asserting the API always returns primary for len of 1 responses (to replicate current provider equinix behavoir)
- Solved r/equinix_metal_connection producing non-empty plan after apply if type is dedicated (More details https://github.com/equinix/terraform-provider-metal/issues/222)
- Fix https://github.com/equinix/terraform-provider-equinix/issues/144

### DIFF
--- a/equinix/data_source_metal_connection.go
+++ b/equinix/data_source_metal_connection.go
@@ -196,11 +196,8 @@ func dataSourceMetalConnection() *schema.Resource {
 }
 
 func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
-	ret := make([]map[string]interface{}, len(cps))
-	order := map[packngo.ConnectionPortRole]int{
-		packngo.ConnectionPortPrimary:   0,
-		packngo.ConnectionPortSecondary: 1,
-	}
+	ret := make([]map[string]interface{}, 0, 1)
+
 	for _, p := range cps {
 		vcIDs := []string{}
 		for _, vc := range p.VirtualCircuits {
@@ -215,8 +212,7 @@ func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
 			"link_status":         p.LinkStatus,
 			"virtual_circuit_ids": vcIDs,
 		}
-		// sort the ports by role, asserting the API always returns primary for len of 1 responses
-		ret[order[p.Role]] = connPort
+		ret = append(ret, connPort)
 	}
 	return ret
 }
--- a/equinix/data_source_metal_connection.go
+++ b/equinix/data_source_metal_connection.go
@@ -216,7 +216,6 @@ func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
 			"link_status":         p.LinkStatus,
 			"virtual_circuit_ids": vcIDs,
 		}
-		ret = append(ret, connPort)
 		// sort the ports by role, asserting the API always returns primary for len of 1 responses
 		ret[order[p.Role]] = connPort
 	}
--- a/equinix/data_source_metal_connection.go
+++ b/equinix/data_source_metal_connection.go
@@ -196,7 +196,11 @@ func dataSourceMetalConnection() *schema.Resource {
 }
 
 func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
-	ret := make([]map[string]interface{}, 0, 1)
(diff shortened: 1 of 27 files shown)


equinix/resource_metal_reserved_ip_block_acc_test.go
```diff
@@ -34,7 +34,7 @@ resource "equinix_metal_project" "foobar" {
 
 resource "equinix_metal_reserved_ip_block" "test" {
 	project_id  = equinix_metal_project.foobar.id
-	facility    = "ny5"
+	facility    = "ewr1"
```
displague, line 37: ewr1 is legacy

equinix/resource_metal_virtual_circuit_acc_test.go
```diff
@@ -3,8 +3,8 @@ package equinix
 import (
 	"fmt"
 	"log"
-	"os"
 	"testing"
+	"os"
```
displague, line 7: this might not be gofmt'd - I expect "os" before "testing".

Perhaps this was applied from a patch before "testing" was added.

equinix/data_source_metal_connection.go
```diff
@@ -215,8 +212,7 @@ func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
 			"link_status":         p.LinkStatus,
 			"virtual_circuit_ids": vcIDs,
 		}
-		// sort the ports by role, asserting the API always returns primary for len of 1 responses
-		ret[order[p.Role]] = connPort
+		ret = append(ret, connPort)
```
displague, line 215: This looks like a revert of the correct approach.

equinix/resource_metal_connection.go
```diff
@@ -104,8 +104,8 @@ func resourceMetalConnection() *schema.Resource {
 			},
 			"project_id": {
 				Type:        schema.TypeString,
-				Required:    true,
-				Description: "ID of the project where the connection is scoped to, only used for type == \"shared\"",
+				Optional:    true,
+				Description: "ID of the project where the connection is scoped to. Required with type \"shared\"",
```
displague, line 108: I don't know which of these is correct. The previous hunk that reverts `ret[order[p.Role]] = connPort` has me calling everything in this patch into question.

displague (changes requested):
I think the patch may have been applied in reverse in some sections.

equinix/resource_metal_connection.go
ocobleseqx: Thanks for the detailed review @displague, I will check again all the changes. In this case, both project and organization should be optional. Some of the introduced errors you mentioned were my fault in choosing the wrong option when reviewing the rejects.

equinix/data_source_metal_connection.go
```diff
@@ -213,6 +217,8 @@ func getConnectionPorts(cps []packngo.ConnectionPort) []map[string]interface{} {
 			"virtual_circuit_ids": vcIDs,
 		}
 		ret = append(ret, connPort)
```
displague, line 219: `ret[order[p.Role]] = connPort` is a replacement for this `append` line should.

equinix/data_source_metal_vlan_acc_test.go
```diff
@@ -115,6 +115,7 @@ data "equinix_metal_vlan" "dsvlan" {
 resource "equinix_metal_vlan" "barvlan" {
     project_id = equinix_metal_project.foobar.id
     metro = equinix_metal_vlan.foovlan.metro
+	description = "%s"
```
displague, line 118: this appears to lead with a tab where there others lead with spaces

displague (changes requested):

displague:
The branch has merge conflicts.

Also this failed acceptance test might need a look:
```
=== CONT  TestAccMetalConnection_shared
    testing_new.go:70: Error running post-test destroy, there may be dangling resources: exit status 1
        Error: Timeout waiting for connection to be deleted
--- FAIL: TestAccMetalConnection_shared (71.09s)
```

displague:
merge conflicts - otherwise looks good

PR #208, opened 2022-06-23
specify test location dynamically in L2 VLAN attachment tests

ocobleseqx:
Formerly: https://github.com/equinix/terraform-provider-metal/pull/66 created by @displague 

I transferred this PR as a draft, it seems it got stuck due to an unrelated error

@displague can you confirm this PR is ready to go?


displague:
The PR looks like general test cleanup now, which was the original intent before it ran into other problems (which you discovered).  I believe all of that was settled.

This has some merge conflicts to resolve. (We didn't get the 504 error)

displague (approved):

PR #210, opened 2022-06-23
Add end_at param to spot_market_request to allow limiting SMR duration

ocobles:
Formerly: https://github.com/equinix/terraform-provider-metal/pull/191

> This PR adds end_at parameter to metal_spot_market_request to allow removing the SMR in future. Will be useful for our acceptance tests.
> 
> Signed-off-by: Tomas Karasek [tom.to.the.k@gmail.com](mailto:tom.to.the.k@gmail.com)

I have added some changes based on the conversation in the original PR thread:

- Replaced termin**t**ation with termination in both doc and code (should instead mark the wrong schema attribute "termintation_time" as deprecated? https://www.terraform.io/plugin/sdkv2/best-practices/deprecations#renaming-a-computed-attribute)
- Update docs to keep termination_time in the attributes section
- Moved testSpotMarketRequestTerminationTime function to the SMR acceptance test file

### DIFF
--- a/docs/resources/equinix_metal_spot_market_request.md
+++ b/docs/resources/equinix_metal_spot_market_request.md
@@ -39,11 +39,11 @@ The following arguments are supported:
 * `wait_for_devices` - (Optional) On resource creation wait until all desired devices are active.
 On resource destruction wait until devices are removed.
 * `facilities` - (Optional) Facility IDs where devices should be created.
+* `end_at` - (Optional) Deadline for the request. For example \"2021-09-03T16:32:00+03:00\". If you don't supply timezone info, timestamp is assumed to be in UTC."
 * `metro` - (Optional) Metro where devices should be created.
 * `locked` - (Optional) Blocks deletion of the SpotMarketRequest device until the lock is disabled.
 * `instance_parameters` - (Required) Key/Value pairs of parameters for devices provisioned from
-this request. Valid keys are: `billing_cycle`, `plan`, `operating_system`, `hostname`,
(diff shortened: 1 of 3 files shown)


displague:
https://github.com/equinix/terraform-provider-equinix/runs/7029249739?check_suite_focus=true#step:5:1629

running into the same problem discussed in this thread: https://github.com/equinix/terraform-provider-metal/pull/191#discussion_r709127245 (termination_time can't always be imported)

pamoedom:
My 2 cents:

This option will be really good because when using TF `equinix_metal_spot_market_request` resource, the `end_at` parameter is always exported with `null` value, causing the SMR to be created without end time, which means it will continue requesting new instances until deletion, for automation, adding this parameter will allow us to simulate `one-time` SMR deploy without having to worry about new instances being spawned when not needed.

Best Regards.

rverdam:
Any change this PR will be completed, because I think it would be very helpful for my creating my lab environments based upon spot instances.

pamoedom:
Hi @ocobleseqx , any progress/update here? thanks in advance.

displague:
@ocobles Do you suppose we could rebase this and just add `end_dt` to the import ignore list? We would also have to mention that the field can not be imported in the import documentation.

https://github.com/equinix/terraform-provider-equinix/pull/210/files#diff-3e30b20e0e9e4d15ddbb5981a99fdaa23532e2918b0a86caa7be2c3cd38f97b2R166

ctreatma:

Issue #169, opened 2021-07-22
metal_plan datasource


displague:
Users would benefit from the ability to choose a plan heuristically or access the details of an Equinix Metal plan.

This may serve as an alternative to equinix/terraform-provider-equinix#180 because the [capacity endpoints](https://metal.equinix.com/developers/api/capacity/) are affectively consumed for plan filtering. (more details about the capacity endpoints in packngo: https://github.com/packethost/packngo/blob/master/capacities.go)

```hcl
data "metal_plan" "best_plan" {
  # all fields are optional.
  # uses GET /plans and GET /capacity based on criteria
  # filters responses to match criteria
  # if 'quantity' is specified, a POST /capacity check is used against each result
  #   skipped if level='unavailable'
  # should metro and facility be handled through one resource? with facility set to "" on metro filtered requests? and vice-versa?
  metros = ["am"] # optional, will only return the best plan if all metros have availability
  quantity = "3" # optional, minimum quantity for each metro
  level = "normal" # optional
  name = ["c3.small.x86"] # optional, list of plans that are ok to choose from. this could also be "class"
  slug = ["c3.small.x86"] # optional, list of plans that are ok to choose from. these are different from name for some plans.
  deployment_types = ['on_demand'] # optional, filter by deployment type
  line = 'baremetal'
  pricing_hour_min = '' 
  pricing_hour_max = '' ​
  spec_cpus_min = ''
  spec_memory_min = ''
  spec_drives_count_min = ''
  legacy = false
  order_by = 'price' # level
  order_dir = 'asc'
} 


output "foo" { value = data.metal_plan.best_plan }
```

displague:
Fixed by https://github.com/equinix/terraform-provider-equinix/pull/215

PR #215, opened 2022-07-01
New `equinix_metal_plans` data source

ocobleseqx:
Sync https://github.com/equinix/terraform-provider-metal/pull/194

> This is a new datasource for querying plans, fixing https://github.com/equinix/terraform-provider-equinix/issues/169 This is just a draft, for discussion.
> 
> It's done in the digitalocean style, like
> https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/images
> 
> ..with the filter blocks. The functionality allowes more than filters, it can also sort, and match by RE.
> 
> @displague What do you think? I think that if we want ot accumulate filters, we should do it in a clean and structured way. And here we can take advantage of an existing clean and structured way.
> 
> I basically took the filtering logic code from DO provider. It's under the same licences (MPL2), and after brief research, I think there's nothing we should explicitly do in order to have this legit. I also didn't find a way how to properly acknowledge this.
> 
> Signed-off-by: Tomas Karasek [tom.to.the.k@gmail.com](mailto:tom.to.the.k@gmail.com)


displague:
merge conflicts - otherwise looks good

PR #218, opened 2022-07-06
fix missing attribute connection_id in ds equinix_metal_virtual_circuit

ocobleseqx:
This PR adds missing `connection_id` in data source `equinix_metal_virtual_circuit`

All data source using same read function than their matching resource must have same schema keys or it [will produce an error in the d.Set(key, value) function if the key is invalid.](https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#ResourceData.Set
) 

This error was found in the equinix metal provider: https://github.com/equinix/terraform-provider-metal/runs/7204768928?check_suite_focus=true#step:5:629

### DIFF
--- a/docs/data-sources/equinix_metal_virtual_circuit.md
+++ b/docs/data-sources/equinix_metal_virtual_circuit.md
@@ -33,6 +33,7 @@ The following arguments are supported:
 In addition to all arguments above, the following attributes are exported:
 
 * `name` - Name of the virtual circuit resource.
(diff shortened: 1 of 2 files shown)


ocobleseqx:
Test `TestAccMetalVirtualCircuit_dedicated`, using `data.equinix_metal_virtual_circuit.test` PASS 

https://github.com/equinix/terraform-provider-equinix/runs/7210911208?check_suite_focus=true#step:5:687

displague (approved):

Issue #112, opened 2022-03-21
Implement Fabric API v4 resources


displague:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

[Equinix Fabric has a new v4 API](https://app.swaggerhub.com/apis/equinix-api/fabric/4.3) with an improved interface for managing connections, ports, gateways, and tokens. 

This API replaces many of the v2 and v3 APIs. Details to follow in the comments about which endpoints/resources will need to be modified or deprecated to meet the API timelines.

A Go SDK will be needed to implement the new Terraform resources.  ecx-go could be extended, but this is a hand-edited SDK with legacy branding. A new generated SDK (using openapitools/openapi-generator-cli, for example) would be advisable. SwaggerHub provides a downloadable Go client, automated or manual publishing of this client would be helpful.

### New or Affected Resource(s)

The following items should be linked to individual issues when they exist.

Proposed New Resources:
* [x] equinix_fabric_connection
       #209 
* [ ] equinix_fabric_connection_action - unclear if this should be a resource or a property
* [ ] equinix_fabric_connection_routing_protocols - unclear if this should be a resource or a property
* [ ] equinix_fabric_connection_bulk_routing_protocols - unclear if this should be a resource or a property
* [ ] equinix_fabric_bulk_connection
* [x] ~equinix_fabric_gateway~ equinix_fabric_cloud_router
* [x] equinix_fabric_service_profile
* [ ] equinix_fabric_service_tokens

Proposed New Datasources:
* [ ] equinix_fabric_connections (using /search)
* [x] equinix_fabric_connection (using id or other unique criteria)
* [ ] equinix_fabric_metros (unclear if this is needed over a generic equinix_metros datasource)
* [ ] equinix_fabric_metro (by id, unclear if this is needed over a generic equinix_metro)
* [ ] equinix_fabric_prices (uses /search)
* [ ] equinix_fabric_gateways (using /search)
* [ ] ~equinix_fabric_gateway~ equinix_fabric_cloud_router (using id or other unique criteria)
* [ ] equinix_fabric_gateway_packages
[...]

PR #230, opened 2022-07-26
Fabric v4 - Integration Branch

displague:
This is a draft PR representing all work in progress for #112.


equinix/data_source_fabric_connection.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceFabricConnection() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceFabricConnectionRead,
+		Schema:      readFabricConnectionResourceSchema(),
```
ocobleseqx, line 13: missing `Description:` for all new data sources

equinix/resource_fabric_connection.go
```diff
@@ -0,0 +1,265 @@
...
+		ReadContext:   resourceFabricConnectionRead,
+		CreateContext: resourceFabricConnectionCreate,
+		UpdateContext: resourceFabricConnectionUpdate,
+		DeleteContext: resourceFabricConnectionDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: createFabricConnectionResourceSchema(),
+
+		Description: "Resource allows creation and management of Equinix Fabric	layer 2 connections",
```
ocobleseqx, line 33: It would be good for users to specify in the description this resource is for Fabric v4. Same for all other `fabric_` resources/datasources

equinix/fabric_port_read_schema.go
```diff
@@ -0,0 +1,350 @@
...
+			Description: "Port type",
+		},
+		"href": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Port URI information",
+		},
+		"uuid": {
+			Type:        schema.TypeString,
+			Optional:    true,
```
ocobleseqx, line 216: "uuid" must be `Required` for `equinix_fabric_port` data source. However, since it is optional for  `equinix_fabric_ports` I would add a separate function to retrieve/modify `equinix_fabric_ports` schema

```
func readFabricPortsResourceSchema() map[string]*schema.Schema {
	sch := readFabricPortResourceSchema()
	sch["uuid"].Optional = true
	sch["uuid"].Required = false
	return sch
}
```

And reference it in

https://github.com/equinix/terraform-provider-equinix/pull/230/files#diff-f9ecc09f3eb05f0a73dfdafa8f04b48297432eb40fab00a66b7b070daa5a646aR338

equinix/fabric_service_profile_read_schema.go
```diff
@@ -0,0 +1,467 @@
...
+			Description: "Service profile visibility - PUBLIC, PRIVATE",
+		},
+		"name": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Customer-assigned service profile name",
+		},
+		"uuid": {
+			Type:        schema.TypeString,
+			Optional:    true,
```
ocobleseqx, line 31: Same than for `equinix_fabric_port`, the "uuid" here must be `Required`

ocobleseqx (commented):
Added some comments

equinix/fabric_port_read_schema.go
```diff
@@ -0,0 +1,350 @@
...
+	return map[string]*schema.Schema{
+		"data": {
+			Type:        schema.TypeList,
+			Computed:    true,
+			Description: "List of  Ports",
+			Elem: &schema.Resource{
+				Schema: readFabricPortResourceSchema(),
+			},
+		},
+		"local_var_optionals": {
```
ocobleseqx, line 341: It seems `local_var_optionals` is not an API parameter and it was made up in the go sdk. It is not a very intuitive name for a field, also why do we need to group them? Wouldn't it be better to have a `Name` field directly without needing that schema Set ?

equinix/resource_fabric_connection_acc_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+func checkConnectionDelete(s *terraform.State) error {
+	client := testAccProvider.Meta().(*Config).fabricClient
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, v4.ContextAccessToken, testAccProvider.Meta().(*Config).FabricAuthToken)
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "equinix_fabric_connection" {
+			continue
+		}
+		_, err := waitUntilConnectionDeprovisioned(rs.Primary.ID, client, ctx)
```
ocobleseqx, line 53: This wait function is being tested but not implemented in the process of deleting the connection resource

examples/connectivity/v4/alibaba/main.tf
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+  }
+}
+
+output "connection_result" {
+  value = equinix_fabric_connection.alibaba-qinq.id
+}
+
+resource "time_sleep" "wait_for_ingress_alb" {
```
ocobleseqx, line 66: All v4 examples have a time sleep. It seems that the name `ingress_alb` is not related to this provider. Why do we need it? I presume it has to do with the absence of a wait for deprovision function I mentioned before

equinix/provider.go
```diff
@@ -136,6 +142,8 @@ func Provider() *schema.Provider {
 			"equinix_ecx_l2_connection":          resourceECXL2Connection(),
 			"equinix_ecx_l2_connection_accepter": resourceECXL2ConnectionAccepter(),
 			"equinix_ecx_l2_serviceprofile":      resourceECXL2ServiceProfile(),
+			"equinix_fabric_connection":          resourceFabricConnection(),
+			"equinix_fabric_service_profile":     resourceFabricServiceProfile(),
 			"equinix_network_device":             resourceNetworkDevice(),
```
ctreatma, line 147: It looks like `equinix_fabric_port` is missing from the resources map

equinix/provider.go
```diff
@@ -136,6 +142,8 @@ func Provider() *schema.Provider {
 			"equinix_ecx_l2_connection":          resourceECXL2Connection(),
 			"equinix_ecx_l2_connection_accepter": resourceECXL2ConnectionAccepter(),
 			"equinix_ecx_l2_serviceprofile":      resourceECXL2ServiceProfile(),
+			"equinix_fabric_connection":          resourceFabricConnection(),
+			"equinix_fabric_service_profile":     resourceFabricServiceProfile(),
```
ctreatma, line 146: It looks like `equinix_fabric_port` is missing from the resources map

ctreatma:
The [Go Test / test job](https://github.com/equinix/terraform-provider-equinix/actions/runs/3566921756/jobs/5993978115) failed because it doesn't think the provider has an `equinix_fabric_connection`, but I can see it in [the resource map](https://github.com/equinix/terraform-provider-equinix/pull/230/files#diff-17a1d27f648a15532c90c0ed8e7b143adde23b59dec5bbe4222db41a0a94ea0aR145).  Does the "Go Tests / tests" job pull down a published equinix provider instead of building from source?  Or is there something missing in the updated code & I just don't know what it is?

ocobleseqx:
> The [Go Test / test job](https://github.com/equinix/terraform-provider-equinix/actions/runs/3566921756/jobs/5993978115) failed because it doesn't think the provider has an `equinix_fabric_connection`, but I can see it in [the resource map](https://github.com/equinix/terraform-provider-equinix/pull/230/files#diff-17a1d27f648a15532c90c0ed8e7b143adde23b59dec5bbe4222db41a0a94ea0aR145). Does the "Go Tests / tests" job pull down a published equinix provider instead of building from source? Or is there something missing in the updated code & I just don't know what it is?

These new Fabric tests run against the examples folders (e.g. [TestGcpCreateConnection](https://github.com/equinix/terraform-provider-equinix/pull/230/files#diff-6f6777cf30ef72856b8aca8d1f06fe2cb315656c105877b4185d52747360601aR13)),and [there](https://github.com/equinix/terraform-provider-equinix/blob/540d6ed20ab2f117b1b67ae583941fbd00248296/examples/connectivity/v4/google/terraform.tf) terraform uses the latest available version that doesn't have the new resources yet.

equinix/provider.go
ctreatma: Nevermind, I was thrown off by the existence of a `resource_fabric_port.go` file, but that file seems to only have read functions.

Nit:  if we can, I think it would be preferable to put those in the data source file & rename `resource_fabric_port_acc_test.go` to `data_source_fabric_port_acc_test.go`.

ctreatma (approved):
I'm approving this under the assumption that the latest commit on this branch has been built & released as an alpha & works.

PR #232, opened 2022-07-27
Fabric v4 - Port data schema implementation and Terraform Script - Create Connection example

syedluri:
Port data schema implementation and Terraform Script - Create Connection example.

Part of #112

### DIFF
--- a/equinix/data_source_fabric_port.go
+++ b/equinix/data_source_fabric_port.go
@@ -1,0 +1,21 @@
+package equinix
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceFabricPort() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceFabricPortRead,
+		Schema:      readFabricPortResourceSchema(),
+	}
+}
+
+func dataSourceFabricPortRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	uuid, _ := d.Get("uuid").(string)
+	d.SetId(uuid)
+	return resourceFabricConnectionRead(ctx, d, meta)
+}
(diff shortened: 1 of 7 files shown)


equinix/fabric_port_read_schema.go
```diff
@@ -0,0 +1,251 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Date and time at which port availability changed",
+		},
+	}
+}
+
+func readFabricPortTether() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"cross_onnect_id": {
```
displague, line 47:
```suggestion
		"cross_connect_id": {
```

I assume this is a typo

equinix/data_source_fabric_port.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceFabricPort() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceFabricPortRead,
+		Schema:      readFabricPortResourceSchema(),
```
displague, line 13: Is this schema Computed?

equinix/fabric_port_read_schema.go
```diff
@@ -0,0 +1,251 @@
...
+			Description: "Port URI information",
+		},
+		"id": {
+			Type:        schema.TypeInt,
+			Computed:    true,
+			Description: "Port ID",
+		},
+		"uuid": {
+			Type:        schema.TypeString,
+			Computed:    true,
```
displague, line 99:
```suggestion
			Required:    true,
```

displague (changes requested):
We are missing the commit that introduces the provider resource and datasource definition. `equinix/provider.go`

We'll also want `func resourceFabricPort() *schema.Resource { return &schema.Resource{ ...`

I think we can leave Create/Update/Delete unimplemented if we want to merge early and iterate from there.

PR #236, opened 2022-07-28
NFV-19565: add description field for the acl template inbound rule

rling-equinix:
As part of August release, we add a description field for each inbound rule of acl template. We want to add that field as part of terraform script as well.
This description field will always be optional and the field length should not be longer than 200 characters.
It will be part of create/update request. It will also be part of get response.
Sample payload as follows:
```
  "inboundRules": [
        {
            "description": "description of the rule",
            "protocol": "TCP",
            "srcPort": "any",
            "dstPort": "any",
            "subnet": "216.221.225.13/32",
            "seqNo": 1
        },
        {
            "protocol": "TCP",
            "srcPort": "53",
            "dstPort": "any",
            "subnet": "1.1.1.1/32",
            "seqNo": 2
        }
  ]
```

### DIFF
--- a/docs/resources/equinix_network_acl_template.md
+++ b/docs/resources/equinix_network_acl_template.md
@@ -22,6 +22,7 @@ resource "equinix_network_acl_template" "myacl" {
     protocol = "IP"
     src_port = "any"
     dst_port = "any"
+    description = "inbound rule description"
   }
   inbound_rule {
     subnet  = "172.16.25.0/24"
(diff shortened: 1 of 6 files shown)


ocobleseqx (approved):
Looks good!

- [x] flatten/expand inbound rules updated
- [x] docs updated
- [x] tests updated
- [x] length validations in place
- [x] ne-go version updated

PR #244, opened 2022-08-18
metal device operating_system should only be forcenew when reinstall is not enabled

displague:
Part of #158


ocobleseqx (approved):
CustomizeDiff logic implemented in https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_device.go#L413

PR #253, opened 2022-09-21
termintation -> termination

grahamc:
Signed-off-by: Graham Christensen <graham@grahamc.com>


displague:
`git grep` for `termintation_time`, this will also need to be corrected in the docs.

Note: this is a breaking change.

PR #255, opened 2022-09-27
add module_name provider metadata

### LINKED ISSUE #252: UA Prefix
### Community Note

### Description

The Terraform provider should offer a UA Prefix provider parameter. This would be used in Terraform modules as a way to identify the module as the API consumer. This could also be used by providers like Pulumi, Kubeform, and Crossplane which extend the TF Provider and are innacurately identified as "Terraform" agents today.

For example: https://github.com/linode/terraform-provider-linode/pull/40/commits/1d6bb71e4d3efac46af78589c11afbd008397ffc

* `provider`

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
```

### References

* #0000

displague:
Fixes #252 

Env `TF_APPEND_USER_AGENT` is also available.

### DIFF
--- a/equinix/data_source_metal_project_acc_test.go
+++ b/equinix/data_source_metal_project_acc_test.go
@@ -38,6 +38,12 @@ func TestAccDataSourceMetalProject_basic(t *testing.T) {
 
 func testAccDataSourceMetalProject_basic(r string) string {
 	return fmt.Sprintf(`
+terraform {
+	provider_meta "equinix" {
+		module_name = "test"
+	}
+}
+
 resource "equinix_metal_project" "foobar" {
 	name = "tfacc-project-%s"
 	bgp_config {
--- a/equinix/config.go
+++ b/equinix/config.go
@@ -78,6 +78,7 @@ type Config struct {
 	RequestTimeout time.Duration
 	PageSize       int
 	Token          string
+	UAPrefix       string
 
 	ecx   ecx.Client
 	ne    ne.Client
@@ -123,10 +124,10 @@ func (c *Config) Load(ctx context.Context) error {
 		neClient.SetPageSize(c.PageSize)
 	}
 	ecxClient.SetHeaders(map[string]string{
-		"User-agent": c.fullUserAgent("equinix/ecx-go"),
+		"User-agent": c.fullUserAgent(c.UAPrefix, "equinix/ecx-go"),
 	})
 	neClient.SetHeaders(map[string]string{
-		"User-agent": c.fullUserAgent("equinix/ne-go"),
+		"User-agent": c.fullUserAgent(c.UAPrefix, "equinix/ne-go"),
 	})
 
 	c.ecx = ecxClient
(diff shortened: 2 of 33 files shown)


equinix/data_source_metal_project_acc_test.go
```diff
@@ -38,6 +38,12 @@ func TestAccDataSourceMetalProject_basic(t *testing.T) {
 
 func testAccDataSourceMetalProject_basic(r string) string {
 	return fmt.Sprintf(`
+terraform {
```
displague, line 41: I didn't see evidence of the module_name making it to the Provider Configure in this PR.
```
export TF_PROVIDER_LOG=debug; export PACKNGO_DEBUG=1; 
make testacc TESTS_REGEXP='(TestAccDataSourceMetalProject_basic)'
```

ocobleseqx: After some tests I confirm that:

- GetProviderMeta returns nil during provider configuration, so a build user-agent function must be included in all create/read/update/delete resource functions
- Only works with resources and not with datasources
- Works fine with both tf files and modules

displague (commented):
In some cases the module_name is being repeated in the User-Agent. I initially thought this would happen when Create and Update call Read, but:

With the following patch:
```patch
diff --git a/equinix/resource_metal_project.go b/equinix/resource_metal_project.go
index f546e2c4..9b6e0eee 100644
--- a/equinix/resource_metal_project.go
+++ b/equinix/resource_metal_project.go
@@ -176,7 +176,7 @@ func resourceMetalProjectRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 	client.UserAgent = userAgent
-
+	fmt.Println("look here !!!!!!!!!!!!!", client.UserAgent)
 	proj, _, err := client.Projects.Get(d.Id(), nil)
 	if err != nil {
 		err = friendlyError(err)
diff --git a/equinix/resource_metal_project_acc_test.go b/equinix/resource_metal_project_acc_test.go
index ebae59cf..040d4e96 100644
--- a/equinix/resource_metal_project_acc_test.go
+++ b/equinix/resource_metal_project_acc_test.go
@@ -391,6 +391,11 @@ resource "equinix_metal_project" "foobar" {
 
 func testAccMetalProjectConfig_basic(r int) string {
 	return fmt.Sprintf(`
+	terraform {
+		provider_meta "equinix" {
+			module_name = "test"
+		}
+	}
 resource "equinix_metal_project" "foobar" {
     name = "tfacc-project-%d"
 }`, r)
```

I see the following:
```
=== CONT  TestAccMetalProject_importBasic
2022/09/29 20:10:54 [DEBUG] POST https://api.equinix.com/metal/v1/projects
look here !!!!!!!!!!!!! test test HashiCorp Terraform/1.3.0 (+https://www.terraform.io) Terraform Plugin SDK/2.9.0 terraform-provider-equinix/dev packngo/(devel)
2022/09/29 20:10:55 [DEBUG] GET https://api.equinix.com/metal/v1/projects/{id}
```

ctreatma (approved):

Issue #133, opened 2022-05-05
Implement Metal User resource and datasources


displague:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

The Equinix Metal API allows for fetch organization users and creating new users.  The Terraform implementation does not currently cover these APIs: https://metal.equinix.com/developers/api/users/

### New or Affected Resource(s)

* ` equinix_metal_users` datasource (lists users in the organizations/projects that the requesting user has access to via `GET /users`, ideally offering organization or project id filtering if possible)
* ` equinix_metal_user` datasource (fetches details for a single user `GET /users/{id}`)
* `equinix_metal_user` resources (reads like datasource, updates via `PUT /users/{id}`, creates via `POST /users`). These resources can not be deleted.

### Potential Terraform Configuration

The experience around editing ones own user profile could be solved by creating a secondary set of resources (`equinix_metal_profile`?) that return and update the "current" user (`GET /user`, `PUT /user`).

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
```

### References

Relates to https://github.com/equinix/terraform-provider-equinix/issues/175

PR #256, opened 2022-09-28
add equinix_metal_organization_member resource

### LINKED ISSUE #175: Implement organization membership resources
The Equinix Metal API offers organization member endpoints.

Organizations have members:
* `GET /organization/{id}/members`

New members can be added:
* `POST /organization/{id}/invitations`
* Can these invitations be updated or deleted by the org owner?

Memberships can be deleted:
* `DELETE /memberships/{id}`

Memberships can be updated:
* `PUT /memberships/{id}`  (`role` can be changed. Is `project_ids` available but undocumented for changes to `collaborator` or for changing collaborator `project_ids`? Network Inspect the Portal.)

Organization members can be assigned projects and roles on creation.

Roles include `admin`, `collaborator`, and `billing`. Additional roles may be added later.

What does this look like:
```hcl
resource "metal_organization_member" "marques" {
  organization_id = "" # required
  roles = ["admin", "collaborator", "billing"] # required.
  # What does it mean to have multiple roles?

  invitation_message = ""

  # project_ids is included in the invitation, but is also returned in the organization membership response
  project_ids = ["", ""] # only used/needed for 'collaborators'

  # user is returned but we don't need to include all of that in this resource
  # presenting user.email as user_email fits our TF pattern in other resources.
  # user_email would act as the "invitee" in the invitation field.
  # Does this create a problem if the invitee email differs from the primary user email?
  user_email = "" # required

  # id - memberships have ids
}
```

The single result data source would query on `user_email` and `organization_id`, or by `id` / `membership_id`.

```hcl
data "metal_organization_member" "member" {
  # both of:
  # organization_id
  # user_email = "email@example.com"

  # or:  
  # membership_id
}
```

The attributes would include roles, `project_ids`, `user_email`, `organization_id`, `membership_id`. Should we include other user properties, or can we offer a `metal_user` data source to query users? 
Can organization or project members fetch user records for fellow team members?

Should we also provide a multiple resource datasource:

```hcl
data "metal_organization_members" "member_list" {
  # organization_id -- optional with project_id, should this default to some organization like metal_projects does?
  # project_id -- if specified, we can use /projects/{id}/memberships instead of /organizations/{id}/members
  # user_email_filter = "*@gmail.com" ? should we offer a filter?
}
```

displague:
Fixes #175 

Depends on packethost/packngo#348


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/go-retryablehttp v0.6.6
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.9.0
-	github.com/packethost/packngo v0.26.0
+	github.com/packethost/packngo v0.27.1-0.20220928224154-65e26866e234
```
displague, line 15: Once https://github.com/packethost/packngo/pull/348 is merged, `go.sum` must be updated and this line is expected to be:
```suggestion
	github.com/packethost/packngo v0.28.0
```

displague (commented):
I haven’t tested the "existing member" flow.

I think it should be possible to do that by using an import test against the current user, and we can assume it is an admin since it would have created the organization.

displague:
> I think it should be possible to do that by using an import test against the current user, and we can assume it is an admin since it would have created the organization.

Testing against the current owner would cause Terraform to attempt to delete the current owner, which would not be possible.

displague:
I've added some tests against the owner if only to ensure that the existing member path works for initial reads. In general users should not attempt to create primary owners as members.  Additionally, there may be some bugs with how existing members are being handled. It would help if we had #133 and a way to fetch the current user, alternatively, if the organization primary_owner were made available in the metal_organization resource, that could be helpful.

docs/resources/equinix_metal_organization_member.md
```diff
@@ -0,0 +1,58 @@
...
+
+* `id` - The unique ID of the membership.
+* `nonce` - The nonce for the invitation (only known in the invitation stage)
+* `created` - When the invitation was created (only known in the invitation stage)
+* `updated` - When the invitation was updated (only known in the invitation stage)
+* `state` - The state of the membership ('invited' when an invitation is open, 'active' when the user is an organization member)
+
+## Import
+
+This resource can be imported using the `invitee` defined in the resource:
```
displague, line 54: This isn't correct. The import should be in the form: `invitee:organization_id`

equinix/resource_metal_organization_member.go
```diff
@@ -0,0 +1,270 @@
...
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				// TODO: Update should be supported. packngo.InvitationService does not offer an Update	method.
+				ForceNew: true,
+			},
+			"state": {
+				Type:        schema.TypeString,
+				Description: "The state of the membership ('invited' when an invitation is open, 'active' when the user is an organization member)",
+				Computed:    true,
+			},
+		},
```
displague, line 108: nit: the `message` field was not included. I'm honestly not sure if this even appears in the emailed invite or anywhere else.

equinix/resource_metal_organization_member.go
```diff
@@ -0,0 +1,277 @@
...
+			"roles":           stringArrToIfArr(member.Member.Roles),
+			"projects_ids":    stringArrToIfArr(projectIDs),
+			"organization_id": member.Member.Organization.ID,
+		})
+	} else if member.isInvitation() {
+		projectIDs := []string{}
+		for _, project := range member.Invitation.Projects {
+			projectIDs = append(projectIDs, path.Base(project.Href))
+		}
+		return setMap(d, map[string]interface{}{
```
ocobleseqx, line 207: If `invited_by` is computed I expect it here in the setMap

displague: invited_by is a user in the database. We may want to change this field to a user_id value.

displague: updated to user_id in 9eee41d

equinix/resource_metal_organization_member.go
```diff
@@ -0,0 +1,275 @@
...
+		// If the org was destroyed, mark as gone.
+		if isNotFound(err) {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+	member, err := findMember(invitee, members, invitations)
+	if !d.IsNewResource() && err != nil {
+		return friendlyError(fmt.Errorf("Could not find member %s in organization %s", invitee, orgID))
```
ocobleseqx, line 185: If member leave the project we don't want an error but a WARN and remove the id from the state

displague: fixed in 98a070f

ocobleseqx (approved):
good work!

ocobleseqx:
Related test passed successfully

PR #261, opened 2022-11-08
Remove optional flag for address fields in organization data source

### LINKED ISSUE #151: ds equinix_metal_organization address field wrong schema behavoir
> `Optional: true` here (and in createOrganizationAddressDataSourceSchema) implies that address fields can be used as search filters. For the datasource, Optional should be false. 
> 
> For the resource, I think the address fields should not be computed, they should be whatever the user has defined in Terraform.
> 
> _Originally posted by @displague in https://github.com/equinix/terraform-provider-metal/pull/231#discussion_r896846711_

ctreatma:
The organization data source does not support filtering organizations by address, so the address field (and all nested fields within it) should not have the `Optional` flag set.  This removes the flag so that the organization data source schema matches the behavior.

Fixes #151.

### DIFF
--- a/equinix/data_source_metal_organization.go
+++ b/equinix/data_source_metal_organization.go
@@ -57,7 +57,6 @@ func dataSourceMetalOrganization() *schema.Resource {
 				Type:        schema.TypeList,
 				Description: "Business' address",
 				Computed:    true,
-				Optional:    true,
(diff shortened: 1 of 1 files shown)


ocobleseqx (approved):
lgtm 🚀

PR #262, opened 2022-11-08
Force new metal connection when VLANs attribute is changed

### LINKED ISSUE #222: equinix_metal_connection field vlans should be forcenew
The `vlans` field of `equinix_metal_connection` can not be updated. If a user changes this field the resource must be recreated.
What happens today is the change is silently updated in state without making remote changes.

https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_connection.go#L141

ctreatma:
The `vlans` field of an `equinix_metal_connection` resource cannot be changed, but the Equinix provider currently attempts to update the `vlans` attribute via the API, and the API silently ignores the change, giving clients the illusion that they have successfully updated the `vlans` field.  This adds the `ForceNew` flag for the `vlans` attribute so that the terraform provider will correctly destroy & recreate an `equinix_metal_connection` when the `vlans` attribute changes.

Fixes #222.

Closes #273

### DIFF
--- a/equinix/resource_metal_connection.go
+++ b/equinix/resource_metal_connection.go
@@ -144,6 +144,7 @@ func resourceMetalConnection() *schema.Resource {
 				Optional:    true,
 				Elem:        &schema.Schema{Type: schema.TypeInt},
 				MaxItems:    2,
+				ForceNew:    true,
 			},
 			"service_token_type": {
 				Type:        schema.TypeString,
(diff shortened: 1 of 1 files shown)


ctreatma:
*Note:* this is an alternative to #273 and only one of them should be adopted.  If that PR gets merged, close this one without merging (and vice versa).

ocobleseqx:
Closed in favor of the #273 alternative, letting update VLANs attribute

PR #271, opened 2022-11-23
Keep value of metal_connection.vlans in sync with API

ctreatma:
This updates the `metal_connection` resource and data source to read the value of `vlans` from the API response to ensure that terraform state stays up-to-date with the real infrastructure if someone modifies the VLANs for a terraform-managed `metal_connection` outside of terraform.

Closes #270

PR #273, opened 2022-11-23
Update VLANs for shared metal connections

### LINKED ISSUE #272: Support updating VLANs for a shared `metal_connection`
### Community Note

### Description

Currently, if a user tries to change the `vlans` attribute for a shared `metal_connection`, the terraform state is updated, but the infrastructure is not.  The `vlans` for a shared `metal_connection` are attached to "hidden" virtual circuits that are created automatically by the API.  These virtual circuits cannot be managed with the `metal_virtual_circuit` resource because they are created implicitly, and if they are imported to terraform, they cannot be deleted.

* equinix_metal_connection

```tf
```

### References

This issue was created out of the discussion on #222.  That issue proposes forcing recreation of shared `metal_connection` resources when the `vlans` attribute is changed via terraform, which exposes customers who use terraform to more risk.

ctreatma:
This updates the `metal_connection` resource so that, when the `vlans` attribute is updated for a shared connection, those changes are sent to the Equinix Metal API.

Closes #272
Closes #262


equinix/resource_metal_connection.go
```diff
@@ -340,9 +343,59 @@ func resourceMetalConnectionUpdate(d *schema.ResourceData, meta interface{}) err
...
+
+		if connType == packngo.ConnectionShared {
+			old, new := d.GetChange("vlans")
+			oldVlans := convertIntArr2(old.([]interface{}))
+			newVlans := convertIntArr2(new.([]interface{}))
+			maxVlans := int(math.Max(float64(len(oldVlans)), float64(len(newVlans))))
+
+			ports := d.Get("ports").([]interface{})
+
+			for i := 0; i < maxVlans; i++ {
```
ctreatma, line 359: This approach addresses most of the concerns expressed in https://github.com/equinix/terraform-provider-equinix/pull/273#discussion_r1030831300:
1. Removing a VLAN from the `vlans` attribute (i.e., changing vlans: `[a, b]` to `vlans: [a]` ) will cause `b` to be unassigned, as expected
2. Removing the first VLAN from `vlans` attribute (i.e., changing `vlans: [a, b]` to `vlans: [b]`) will cause `b` to be unassigned from the secondary port and then reassigned to the primary port, as expected
3. Swapping the VLANs (i.e., changing `vlans: [a, b]` to `vlans: [b, a]`) will cause `b` to be unassigned from the secondary port and reassigned to the primary port, but assigning `a` to the secondary port will fail with a 422 error because the circuit is still deactivating.  Re-running `terraform apply` will complete the swap.  I'm not convinced we should worry about swapping, since it's not easily supported anywhere else.

ctreatma:
#271 needs to be merged before this one in order for us to implement acceptance tests around this change

equinix/resource_metal_connection.go
```diff
@@ -340,9 +343,59 @@ func resourceMetalConnectionUpdate(d *schema.ResourceData, meta interface{}) err
 			return friendlyError(err)
 		}
 	}
+
+	// Don't update VLANs until _after_ the main ConnectionUpdateRequest has succeeded
+	if d.HasChange("vlans") {
+		connType := packngo.ConnectionType(d.Get("type").(string))
+
+		if connType == packngo.ConnectionShared {
+			old, new := d.GetChange("vlans")
```
ctreatma, line 352: @ocobleseqx @displague I have a local commit that updates this code to throw an error if `status` is `pending`, but that change causes the acceptance test to fail.  Without the `status` check, the acceptance test passes, so I _think_ that indicates that the API is correctly accepting the VLANs updates for `pending` shared connections.  I'm hesitant to update the tests to handle the `status` check if everything is working without the `status` check.

ocobleseqx: I checked it again using the API and indeed if it can be updated even it is still `pending`, I don't know what was wrong with my prev tests

ocobleseqx (approved):

ocobleseqx:
💯❗  @ctreatma

PR #287, opened 2023-01-17
NFV-22886: add cloudInitFile field for device creation request and use new file upload API to upload the file

rling-equinix:
In Dec release, we onboarded Aviatrix as our new vendor. It requires the customer to upload cloud init file and do some basic file processing from the file. Hence, new file upload API is introduced to handle such file processing and this new file type. Also, in the device creation flow, new field cloudInitFileId is introduced for the given file. Noted that this new file API can also handle other file type. The processType field is used to identify different file types. More details can be found in the public API documentation.

This PR allows customer to upload such cloud init file through cloud_init_file field in terraform script. It supports both single and HA device request. README file has been updated accordingly to reflect the changes. Two new device type (C8KV router and Aviatrix) scripts have been provided as well for reference. 

Besides, since we no longer support Equinix managed device any more and encourage customers to use BYOL license mode, the default value for self-managed and byol field has been changed to be both true. 

reference: [NE API Documentation](https://developer.equinix.com/docs?page=/dev-docs/ne/overview)


docs/resources/equinix_network_device.md
```diff
@@ -102,6 +104,61 @@ resource "equinix_network_device" "panw-cluster" {
...
+  type_code       = "AVIATRIX_EDGE"
+  self_managed    = true
+  byol            = true
+  package_code    = "STD"
+  notifications   = ["john@equinix.com"]
+  term_length     = 6
+  account_number  = data.equinix_network_account.sv.number
+  version         = "6.9"
+  core_count      = 2
+  cloud_init_file = "cloudInitFileFolder/AVX-cloud-init-file.txt"
```
ctreatma, line 126: Rather than taking a `cloud_init_file` parameter that contains the path to a local file, we could take a `cloud_init` parameter that has the contents of the cloud-init. Expecting that value means that a user can load the contents from a file on disk _or_ use a template to generate the cloud-init contents.  For example:
```hcl
resource "equinix_network_device" "aviatrix-with-cloudinit-from-file" {
   # ... other resource properties
   cloud_init = file("cloudInitFileFolder/AVX-cloud-init-file.txt")
}

resource "equinix_network_device" "aviatrix-with-cloudinit-from-template" {
   # ... other resource properties
   cloud_init = templatefile("cloud-init-file.tpl", {
      my_template_variable = "some_value"
   })
}
```

equinix/resource_network_device.go
```diff
@@ -1508,6 +1553,28 @@ func uploadDeviceLicenseFile(openFunc openFile, uploadFunc uploadLicenseFile, ty
 	return nil
 }
 
+func uploadDeviceCloudInitFile(openFunc openFile, uploadFunc uploadFile, typeCode string, device *ne.Device) error {
+	if device == nil || ne.StringValue(device.CloudInitFile) == "" {
```
ctreatma, line 1557: Taking the cloud-init contents as a parameter, instead of the file name, would enable us to remove this file-loading code.

equinix/resource_network_device.go
```diff
@@ -298,7 +302,7 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 		neDeviceSchemaNames["IsBYOL"]: {
 			Type:        schema.TypeBool,
 			Optional:    true,
-			Default:     false,
+			Default:     true,
```
ctreatma, line 305: Are customers currently using this resource?  I'm a bit concerned about changing this default, especially since this attribute is `ForceNew`, so if a customer fails to notice the change, they could accidentally delete something.

ctreatma (changes requested):

equinix/resource_network_device.go
rling-equinix: Yes, they are using the resource to create the device. Since we will publish a new version for terraform provider with the new change, the customer should know this change of the default value. Like I mentioned in the description of the PR, for all the current supporting vendors, we no longer offer the Equinix managed option, same for this license mode, BYOL will be the major one (Only limited device types would have subscription option). It would be good to change the default value to be self configured and BYOL. If they specify the field in their script, they should not see any changes. If not, terraform would at least prompt the changes to warn the customer. 

If a customer uses the previous version of terraform provider to provision the device, are they allowed to upgrade the provider version to manage the existing deployment? This is the only situation that will cause the problem you mentioned above. Wish to hear more advice for the scenario.

docs/resources/equinix_network_device.md
rling-equinix: Hi @ctreatma, your idea is good here. You are trying to provide a general way to allow the customer dynamically change the file content and upload it through the template. However, in this use case, we don't allow the customer to change the file contents. Those cloud init files will be downloaded directly from vendor and the customer only needs to upload the file. No changes are required and allowed here for such file uploading. I am trying to do it in the same way as we are doing for uploading the license file. 
Also, the cloud init file cannot be shared across different devices and vendors. Such template behavior is currently offered through vendor now. The customers go to the vendor site, provide some configs and then download the file directly from them. Then they come back to our platform and upload the exact file they got from vendor.

equinix/resource_network_device.go
ctreatma: It is possible for a customer to automatically pull in the new provider version for an existing deployment; it depends on how they have set up their [provider version constraints](https://developer.hashicorp.com/terraform/language/providers/requirements#version-constraints).  We may be able to reduce the likelihood of a customer automatically pulling in this default value change by bumping the major version instead of the minor, but since the terraform provider encompasses multiple resources across multiple services, we may not want to do a major version bump in order to change a single default value.

docs/resources/equinix_network_device.md
displague: It sounds like we're missing datasources and resources to reflect NE file management APIs:
* https://github.com/equinix/ne-go/blob/495d265a2b23c8cc255cfc6205c3cadb8785d87f/rest_file.go#L17-L20
* https://github.com/equinix/ne-go/blob/826e3e4c3868a58e09ce291ee28e991f6af1040b/rest_licensefile.go#L10-L13
(Interestingly, the "/file" endpoint seems from this code to support a 'type:licence', perhaps this endpoint alone could be used?)

With that pattern, a user might approach this as such:

```hcl
variable "filename" { default = "cloudInitFileFolder/AVX-cloud-init-file.txt" }

resource "equinix_network_device_file" "aviatrix-with-cloudinit" {
   filename = var.filename
   cloud_init = file(var.filename)
   process_type = "CLOUD_INIT"
   # metro_code
   # device_type_code
   # license_type
   # device_management_type
}

resource "equinix_network_device" "aviatrix-with-cloudinit-from-file" {
   # ... other resource properties
   cloud_init = equinix_network_device_file.aviatrix-with-cloudinit.filename
}
```

I have reservations about direct local file management within the Terraform provider, although I can't locate documentation to substantiate that.  Possible concerns are around error handling, file permissions, and the availability of the files in shared environments. I can subside those concerns with awareness of complimentary patterns including Hashicorp's own `local` file provider (https://github.com/hashicorp/terraform-provider-local/blob/main/internal/provider/data_source_local_file.go#L55-L89, https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/file) and provider config properties that reference local configuration files for keys and certificates. There may also be existing patterns in the NE TF resources that establish this practice (especially around license files).

As we consider adding yet another file reference parameter, I do wonder if we could make space for file management resources in the future, perhaps by including `_filename` in a property that will utilize the /ne/v1/files reference and omitting that suffix (or using an alternative suffix) for a property that includes local file handling.  As @ctreatma expressed, I would have expected this resource to accept direct file contents with no correlation to local filenames.  I can certainly see us deprecating the indirect file uploads in the future and asking users to adopt a pattern like the one shown above.

PR #291, opened 2023-01-30
feat: use metal-go for device reads (data source and resource)

ctreatma:
This updates the device data source to use `metal-go` instead of
`packngo` and also updates the device resource to use `metal-go`
for read operations.  As part of this migration, the device data source
and resource gain the new `sos_hostname` attribute which enables
customers to reduce their reliance on the deprecated `facility` attribute,
which was previously necessary in order to compute the SOS hostname
for a device.


equinix/data_source_metal_device.go
```diff
@@ -232,26 +237,28 @@ func dataSourceMetalDeviceRead(d *schema.ResourceData, meta interface{}) error {
...
-		d.Set("metro", strings.ToLower(device.Metro.Code))
+		d.Set("metro", strings.ToLower(*device.Metro.Code))
 	}
-	d.Set("operating_system", device.OS.Slug)
+	d.Set("operating_system", device.OperatingSystem.Slug)
 	d.Set("state", device.State)
 	d.Set("billing_cycle", device.BillingCycle)
-	d.Set("ipxe_script_url", device.IPXEScriptURL)
-	d.Set("always_pxe", device.AlwaysPXE)
+	d.Set("ipxe_script_url", device.IpxeScriptUrl)
```
displague, line 257: Little changes like these in properties between Packngo and MetalGo make me wonder if we have the 'better' pattern in one project or the other. If "ID" "PXE" "URL" are preferable, the codegen in metal-go could consider take that route with initialisms (https://github.com/OpenAPITools/openapi-generator/issues/4833 has an associated stale/open PR)

equinix/data_source_metal_device.go
```diff
@@ -266,45 +273,49 @@ func dataSourceMetalDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if device.HardwareReservation != nil {
-		d.Set("hardware_reservation_id", device.HardwareReservation.ID)
+		// Device schema needs to be updated to include hardware reservation attributes
+		//d.Set("hardware_reservation_id", device.HardwareReservation.ID)
```
displague, line 277: This may be corrected by https://github.com/equinix/terraform-provider-equinix/pull/291/files#r1126575481

equinix/data_source_metal_device.go
```diff
@@ -232,26 +237,28 @@ func dataSourceMetalDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	} else {
 		deviceId := deviceIdRaw.(string)
 		var err error
-		device, _, err = client.Devices.Get(deviceId, deviceReadOptions)
+		device, _, err = client.DevicesApi.FindDeviceById(nil, deviceId).Execute() // deviceReadOptions ?
```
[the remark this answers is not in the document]
ctreatma: Includes have no impact here because the spec says that this comes back as an `Href`, so no matter what includes we send, `device.Project` always be an `Href`.  We cannot get this working nicely without improving the devices spec.  We've talked about adding `XOrHref` models across the board for this kind of issue, but IMO the experience of using those in code is not great; I have examples of that and of alternatively including `href` attributes directly in models, but haven't taken the time to open draft PRs in `metal-go` with those changes yet.  When they're up I will cross-link them with this PR.

equinix/data_source_metal_device.go
ctreatma: If/when the generator has broad support for defining acronyms/initialisms, it may be worth looking into proper capitalization of those.  At the moment, the camel-casing is consistent across `metal-go`, `metal-java`, and `metal-python`, so I'd be hesitant to add a bunch of post-patches in one of those (or, even worse, maintain equivalent post-patches for all of them) to change that.

Casing decisions for model names are driven by the spec, and I think that's where we could use more consistency; here's an example where BGP is capitalized differently because of the spec:

- https://github.com/equinix-labs/metal-go/blob/7e28dc6f080125526593528336c15de56e0e99bf/metal/v1/model_bgp_session.go
- https://github.com/equinix-labs/metal-go/blob/7e28dc6f080125526593528336c15de56e0e99bf/metal/v1/model_bgp_session_input.go

At this point, I'd probably lean towards camel-casing acronyms & initialisms in model names in the spec since that's what the generator would do for property names that contain the same acronyms & initialisms.

equinix/data_source_metal_device.go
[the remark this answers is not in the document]
ctreatma: Draft PRs with the two modeling approaches mentioned above:

displague:
#312
> Equinix Metal's use of Deprecation and Sunset headers should be surfaced through Terraform output or logs

It's worth exploring wether we have good access to response headers and a generic method of adding any response Sunset and Deprecation headers to the TF log as warnings. https://developer.hashicorp.com/terraform/tutorials/providers/provider-debug#update-error-messages

equinix/data_source_metal_device.go
displague: Looks like initialism may be fixed. 
https://github.com/OpenAPITools/openapi-generator/pull/15083

equinix/data_source_metal_device.go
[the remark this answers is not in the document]
ctreatma: The spec has been fixed and these properties are now accessible in `metal-go`.

equinix/data_source_metal_device.go
```diff
@@ -197,12 +198,17 @@ func dataSourceMetalDevice() *schema.Resource {
 					},
 				},
 			},
+			"sos_hostname": {
```
ctreatma, line 201: The device resource will gain the `sos_hostname` attribute in a separate PR to prevent this one from getting even larger than it already is.

ctreatma:
> #312
> 
> > Equinix Metal's use of Deprecation and Sunset headers should be surfaced through Terraform output or logs
> 
> It's worth exploring wether we have good access to response headers and a generic method of adding any response Sunset and Deprecation headers to the TF log as warnings. https://developer.hashicorp.com/terraform/tutorials/providers/provider-debug#update-error-messages

This PR has been upgraded to `v0.10.0` of `metal-go`, which replicates `packngo` behavior of dumping `Deprecation` and `Sunset` headers & related links to stdout.  It's probably still worth looking into exposing that information in a different, possibly tool-specific way, but I think doing that here would be adding unnecessary scope to the conversion.

equinix/data_source_metal_device.go
ctreatma: I added equinix-labs/metal-go#112 to investigate the `reservedWordsMapping` and see if it supports initialisms without introducing unintended side effects.  Given that casing for initialisms is a stylistic issue rather than a functional one, I think it would be ok to adopt metal-go in this project without waiting for the results of that SDK investigation.

displague:
This should be rebased now that #353 is merged.

ctreatma:
BGP data source changes were split out to #368 

This could benefit from the helper methods introduced in #366

equinix/data_source_metal_device.go
ctreatma: As an alternative, I've updated this PR to convert the `Read` pathway for the device resource to use `metal-go`.  This enables the resource and the data source to gain the SOS hostname at the same time.  The create/update/delete pathways for the device resource can be converted later in a separate PR.

cprivitere: Seems ok to me.

cprivitere (dismissed):

displague:
> This could benefit from the helper methods introduced in https://github.com/equinix/terraform-provider-equinix/pull/366

Those are available now.

ctreatma:
I've rebased on the latest from `main` and consolidated some helpers that were duplicated across multiple PRs.  Since this PR converts both resource & data source reads, we no longer need the `MetalGo` suffix on helpers.

I've opted not to adopt the `getDeviceMap` helper in this PR because it doesn't cover all of the attributes that are needed in the device resource & data source; while this is an issue that needs to be fixed, I think it will be a big enough change that it will be easier to implement & review that in a separate PR.

equinix/helpers_device.go
```diff
@@ -68,64 +69,46 @@ type NetworkInfo struct {
 	PrivateIPv4    string
 }
 
-func getNetworkInfoMetalGo(ips []metalv1.IPAssignment) NetworkInfo {
+func getNetworkInfo(ips []metalv1.IPAssignment) NetworkInfo {
```
ctreatma, line 72: I cleaned up the unnecessary MetalGo suffixes from helper methods, but it does make some of these changes harder to review.  I can back out the name change if that's preferred, but I'm going to instead try to annotate the current changes to guide review.

equinix/helpers_device.go
```diff
@@ -68,64 +69,46 @@ type NetworkInfo struct {
...
 	ni := NetworkInfo{Networks: make([]map[string]interface{}, 0, 1)}
 	for _, ip := range ips {
 		network := map[string]interface{}{
-			"address": ip.Address,
-			"gateway": ip.Gateway,
-			"family":  ip.AddressFamily,
-			"cidr":    ip.Cidr,
-			"public":  ip.Public,
+			"address": ip.GetAddress(),
```
ctreatma, line 76: This is replacing direct references to the `ip` struct properties with calls to the `Get...()` functions.  The struct properties are pointers and the `Get...()` functions dereference the pointers (returning the zero value if the pointer is nil), so the `Get...()` functions line up with how the previous `packngo` implementation worked.

equinix/helpers_device.go
```diff
@@ -68,64 +69,46 @@ type NetworkInfo struct {
...
 		ni.Networks = append(ni.Networks, network)
 
 		// Initial device IPs are fixed and marked as "Management"
-		if *ip.Management {
-			if *ip.AddressFamily == 4 {
-				if *ip.Public {
-					ni.Host = *ip.Address
-					ni.IPv4SubnetSize = int(*ip.Cidr)
-					ni.PublicIPv4 = *ip.Address
+		if ip.GetManagement() {
```
ctreatma, line 85: This is replacing pointer dereferences with `Get...()` calls for better protection against nil pointer errors.

equinix/helpers_device.go
```diff
@@ -68,64 +69,46 @@ type NetworkInfo struct {
...
-					ni.IPv4SubnetSize = ip.CIDR
-					ni.PublicIPv4 = ip.Address
-				} else {
-					ni.PrivateIPv4 = ip.Address
-				}
-			} else {
-				ni.PublicIPv6 = ip.Address
-			}
+func getNetworkType(device *metalv1.Device) (*string, error) {
+	pgDevice := packngo.Device{}
```
ctreatma, line 103: The `NetworkType` device attribute exposed by `packngo` is not an API-provided attribute; this serves as a shim so we can use the `packngo` logic to calculate `NetworkType`.  We could instead either duplicate the complete `packngo` logic here or copy it into `metal-go`, but my understanding is that `NetworkType` is not something we want to keep around anyway, so this shim is designed to get us through the SDK transition and then be eliminated when `packngo` is no longer used in the provider.

equinix/helpers_device.go
```diff
@@ -140,30 +123,15 @@ func getNetworkRank(family int, public bool) int {
...
+func getPorts(ps []metalv1.Port) []map[string]interface{} {
 	ret := make([]map[string]interface{}, 0, 1)
 	for _, p := range ps {
 		port := map[string]interface{}{
-			"name":   p.Name,
-			"id":     p.ID,
-			"type":   p.Type,
-			"mac":    p.Data.MAC,
-			"bonded": p.Data.Bonded,
+			"name":   p.GetName(),
```
ctreatma, line 130: As with `getNetworkInfo`, this change replaces the innards of `getPorts` with those of `getPortsMetalGo` while also updating the function to use `Get...()` helpers, since those helpers more closely match the behavior of the `packngo` implementation.

displague:
```
panic: interface conversion: interface {} is int32, not *int32

goroutine 61190 [running]:
github.com/equinix/terraform-provider-equinix/equinix.getDeviceMap.func1(0x1, 0x0)
	/home/runner/work/terraform-provider-equinix/terraform-provider-equinix/equinix/helpers_device.go:244 +0x348
sort.insertionSort_func({0xc000568d10?, 0xc00250b260?}, 0x0, 0x3)
	/opt/hostedtoolcache/go/1.20.8/x64/src/sort/zsortfunc.go:12 +0xb1
sort.stable_func({0xc000568d10?, 0xc00250b260?}, 0x3)
	/opt/hostedtoolcache/go/1.20.8/x64/src/sort/zsortfunc.go:343 +0x7a
sort.SliceStable({0x11f8900?, 0xc0028b5290?}, 0x4?)
	/opt/hostedtoolcache/go/1.20.8/x64/src/sort/slice.go:38 +0xe6
github.com/equinix/terraform-provider-equinix/equinix.getDeviceMap({0xc001c4f0ec, 0xc002ce0820, 0xc001c4f0e8, 0xc003012090, 0xc000c76690, 0xc000673080, 0x0, 0xc0009f73e0, 0x0, 0xc002ce0810, ...})
	/home/runner/work/terraform-provider-equinix/terraform-provider-equinix/equinix/helpers_device.go:243 +0x14a
github.com/equinix/terraform-provider-equinix/equinix.flattenDevice({0x14027a0?, 0xc00106f4a0?}, {0x141359a?, 0x6?}, 0xc000fbbc80?)
	/home/runner/work/terraform-provider-equinix/terraform-provider-equinix/equinix/data_source_metal_devices.go:96 +0x159
github.com/equinix/terraform-provider-equinix/equinix/internal/datalist.dataListResourceRead.func1({0x160af40?, 0xc000a7dec0?}, 0x0?, {0x1371b20, 0xc00230db00})
	/home/runner/work/terraform-provider-equinix/terraform-provider-equinix/equinix/internal/datalist/schema.go:99 +0x2ca
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc0003ce7e0, {0x160af40, 0xc000a7dec0}, 0xd?, {0x1371b20, 0xc00230db00})
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:724 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).ReadDataApply(0xc0003ce7e0, {0x160af40, 0xc000a7dec0}, 0xc003010200, {0x1371b20, 0xc00230db00})
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:943 +0x145
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadDataSource(0xc0028e49c0, {0x160af40?, 0xc000a7de00?}, 0xc0022cd280)
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/grpc_provider.go:1195 +0x38f
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ReadDataSource(0xc003d54aa0, {0x160af40?, 0xc000a7d320?}, 0xc0039c0e60)
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.14.3/tfprotov5/tf5server/server.go:658 +0x403
github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5._Provider_ReadDataSource_Handler({0x13aa6a0?, 0xc003d54aa0}, {0x160af40, 0xc000a7d320}, 0xc003bb4a10, 0x0)
	/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.14.3/tfprotov5/internal/tfplugin5/tfplugin5_grpc.pb.go:421 +0x170
google.golang.org/grpc.(*Server).processUnaryRPC(0xc0031ac3c0, {0x160f838, 0xc000930340}, 0xc000ccc7e0, 0xc0018f8720, 0x1e53710, 0x0)
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.55.0/server.go:1337 +0xdf3
google.golang.org/grpc.(*Server).handleStream(0xc0031ac3c0, {0x160f838, 0xc000930340}, 0xc000ccc7e0, 0x0)
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.55.0/server.go:1714 +0xa36
google.golang.org/grpc.(*Server).serveStreams.func1.1()
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.55.0/server.go:959 +0x98
created by google.golang.org/grpc.(*Server).serveStreams.func1
	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.55.0/server.go:957 +0x18c
	```

**Comment by ctreatma** [comment]
> 🐛 in E2E:
> 

Missed a conversion from pointer derefs to values when collapsing the duplicate helpers; [it has been fixed](https://github.com/equinix/terraform-provider-equinix/pull/291/commits/9a912ca6781b0ad12b3e2f8917bfaf4b21068ac3).

File: equinix/data_source_metal_device.go [hunk 10]
```diff
@@ -299,14 +308,14 @@ func dataSourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta
...
 }
 
-func findDeviceByHostname(devices []packngo.Device, hostname string) (*packngo.Device, error) {
-	results := make([]packngo.Device, 0)
-	for _, d := range devices {
-		if d.Hostname == hostname {
+func findDeviceByHostname(devices *metalv1.DeviceList, hostname string) (*metalv1.Device, error) {
+	results := make([]metalv1.Device, 0)
+	for _, d := range devices.GetDevices() {
+		if *d.Hostname == hostname {
```
displague, line 318: nit: presumably this will never be null in a returned devices list, so this should be fine. `GetHostname()` is an option.

displague (approved):

PR #294, opened 2023-02-07
NFV-22745: introduce new file resource and add cloudInitFileId in device resource to support Aviatrix vendor

rling-equinix:
In Dec release, we onboarded Aviatrix as our new vendor. It requires the customer to upload cloud init file and do some basic file processing from the file. Hence, new file upload API is introduced to handle such file processing and this new file type. Also, in the device creation flow, new field cloudInitFileId is introduced for the given file. Noted that this new file API can also handle other file type. The processType field is used to identify different file types. More details can be found in the public API documentation.

This PR introduces new file resource as discussed in https://github.com/equinix/terraform-provider-equinix/pull/287. It allows customer to upload such cloud init file through file resource and apply that file uuid to the device resource's cloud_init_file_id field in terraform script. For example:
```hcl
variable "filepath" { default = "cloudInitFileFolder/TF-AVX-cloud-init-file.txt" }

resource "equinix_network_file" "aviatrix-cloudinit-file" {
  file_name = "TF-AVX-cloud-init-file.txt"
  content = file("${path.module}/${var.filepath}")
  metro_code = data.equinix_network_account.sv.metro_code
  device_type_code = "AVIATRIX_EDGE"
  process_type = "CLOUD_INIT"
  self_managed = true
  byol = true
}

resource "equinix_network_device" "aviatrix-single" {
  # ... other resource properties
  cloud_init_file_id = equinix_network_file.aviatrix-cloudinit-file.uuid
}
```
License_file_id has also been updated to support similar use cases. 
README file has been updated accordingly to reflect the changes. Two new device type (C8KV router and Aviatrix) scripts have been provided as well for reference.

Besides, since we no longer support Equinix managed device any more and encourage customers to use BYOL license mode, the default value for self-managed and byol field in the file resource has been set to be both true.

reference: [NE API Documentation](https://developer.equinix.com/docs?page=/dev-docs/ne/overview)


rling-equinix:
go.mod and go.sum needs to be updated after https://github.com/equinix/ne-go/pull/19 (go client) is merged.

docs/resources/equinix_network_device.md
```diff
@@ -163,7 +234,9 @@ The `secondary_device` block supports the following arguments:
 * `hostname` - (Optional) Secondary device hostname.
 * `license_token` - (Optional) License Token can be provided for some device types o the device.
 * `license_file` - (Optional) Path to the license file that will be uploaded and applied on a
-secondary device. Applicable for some devices types in BYOL licensing mode.
+secondary device. Applicable for some device types in BYOL licensing mode.
+* `license_file_id` - (Optional) Identifier of a license file that will be applied on a secondary device.
```
ctreatma, line 238: Should a customer be able to specify both `license_file` and `license_file_id`?  If not, we should call that out here in the docs.

docs/resources/equinix_network_file.md
```diff
@@ -0,0 +1,54 @@
...
+## Argument Reference
+
+The following arguments are supported:
+
+* `file_name` - (Required) File name.
+* `content` - (Required) Uploaded file content, expected to be a UTF-8 encoded string.
+* `metro_code` - (Required) File upload location metro code. It should match the device location metro code.
+* `type_code` - (Required) Device type code.
+* `process_type` - (Required) File process type (LICENSE or CLOUD_INIT).
+* `self_managed` - (Optional) Boolean value that determines device management mode, i.e.,
```
ctreatma, line 34: I think the default is `false` for `self_managed`.  I'd recommend specifying the mapping (`true` for self-managed, `false` for Equinix managed).

docs/resources/equinix_network_device.md
rling-equinix: README docs have been updated to reflect that.

docs/resources/equinix_network_file.md
rling-equinix: Mapping is already defined in the code. Docs here mention that it's a boolean value. As discussed, to avoid confusion with the device resource, both self_managed and byol has been marked as required fields. No default will be provided. They are all boolean value similar to what we have in device resource.

ctreatma (approved):

PR #299, opened 2023-02-20
Fix missing projectId API param in resource_network_ssh_key

artraf:


artraf:
dependence: https://github.com/equinix/ne-go/pull/20

displague:
This will be a breaking change, @artraf. Does the API absolutely require the project_id field now or is a default value used if the parameter is omitted? Has this field always been required (and if so, how did that go unnoticed in Terraform)?

I believe we should add a CHANGELOG entry for 1.13.0 to introduce this breaking change.  1.12.0's changelog entry was missed. It can be copied from here: https://github.com/equinix/terraform-provider-equinix/releases/tag/v1.12.0

artraf:
@displague It's can't be omitted. I added a CHANGELOG entry in my last commit. I think this required field appeared after implementation of projects functionality by Eqx.

ctreatma:
After chatting with the NE team, it sounds like this behavior was a bug that has since been resolved, and the `project_id` should not be required for `resource_network_ssh_key`.  Are you still experiencing this issue?

artraf:
@ctreatma yes. In the case of several projects, the project_id is still required. Otherwise, the created key will be assigned to a random project and could be inaccessible

ctreatma:
@rling-equinix can you take a look at this PR as well as https://github.com/equinix/ne-go/pull/20 and let us know if we can move forward with them?

Issue #277, opened 2022-11-30
Can we use overrides to fix tests that are driven by configs in `examples/`


ctreatma:
The tests for the new Fabric resources & data sources use the terraform configs in the `examples/` directory.  This is convenient because we can regularly validate that our examples still work, and we can avoid defining duplicate configurations in the tests themselves.  Problems arise when we introduce tests in this style that cover unreleased resources or data sources.  The `examples/` are intended for end-user consumption, so they pull down the latest official release of our provider, which means that the tests can't pass until the code they're testing has already been released to customers.

Terraform has a notion of override files that can be used to change parts of an existing configuration: https://developer.hashicorp.com/terraform/language/files/override

Can we update our tests to create an override file that uses a locally-built provider instead of downloading the latest publicly-available release?

ctreatma:
Note: the Fabric tests are using `terratest`, so we'll have to dig into how `terratest` works to see if/where we can inject override files into the configuration under test.

https://terratest.gruntwork.io/docs/

displague:
@ctreatma can this be closed now that #684 was merged?

ctreatma:
Yep, this is irrelevant now.  The newer Fabric acceptance tests are not terratest-based, and terratest has been torn out of this repo.

PR #300, opened 2023-02-28
Fix provider requirement for port2serviceprofilepublic example

ctreatma:


ctreatma:
This is an example of the test failure in the `Go Tests` action that is fixed by this change: https://github.com/equinix/terraform-provider-equinix/actions/runs/4298089969/jobs/7491816113#step:6:1345

ocobleseqx (approved):

ocobleseqx:
It is fine to have the example using the official provider, but we will want to run tests against the developing version. That is something that may need some research, perhaps building the binary in GH actions and replacing the definition of the provider to use the development binary

ctreatma:
Yes, longer-term all of the tests that are run via terratest will need to be updated to use a local build of the provider; one approach for that is captured in #277.

Issue #279, opened 2022-12-01
Acceptance test sweeper for Terratest tests


ocobles:
https://github.com/equinix/terraform-provider-equinix/pull/278 
> Test sweeper is [failing](https://github.com/equinix/terraform-provider-equinix/actions/runs/3588142143/jobs/6039220202#step:6:227) after introducing new Terratest acc test of `tests` folder. Terratest does not include [sweeper flags](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/testing.go#L48-L50).
> 
> This change just remove checking of test under `tests` folder but doesn't introduce an alternative solution to prevent failed tests leave active resources, I will open a new issue to find a global solution to cover all acceptance tests

PR #301, opened 2023-02-28
Allow updating some `equinix_metal_device` fields without recreating the instance

### LINKED ISSUE #159: metal_device.userdata should offer a behavioral override to avoid forcenew

ctreatma:
By default, we force recreation of Equinix Metal devices when the `user_data` or `custom_data` attributes change, but this decision only exists within the terraform provider; the Equinix Metal API will happily accept updates to those fields without deleting a device.

Previously we had narrow support to bypass this behavior with a `reinstall` block, but that still involves downtime.  This adds support for `behavior.allow_changes`, which is a list of attributes that are allowed to change _without_ forcing device recreation.

Closes #159

### DIFF
--- a/equinix/resource_metal_device.go
+++ b/equinix/resource_metal_device.go
@@ -427,7 +427,7 @@ func resourceMetalDevice() *schema.Resource {
 									return
 								},
 							},
-							Description: "List of attributes that are allowed to change without recreating the instance",
+							Description: "List of attributes that are allowed to change without recreating the instance. Supported attributes: `custom_data`, `user_data`",
 							Optional:    true,
 						},
 					},
@@ -452,13 +452,10 @@ func reinstallDisabled(_ context.Context, d *schema.ResourceDiff, meta interface
 		return true
 	}
 
+	// To reach this point, the device config had to include a `reinstall` block,
+	// so we can assume all necessary parts of that block are filled in
 	reinstall_list := reinstall.([]interface{})
-	reinstall_config, ok := reinstall_list[0].(map[string]interface{})
-	if !ok {
-		// This should be unreachable (to get here, reinstall attribute had to be specified,
-		// but with an invalid specification...maybe panic?)
-		return true
-	}
+	reinstall_config := reinstall_list[0].(map[string]interface{})
 
 	return !reinstall_config["enabled"].(bool)
 }
@@ -476,14 +473,10 @@ func reinstallDisabledAndNoChangesAllowed(attribute string) customdiff.ResourceC
 				return true
 			}
 
+			// To reach this point, the device config had to include a `behavior`
+			// block, so we can assume all necessary parts of that block are filled in
 			behavior_list := behavior.([]interface{})
-			behavior_config, ok := behavior_list[0].(map[string]interface{})
-
-			if !ok {
-				// This should be unreachable (to get here, behavior attribute had to be specified,
-				// but with an invalid specification...maybe panic?  Or just don't capture & check `ok`?)
-				return true
-			}
+			behavior_config := behavior_list[0].(map[string]interface{})
 
 			allow_changes := convertStringArr(behavior_config["allow_changes"].([]interface{}))
 
@@ -809,22 +802,16 @@ func doReinstall(client *packngo.Client, d *schema.ResourceData, meta interface{
 		reinstall, ok := d.GetOk("reinstall")
 
 		if !ok {
-			// Assume we're here because behavior.allow_changes was set
+			// Assume we're here because behavior.allow_changes was set (not an error)
 			return nil
 		}
 
 		reinstall_list := reinstall.([]interface{})
-		reinstall_config, ok := reinstall_list[0].(map[string]interface{})
-
-		if !ok {
-			// This means a reinstall block was provided but it's invalid, so user's intent
-			// is unclear and we should error (this should maybe be a panic?)
-			return fmt.Errorf("expected reinstall configuration and none available")
-		}
+		reinstall_config := reinstall_list[0].(map[string]interface{})
 
 		if !reinstall_config["enabled"].(bool) {
 			// This means a reinstall block was provided, but reinstall was explicitly
-			// enabled.  Assume we're here because behavior.allow_changes was set
+			// disabled.  Assume we're here because behavior.allow_changes was set (not an error)
 			return nil
 		}
 
--- a/equinix/resource_metal_device.go
+++ b/equinix/resource_metal_device.go
@@ -67,9 +67,9 @@ func resourceMetalDevice() *schema.Resource {
 
 			"operating_system": {
 				Type:        schema.TypeString,
-				Description: "The operating system slug. To find the slug, or visit [Operating Systems API docs](https://metal.equinix.com/developers/api/operatingsystems), set your API auth token in the top of the page and see JSON from the API response",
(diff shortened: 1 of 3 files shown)


equinix/resource_metal_device_acc_test.go
```diff
@@ -400,6 +400,36 @@ func TestAccMetalDevice_IPXEConfigMissing(t *testing.T) {
...
+			{
+				Config: testAccMetalDeviceConfig_allowUserdataChanges(rInt, rs),
+				Check: resource.ComposeTestCheckFunc(
+					testAccMetalDeviceExists(r, &d1),
+					resource.TestCheckResourceAttr(r, "user_data", fmt.Sprintf("#!/usr/bin/env sh\necho 'Allow userdata changes %d'\n", rInt)),
+				),
+			},
+			{
+				Config: testAccMetalDeviceConfig_allowUserdataChanges(rInt+1, rs),
+				Check: resource.ComposeTestCheckFunc(
```
displague, line 423: I think we should be able to test that the device would have been destroyed rather than waiting for the full destroy/create cycle.

https://github.com/hashicorp/terraform-provider-google/issues/12428#issuecomment-1232142860 for example would cause the step to fail with an error that we could compare. 

Perhaps `PlanOnly` would help with that too.

equinix/resource_metal_device.go
```diff
@@ -442,6 +458,42 @@ func shouldReinstall(_ context.Context, d *schema.ResourceDiff, meta interface{}
 	return !reinstall_config["enabled"].(bool)
 }
 
+func shouldReinstallOrAllowChanges(attribute string) customdiff.ResourceConditionFunc {
+	return func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+		reinstall := shouldReinstall(ctx, d, meta)
```
displague, line 463: I got the wrong impression reading this at first. Perhaps renaming `reinstall` to `forceNow` would help with legibility.

equinix/resource_metal_device.go
```diff
@@ -442,6 +458,42 @@ func shouldReinstall(_ context.Context, d *schema.ResourceDiff, meta interface{}
...
+			}
+
+			behavior_config, ok := behavior_list[0].(map[string]interface{})
+
+			// We didn't get a behavior configuration
+			if !ok {
+				return true
+			}
+
+			allow_changes, ok := behavior_config["allow_changes"].([]string)
```
displague, line 489: `allow_changes` - one size fits all - nice.  It may be tricky to keep up with this - whenever we add new ForceNew fields (defaulted to ForceNew or dynamically determined), we'll have to be careful to check against allow_changes.

Would we apply this pattern across all resources? If so, would we create a single reusable `behavior` base schema and a `customdiff.ForceNewIf` helper?

examples/connectivity/v4/port2serviceprofilepublic/terraform.tf
```diff
@@ -1,8 +1,7 @@
 terraform {
   required_providers {
     equinix = {
-      source = "developer.equinix.com/terraform/equinix"
-      version = "9.0.0"
+      source = "equinix/equinix"
```
displague, line 4: I haven't checked to see if this is still necessary, but the reason this pointed to developer.equinix.com/terraform/equinix (at one point, perhaps still) is that [Terratest runs this test](https://github.com/equinix/terraform-provider-equinix/tree/master/tests) using a build of the Terraform plugin that is stashed in the home directory.  If `equinix/equinix` is not overridden in someway (there are other ways), the latest released plugin will be downloaded and the build of the local branch will not be used.

You've been in this area before with #277.
 
#279 may also be related.

ctreatma: This change was meant to stay in #300 and has been rebased out of this PR.  This change brings the `port2serviceprofilepublic` config in line with all other examples; the test was failing in terratest because of the invalid provider specification.

equinix/resource_metal_device.go
```diff
@@ -270,18 +270,18 @@ func resourceMetalDevice() *schema.Resource {
 
 			"user_data": {
 				Type:        schema.TypeString,
-				Description: "A string of the desired User Data for the device",
+				Description: "A string of the desired User Data for the device.  By default, changing this attribute will cause terraform to destroy and recreate your device.  If `reinstall` is specified or `behavior.allow_changes` includes `\"user_data\"`, the device will be updated in-place instead of recreated.",
```
displague, line 273: This documentation is reused in Terraform based providers. I think we'll get more mileage in referring to the code/project as "the provider". 

> changing this attribute will cause _the provider_ to destroy ...

equinix/resource_metal_device_acc_test.go
ctreatma: The comment you linked seems to be an unimplemented proposal to address an issue with detecting when resources are recreated in tests; I didn't see anything actionable in there.

Since this is an acceptance test, I think we benefit from actually applying the plan here: this test confirms both that the device was updated in-place and that the attribute value was updated correctly.

equinix/resource_metal_device.go
ctreatma: For the first cut I aligned with the (incorrect) `shouldReinstall` function name; I've since renamed both functions to have names that more accurately describe what they're doing.

equinix/resource_metal_device.go
ctreatma: IMO we should wait until we're adding a `behavior` to another resource or 2 before generalizing the implementation; if we make this reusable too early we may find that it doesn't suit the real needs that come up later.

equinix/resource_metal_device_acc_test.go
displague: Apparently, I was reading more into it than what the commenter posted.  

```hcl
{
Config: `resource "equinix_metal_device" "foo" {
   behavior { allow_changes = false }
   userdata = "original"  ...
   lifecycle { prevent_destroy = true }
}`
PlanOnly: true
},
{
Config: `resource "equinix_metal_device" "foo" {
   behavior { allow_changes = false }
   userdata = "new"  ...
   lifecycle { prevent_destroy = true }
}`
PlanOnly: true,
ExpectError: regexp.MustCompile(`something about prevent_destroy`),
},
{
Config: `resource "equinix_metal_device" "foo" {
   behavior { allow_changes = false }
   userdata = "new"  ...
}`
PlanOnly: true,
},
```

ocobleseqx (commented):
Apart from a single comment, I just miss the documentation update, it would be nice if you could include it in this PR

equinix/resource_metal_device_acc_test.go
```diff
@@ -625,6 +706,41 @@ resource "equinix_metal_device" "test" {
...
+  hostname         = "tfacc-test-device-%d"
+  plan             = local.plan
+  metro            = local.metro
+  operating_system = local.os
+  billing_cycle    = "hourly"
+  project_id       = "${equinix_metal_project.test.id}"
+  tags             = ["%d"]
+  user_data        = %q
+  custom_data      = %q
+
```
displague, line 727: most (all?) of the metal_device tests include `termination_time` as a stop-gap to flush out test instances in the event that E2E sweeping doesn't pick the devices up.

ctreatma:
> Apart from a single comment, I just miss the documentation update, it would be nice if you could include it in this PR

@ocobleseqx I got our different repos mixed up & thought this one automatically updated docs in response to resource changes.  Docs updates are manual in here or is there a tool I should run?

displague (approved):
Looks good to me. 

The suggestion I made about an alternate testing format was something we could pick up at any time. The fewer full e2e provisions we can get away with and still have accurate testing the better (or at least faster).

PR #302, opened 2023-03-01
Move terratest tests to the acceptance test workflow

ctreatma:
The `terratest` tests are end-to-end tests; the `Go test` workflow is not set up for real API interactions, so including `terratest` tests in that workflow means it will always fail.

As a first pass, this attempts to get the `Go test` workflow passing by excluding `terratest` tests in that workflow.  The acceptance test workflow has also been updated to run the `terratest` tests, but they still fail (although they fail _differently_, which counts for something).


ctreatma:
The `terratest` tests now fail due to `401 Unauthorized` errors (previously they were failing to request a token b/c the `Go test` workflow doesn't include any of the env vars needed for API authentication).

Did our `EQUINIX_API_TOKEN` lose some access it used to have, or did these tests never work in CI?  If they never worked, what is the benefit of running them in CI?

displague:
> Did our EQUINIX_API_TOKEN lose some access it used to have, or did these tests never work in CI? If they never worked, what is the benefit of running them in CI at all?

GH isn't configured with EQUINIX_API_TOKEN (or CLIENT_ID / CLIENT_SECRET, which is what we would probably do here).

The pattern that is used in E2E tests could be made to match only the metal tests where AUTH_TOKEN (metal) is defined.  I thought this was in place (@ocobleseqx). The Fabric PR may have changed that.

Persistent EQUINIX_API_TOKEN values are only used in the sandbox API, which is not made for e2e tests (endpoints return a valid result but there is no state or way to simulate alternate responses). https://developer.equinix.com/environment/sandbox

EQUINIX_CLIENT_ID and EQUINIX_CLIENT_SECRET would need to run full E2E on Fabric and NE resources. We would also need an org and project for Terraform to create and destroy the related resources.

displague:
> The pattern that is used in E2E tests could be made to match only the metal tests where AUTH_TOKEN (metal) is defined. I thought this was in place (@ocobleseqx). The Fabric PR may have changed that.

It looks like the `TF acceptance tests` are filtered to Metal.

I see what you mean now. The `terratest` action will always fail. It could be moved to a workflow that is only triggered manually with CLIENT_ID and CLIENT_SECRET and other required params as inputs to GH. (or the action could be commented out).

ctreatma:
For now I think we should exclude the `terratest` tests from actions entirely, as we do for non-Metal acceptance tests, and add a separate issue to get all of those tests added to & working in CI.

.github/workflows/test.yml
```diff
@@ -17,4 +17,4 @@ jobs:
     - name: Build
       run: go build -v .
     - name: TF tests
-      run: go test -v -cover -parallel 4 ./...
+      run: go test -v -cover -parallel 4 $(go list ./... |grep -v 'tests')
```
displague, line 20: We could rename the tests to `TestExampleFoo` or `TestTerragruntFoo` and then use `go test- skip {Pattern}`, but this seems well enough for now.

displague (approved):

PR #306, opened 2023-03-15
Fixing typo in BGP config

c0dyhi11:
The BGP `deployment_type` states it accepts `private` or `public` but the module actually requires `local` or `global`


displague (approved):
This matches what is validated here: https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_project.go#L83-L88 
and what is documented here: https://deploy.equinix.com/developers/docs/metal/bgp/local-bgp/#enabling-bgp-on-the-project

PR #319, opened 2023-05-11
Point users at the GitHub releases page for release details

ctreatma:
This removes the manual step of updating the changelog in order to cut a new release


cprivitere:
I'm cool with it.

displague:
lgtm

An open question for me (not merge blocking) is wether any Hashicorp tools or the registry _care_ about the CHANGELOG.md file or its format.  It was important for consistency in the days when Hashicorp managed the release of providers.

We've been adopting semantic releases in other projects and if we were to adapt that here we could also require that changes follow the CHANGELOG.md format.

I don't know of any reasons why we can't break away from the format and revisit it later. In fact, I've used the informal "generate release notes" feature in this provider for alpha releases. Nothing broke.

I'm curious what @dcallao and @0xch4z (https://github.com/linode/terraform-provider-linode/pull/235) think.

0xch4z:
> An open question for me (not merge blocking) is whether any Hashicorp tools or the registry care about the CHANGELOG.md file or its format. It was important for consistency in the days when Hashicorp managed the release of providers.

The registry doesn't care and doesn't seem to link to a changelog for providers. I'm not aware of any hashi tools that care about this either.

When we switched to generating changelogs for the Linode provider, "generate release notes" wasn't a thing so we went with [release-drafter](https://github.com/release-drafter/release-drafter) which lets you generate changelog entries based on PR labels (looks like [this](https://github.com/linode/terraform-provider-linode/releases/tag/v1.25.0)). Can't go wrong with either.

dcallao:
@displague, Thanks for the feedback. As @0xch4z pointed out, HashiCorp TF registry will not observe or link to a TF provider changelog. That is indeed the case. The registry will look at the github release tags and use the latest tag to assigned it as the `version` of the provider. Generating release notes in CHANGELOG.md or a release page is completely up to the provider maintainers. :-)

displague (approved):

scripts/changelog-links.sh
```diff
@@ -1,31 +0,0 @@
-#!/bin/bash
```
displague, line 1: Removing this seems fine to me. I don't believe we ever used it. 

GitHub automatically does what this intended too: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls

displague:
Verified there are no other references to CHANGELOG.md in the repo

Issue #130, opened 2022-05-02
Some Metal CI/CD tests are failing on legacy slug use


displague:
The use of older Equinix Metal device plan slugs and facility slugs have started causing CI failures as these plans and locations are phased out and the testing account has lost access to these resources.

Update Equinix Metal Facility and Device plan codes to modern codes:
* Legacy plans include: s1.large.x86, t1.small.x86
* Legacy facility codes include: dfw1, sjc1, ewr1

ocobles:
I think this issue can be closed, with data plans and the fixes introduced by @ctreatma we saw all the tests green

Issue #198, opened 2021-02-02
test TestAccMetalPortVlanAttachment_HybridBasic failures


displague:
From L2 Hybrid basic test:
```
 --- FAIL: TestAccMetalPortVlanAttachment_HybridBasic (319.13s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to metal_port_vlan_attachment.test, provider "metal"
        produced an unexpected new value for was present, but now absent.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
```

```
--- FAIL: TestAccMetalPortVlanAttachment_HybridBasic (298.32s)
    testing.go:569: Step 1 error: errors during apply:
        
        Error: POST https://api.equinix.com/metal/v1/ports/30f932b1-effc-4b09-bf6f-00494581d97f/disbond: 504  Unexpected Content-Type text/html with status 504 Gateway Time-out
        
          on /tmp/tf-test369568634/main.tf line 16:
          (source code not available)
```

_Originally posted by @t0mk in https://github.com/equinix/terraform-provider-metal/issues/21#issuecomment-771653885_

PR #320, opened 2023-05-11
fix: don't set empty vnid before deleting virtual circuit

ctreatma:
A previous fix (removing unnecessary `port` from the VC state
waiter in #332) uncovered a downstream API error in the virtual circuit
delete code when deleting a VRF virtual circuit.

When any virtual circuit is deleted, the terraform provider was
trying to disconnect the VLAN from the circuit. For the VRF virtual
circuit tests, this request was getting a 422 response from the
API with the message "Update params can't be blank". Running
locally, both the VRF and VLAN virtual circuit tests pass without
this explicit disconnect request.

### DIFF
--- a/equinix/resource_metal_spot_market_request_acc_test.go
+++ b/equinix/resource_metal_spot_market_request_acc_test.go
@@ -2,12 +2,18 @@ package equinix
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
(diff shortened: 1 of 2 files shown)


ctreatma:
We got a passing run! https://github.com/equinix/terraform-provider-equinix/actions/runs/5082022133/jobs/9132074964?pr=320

Once #334 and #332 are merged I'll rebase this, confirm the workflow is still capable of passing & then mark this ready for review.

displague:
We should be able to close #130 and #198 when this is merged.

ocobleseqx (approved):
good work!

PR #322, opened 2023-05-16
docs: update docs to clarify that the IDs in user_ssh_key_ids are user IDs

ctreatma:
Our documentation incorrectly stated that the IDs in `user_ssh_key_ids` for an `equinix_metal_device` were the IDs of the SSH keys that will be supported on the device; in reality they are the IDs of the _users_ whose SSH keys will be supported.

In addition to the documentation, our tests were incorrectly attempting to pass an SSH key ID as a user ID, resulting in test failures due to errors from the API.

### DIFF
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -190,8 +190,18 @@ func TestAccMetalDevice_sshConfig(t *testing.T) {
 			{
 				Config: testAccMetalDeviceConfig_ssh_key(rs, userSSHKey, projSSHKey),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						r, "ssh_key_ids.#", "2"),
+					resource.TestCheckTypeSetElemAttrPair(
+						r,
+						"ssh_key_ids.*",
+						"equinix_metal_ssh_key.test",
+						"id",
+					),
+					resource.TestCheckTypeSetElemAttrPair(
+						r,
+						"ssh_key_ids.*",
+						"equinix_metal_project_ssh_key.test",
+						"id",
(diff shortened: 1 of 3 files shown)


equinix/resource_metal_device_acc_test.go
```diff
@@ -905,7 +905,7 @@ resource "equinix_metal_device" "test" {
 	operating_system = local.os
 	billing_cycle    = "hourly"
 	project_id       = equinix_metal_project.test.id
-	user_ssh_key_ids = [equinix_metal_ssh_key.test.id]
+	user_ssh_key_ids = [equinix_metal_ssh_key.test.owner_id]
```
ctreatma, line 908: This fixes the 422 error from the API, but the test is currently checking that the device has 2 SSH keys, and that is an invalid expectation.  It has every SSH key for the users specified in `user_ssh_key_ids` plus the 1 project SSH key specified in `project_ssh_key_ids`.  I'm not really sure how to resolve this in a reliable way; can we instead check that the device SSH keys include the project and user SSH keys that were created as part of the test?

ctreatma:
Updated test passes: https://github.com/equinix/terraform-provider-equinix/actions/runs/5004498703/jobs/8967176160#step:5:3040

displague (approved):

displague:
For reference:
- https://github.com/equinix/terraform-provider-equinix/blob/b88116305505674efb87fb160863f9c82a63442c/equinix/resource_metal_device.go#L560-L568
- https://github.com/equinix-labs/metal-go/blob/808a44026f8c7fc4ca69a0a8f185cbf568259cad/spec/oas3.fetched/components/schemas/DeviceCreateInput.yaml#L121-L185

PR #323, opened 2023-05-16
ci: ignore missing termination_time since API doesn't provide it

ctreatma:
While the `termination_time` attribute is supported for on-demand instances, the API currently cannot include that attribute in responses for on-demand devices (even with an `include` parameter, the API will not return that attribute).

This updates the failing device import test to ignore the `termination_time` attribute.  When the API adds support for this attribute in responses for on-demand devices, we can update the test to once again expect the imported `termination_time` attribute to match state.

### DIFF
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -636,9 +636,10 @@ func TestAccMetalDevice_importBasic(t *testing.T) {
 				Config: testAccMetalDeviceConfig_basic(rs),
 			},
 			{
-				ResourceName:      "equinix_metal_device.test",
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            "equinix_metal_device.test",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"termination_time"}, // Remove when API returns termination_time for on-demand instances
 			},
 		},
 	})


t0mk:
```
TF_LOG=DEBUG TF_ACC=1 go test -v -run=TestAccMetalDevice_impor
```

fails for me with 

```
23-05-16T17:54:50.625+0200 [DEBUG] sdk.helper_resource: Stopping providers: test_name=TestAccMetalDevice_importBasic test_step_number=2 test_terraform_path=/usr/bin/terraform test_working_directory=/tmp/plugintest1981768807
    resource_metal_device_acc_test.go:630: Failed state verification, resource with ID terraform-20230516155450012800000006 not found
```

I don't know where the `terraform-20230516155450012800000006` string comes from.

ctreatma:
@t0mk Does that error happen every time you run it or only occasionally?  It passed locally for me yesterday, and it [passes consistently in CI](https://github.com/equinix/terraform-provider-equinix/actions/runs/4993586062/jobs/8947638704?pr=323#step:5:1238).

I'll try running it locally again and see what happens.

t0mk:
@ctreatma Yes it happens consistently. But it also happens on master branch so it's not specific to this change. It's somehow connected to the data source declaration in the basic device config, here
https://github.com/equinix/terraform-provider-equinix/blob/952d84ad765a0e0ae55dbde7d6fed2667256a7bc/equinix/resource_metal_device_acc_test.go#L99

The `terraform-20230516155450012800000006` is generated ID for the data source. I'm not sure why TF tries to import the datasource.

I've just found out that it's a problem with terraform-plugin-sdk/v2. I updated it to v2.26.0 and then the import code passes with the modification from this PR.

I created a PR updating the TF SDK plugin to newest, you can decide if you want to take it in:
https://github.com/equinix/terraform-provider-equinix/pull/325

As I verified that the change works (albeit with different TF-SDK-plugin version), I approve this PR.

displague:
Relates to 
* https://github.com/equinix/terraform-provider-metal/pull/191#discussion_r709127245
* https://github.com/equinix/terraform-provider-equinix/pull/210

PR #328, opened 2023-05-17
ci: remove test of device port ordering

ctreatma:


ctreatma:
Pass line in CI for the updated test: https://github.com/equinix/terraform-provider-equinix/actions/runs/5005664109/jobs/8969932740#step:5:1518

displague (approved):
I thought we might have docs to update related to this, but we currently make no claim that the the ports will be in any given order.  I do wonder if we should state as much here or in [NetworkPorts](https://github.com/equinix-labs/metal-go/blob/main/docs/Device.md#properties).

https://github.com/equinix/terraform-provider-equinix/blob/master/docs/resources/equinix_metal_device.md#ports-attribute

ctreatma:
I don't know that we need to explicitly call out when there _isn't_ a guaranteed order for a list (for example, we also make no claim of an ordering for `ssh_key_ids` in the docs).  That said, the spec / metal-go docs at least explicitly state that the number of ports can vary between devices/plans, and that's probably worth copying over to the provider docs.

PR #329, opened 2023-05-17
ci: Stabilize plan/metro/facility selection in tests

ctreatma:
This adds the `random` provider to our acceptance tests so that we can use the `random_integer` resource to pick a random plan, facility and metro in each test and persist that choice in state.

By persisting the first random selection in state for each test, we avoid the potential to make a different random choice for each step in a multi-step test.  This enables us to remove `lifecycle` blocks from our tests so that we can more easily test the full behavior of our resources.


equinix/resource_metal_device_acc_test.go
```diff
@@ -111,36 +101,42 @@ data "equinix_metal_plans" "test" {
...
-    //With these operations we use current time seconds as the seed, and avoid using a third party provider in the Equinix provider tests
-    metros             = tolist(local.plans[local.plan_idx].available_in_metros)
-    metros_random_num  = formatdate("s", timestamp())
-    metros_length      = length(local.metros)
-    metros_range_limit = ceil(59 / local.metros_length) == 1 ? local.metros_length : 59
-    metro_idxs         = [for idx, value in range(0, local.metros_range_limit, ceil(59 / local.metros_length)) : idx if local.metros_random_num <= value]
-    metro_idx          = length(local.metro_idxs) > 0 ? local.metro_idxs[0] : 0
-    metro              = local.metros[local.metro_idx]
+    // Select a random plan
+	selected_plan     = data.equinix_metal_plans.test.plans[random_integer.plan_idx.result]
```
displague, line 130: Is the plan order (from the data source) guaranteed or do we need to add a sort argument to the data source? 
https://registry.terraform.io/providers/equinix/equinix/latest/docs/data-sources/equinix_metal_plans#sort

equinix/resource_metal_device_acc_test.go
```diff
@@ -111,36 +101,42 @@ data "equinix_metal_plans" "test" {
...
+    plan              = local.selected_plan.slug
+
+	// Select a random facility from the facilities in which the selected plan is available, excluding decommed facilities
+    facilities             = tolist(setsubtract(local.selected_plan.available_in, ["nrt1", "dfw2", "ewr1", "ams1", "sjc1", "ld7", "sy4", "ny6"]))
+    facility               = local.facilities[random_integer.facility_idx.result]
```
displague, line 135: curious if this is still in use

equinix/resource_metal_device_acc_test.go
displague: Same question for the other indexed data source results

ctreatma: Good call, added sorting to plans, plan facilities, and plan metros

equinix/resource_metal_device_acc_test.go
ocobleseqx: @displague I found some `local.facilities` refs in tests, so still in use https://github.com/equinix/terraform-provider-equinix/pull/329/files#diff-175a9bc69732939f0bf4dcedd034f66e05957561ffee186258e42fba19c2d2bcL62

ocobleseqx:
Some of the plans are not empty forcing replacement of the random resources https://github.com/equinix/terraform-provider-equinix/actions/runs/5015521380/jobs/8991199490?pr=329#step:5:1385, not sure if it's related with the removed 
lifecycle meta-argument

ctreatma:
A `lifecycle` block would ignore those changes, but in those cases the changes are being shown because the Metal `plan` data changed on the API side.  All of the terraform plans in the CI run that you linked show that the `metal_device.plan` attribute is not changing; that means that the available metros and facilities for the selected plan changed.  I'm assuming it should be pretty rare that plan data will change during an acceptance test run.  IMO it's better to fail with an unexpected non-empty plan in that case, and have someone inspect the failure and re-run, than to ignore the change and potentially fail later.

In the re-run of the failed CI job: https://github.com/equinix/terraform-provider-equinix/actions/runs/5015521380/jobs/9011641554

cprivitere (approved):

PR #334, opened 2023-05-24
ci: use spot market price as max bid in tests

ctreatma:
The spot market resource tests use a data source to determine the spot market price, and then attempt to create a spot market request with a bid price that is 20% higher than the spot market price.

This introduces brittleness to our tests because the bid price may be higher than the maximum allowed bid price.  This change uses the spot market price as the max bid price, rather than setting a higher bid, in an attempt to reduce the chance of test failure due to a disallowed bid price.

### DIFF
--- a/equinix/resource_metal_spot_market_request_acc_test.go
+++ b/equinix/resource_metal_spot_market_request_acc_test.go
@@ -93,7 +93,7 @@ data "equinix_metal_spot_market_request" "dreq" {
 
 resource "equinix_metal_spot_market_request" "request" {
   project_id       = equinix_metal_project.test.id
-  max_bid_price    = data.equinix_metal_spot_market_price.test.price * 1.2
+  max_bid_price    = format("%%.2f", data.equinix_metal_spot_market_price.test.price)
   facilities       = [data.equinix_metal_spot_market_price.test.facility]
   devices_min      = 1
   devices_max      = 1
@@ -123,7 +123,7 @@ data "equinix_metal_spot_market_price" "test" {
 
 resource "equinix_metal_spot_market_request" "request" {
   project_id       = equinix_metal_project.test.id
-  max_bid_price    = data.equinix_metal_spot_market_price.test.price * 1.2
+  max_bid_price    = data.equinix_metal_spot_market_price.test.price
   facilities       = [data.equinix_metal_spot_market_price.test.facility]
   devices_min      = 1
   devices_max      = 1


equinix/resource_metal_spot_market_request_acc_test.go
```diff
@@ -93,7 +93,7 @@ data "equinix_metal_spot_market_request" "dreq" {
 
 resource "equinix_metal_spot_market_request" "request" {
   project_id       = equinix_metal_project.test.id
-  max_bid_price    = data.equinix_metal_spot_market_price.test.price * 1.2
+  max_bid_price    = data.equinix_metal_spot_market_price.test.price
```
ctreatma, line 96: Here's an example of failure due to a bid price that is higher than the allowed bid price: https://github.com/equinix/terraform-provider-equinix/actions/runs/5050831197/jobs/9062032812#step:5:624

It's possible we should be extra safe and multiply the spot market price by `0.95` instead?

ctreatma: ...aaand it looks like we _do_ need to further reduce the max bid price, because this test still failed: https://github.com/equinix/terraform-provider-equinix/actions/runs/5072458466/jobs/9110213013#step:5:673

The spot market data source tests have a hard-coded max bid of `0.01`, and I don't think I've seen those fail yet; is there some reason we're using a dynamic max bid for these tests?

ctreatma:
Now the spot market import test seems to be passing and then timing out trying to delete the request: https://github.com/equinix/terraform-provider-equinix/actions/runs/5072719458/jobs/9110803431#step:5:4307

equinix/resource_metal_spot_market_request_acc_test.go
```diff
@@ -93,7 +93,7 @@ data "equinix_metal_spot_market_request" "dreq" {
 
 resource "equinix_metal_spot_market_request" "request" {
   project_id       = equinix_metal_project.test.id
-  max_bid_price    = data.equinix_metal_spot_market_price.test.price * 1.2
+  max_bid_price    = format("%%.2f", data.equinix_metal_spot_market_price.test.price)
```
ctreatma, line 96: With this latest approach to passing in the original spot market price, both tests pass:
- https://github.com/equinix/terraform-provider-equinix/actions/runs/5081637323/jobs/9130308071?pr=334#step:5:1141
- https://github.com/equinix/terraform-provider-equinix/actions/runs/5081637323/jobs/9130308071?pr=334#step:5:1757

displague (approved):

PR #338, opened 2023-06-02
feat: add optional metal_base (EQUINIX_API_METAL_BASE) provider config

displague:
Adds support for alternate Equinix Metal Base Path. 
This is used in Prism spec validation: https://gist.github.com/displague/27363182851f0eab4b599401f83f4b1e. 

There may not be other real-world test scenarios today outside of the use-case described. If there were a `/metal/v1{suffix}`, this could be helpful.

An alternative to this base path variable would be one that replaces the full Metal URL. This would provide ways to test against the legacy api.packet.net path structure (and other environments) for example.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -112,6 +112,10 @@ services, and `auth_token` to interact with Equinix Metal.
 
 * `max_retry_wait_seconds` (Optional) Maximum time to wait in case of network failure.
 
+* `metal_base` (Optional) The Equinix Metal API base path for the API `endpoint`.
+   This argument can also be specified with the `EQUINIX_API_METAL_BASE` shell environment
+   variable. (Defaults to `/metal/v1`)
+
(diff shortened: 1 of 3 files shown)


ctreatma:
I don't think we should do this.  For the Prism case, that seems to be an issue with Prism (it doesn't respect the base path that is baked in to `servers.0.url`) and could possibly be worked around with something like [this](https://github.com/stoplightio/prism/issues/1591#issuecomment-880712301) or by making a contribution to Prism that would have that proxy respect any base path in the specified server URL.  For the `api.packet.net` case, IMO that _should_ be difficult to configure; our customers should use (and be able to rely on) our standard `api.equinix.com` domains.

displague:
Closing based on @ctreatma's recommendations for alternate approaches and because we haven't seen the Spotlight integration manifest since this PR was created.

PR #351, opened 2023-08-07
Typo in spot_market_request argument name: termintation->termination

t0mk:
Fixes #253 

This PR fixes typo in spot_market_request argument name, it was supposed to be "termination" but ended up being "termintation". I added a new arg with a correct name and made the old on deprecated. We can remove it later on.

Done according to
https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-an-optional-attribute

### DIFF
--- a/equinix/resource_metal_spot_market_request.go
+++ b/equinix/resource_metal_spot_market_request.go
@@ -135,6 +135,11 @@ func resourceMetalSpotMarketRequest() *schema.Resource {
 							Required: true,
 						},
 						"termintation_time": {
+							Type:       schema.TypeString,
+							Computed:   true,
+							Deprecated: "Use instance_parameters.termination_time instead",
+						},
+						"termination_time": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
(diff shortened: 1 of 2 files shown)


t0mk:
The tests flake-fail on unrelated issues. What is the protocol for this?

ctreatma:
There appear to be valid test failures here that are related to the changed code:
```
=== RUN   TestProvider
    provider_test.go:139: err: 1 error occurred:
        	* resource equinix_metal_spot_market_request: ConflictsWith: termintation_time references unknown attribute (termination_time) at part (termination_time)
        
--- FAIL: TestProvider (0.01s)
```
https://github.com/equinix/terraform-provider-equinix/actions/runs/5786280225/job/15680673114?pr=351

equinix/resource_metal_spot_market_request.go
```diff
@@ -135,8 +135,15 @@ func resourceMetalSpotMarketRequest() *schema.Resource {
 							Required: true,
 						},
 						"termintation_time": {
-							Type:     schema.TypeString,
-							Computed: true,
+							Type:          schema.TypeString,
+							Optional:      true,
```
ctreatma, line 139: I think these should both still be `Computed: true`

ctreatma: The termination time for a spot market request resource is only read from the API; we don't have any code to include it in the spot market request that is sent to the API.

ctreatma (approved):

PR #352, opened 2023-08-07
chore(ci): fix SSH key test and document project name max length

ctreatma:
At some point, the Metal API introduced validation that rejects project names longer than 80 characters.  This broke one of our device tests, which was using the public SSH key contents as the project suffix instead of a random number.

This updates the broken test to use a random number suffix for project names so that the test can pass again.  The project resource docs are also updated to mention the 80-character limit.

### DIFF
--- a/docs/resources/equinix_metal_project.md
+++ b/docs/resources/equinix_metal_project.md
@@ -67,7 +67,7 @@ template.
 
 The following arguments are supported:
 
-* `name` - (Required) The name of the project.
+* `name` - (Required) The name of the project.  The maximum length is 80 characters
 * `organization_id` - (Required) The UUID of organization under which you want to create the project. If you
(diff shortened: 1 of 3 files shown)


ctreatma:
Pass line for the affected test: https://github.com/equinix/terraform-provider-equinix/actions/runs/5786997385/job/15682943280?pr=352#step:5:2899

At the moment, retrying tests won't lead to a fully-passing suite because of an issue with one of the organization acceptance tests that is being investigated by the billing team.

ctreatma:
Affected test is still passing: https://github.com/equinix/terraform-provider-equinix/actions/runs/5798496984/job/15716390685?pr=352#step:5:2875

t0mk (approved):
lgtm

PR #353, opened 2023-08-07
feat: introduce metal-go client

ctreatma:
This adds a metal-go client to the terraform provider.  The metal-go client can be gradually adopted by Metal resources and data sources in this provider to enable the eventual deprecation of packngo.

An example of using this client to interact with the Metal API can be found in #291.


displague (approved):
We had a few failed E2E tests here but nothing that can be associated with this change.

PR #356, opened 2023-08-14
Support CRUD timeouts for metal device

aayushrangwala:
To support timeouts for the CRUD operations by equinix provider, upgrading to context aware schema from terraform
https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context

#357


aayushrangwala:
One of the solution would be to use the timeout while we wait for the device/resource to be active/created. For ex [here](https://github.com/equinix/terraform-provider-equinix/blob/d672cd5f2cd7a4e8a2b2320eadd5e4a26ef5c19a/equinix/resource_metal_device.go#L846)
Or we use the context aware handlers and use `metal-go` apis to delegate the handling of `context cancellation`

Checked for aws provider [here](https://github.com/hashicorp/terraform-provider-aws/blob/a4393f91f2b25b915a0780ed532b83f2bc68ff2b/internal/service/ec2/ec2_instance.go#L1007) and they have been they consume the timeout while waiting the aws instance to be created

displague:
@aayushrangwala I think one of the places where we'll want to respect the context that is passed in is the `waitFor*` functions (`waitForActiveDevice` for example).

aayushrangwala:
> @aayushrangwala I think one of the places where we'll want to respect the context that is passed in is the `waitFor*` functions (`waitForActiveDevice` for example).

Also, this seems to be a bit out of scope for a single PR. We can break it down to implement timeouts for resource/datasource level, linked to this bug

Also, we can use the terraform helper provided pattern for `wait` methods as this
https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/retry/state.go#L57

examples/connectivity/equinix-metal-to-fabric-connection/main.tf
```diff
@@ -8,7 +8,7 @@ provider "metal" {
 }
 
 data "metal_project" "this" {
-    name = var.metal_project_name
+  name = var.metal_project_name
```
ctreatma, line 11: Could you either revert the formatting changes in the `examples` directory or open a separate PR with only the formatting changes?

aayushrangwala: Will do the changes. Closing this PR in the favour smaller PRs per resource/data source

PR #361, opened 2023-08-17
NFV-25337: Added attribute Connectivity to support device creation with no internet access

schadalawada:
In the August release, NE has introduced a new feature that enables customers to create devices with or without internet access. To accommodate this change, the Terraform code has been updated. Below is the reference for these changes.

Reference: https://developer.equinix.com/docs?page=/dev-docs/ne/overview

### DIFF
--- a/docs/data-sources/equinix_network_device.md
+++ b/docs/data-sources/equinix_network_device.md
@@ -49,6 +49,7 @@ NOTE: Exactly one of either `uuid` or `name` must be specified.
   * APPLIED
   * WAITING_FOR_CLUSTER_SETUP
   * REGISTRATION_FAILED
+  * NA
 * `license_file_id` - Unique identifier of applied license file
 * `ibx` - Device location Equinix Business Exchange name
 * `region` - Device location region
@@ -71,3 +72,4 @@ primary or secondary
 * `zone_code` - Device location zone code
 * `cluster_id` - The id of the cluster
 * `num_of_nodes` - The number of nodes in the cluster
+* `connectivity` - Device accessibility (INTERNET-ACCESS or PRIVATE or INTERNET-ACCESS-WITH-PRVT-MGMT)
--- a/docs/resources/equinix_network_device.md
+++ b/docs/resources/equinix_network_device.md
@@ -43,6 +43,7 @@ resource "equinix_network_device" "csr1000v-ha" {
   metro_code      = data.equinix_network_account.dc.metro_code
   type_code       = "CSR1000V"
   self_managed    = false
+  connectivity    = "INTERNET-ACCESS"
   byol            = false
   package_code    = "SEC"
   notifications   = ["john@equinix.com", "marry@equinix.com", "fred@equinix.com"]
@@ -190,6 +191,7 @@ resource "equinix_network_device" "arista-ha" {
   metro_code      = data.equinix_network_account.sv.metro_code
   type_code       = "ARISTA-ROUTER"
   self_managed    = true
+  connectivity    = "PRIVATE"
   byol            = true
   package_code    = "CloudEOS"
   notifications   = ["test@equinix.com"]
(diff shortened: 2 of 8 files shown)


equinix/data_source_network_device.go
```diff
@@ -275,6 +275,12 @@ func createDataSourceNetworkDeviceSchema() map[string]*schema.Schema {
 			Computed:    true,
 			Description: neDeviceDescriptions["ZoneCode"],
 		},
+		neDeviceSchemaNames["Connectivity"]: {
+			Type:         schema.TypeString,
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"PRIVATE"}, true),
```
rling-equinix, line 281: All the enums need to be listed here.

rling-equinix: This is for data source (GET device by uuid API). Therefore, it's computed. We don't need optional flag here. Computed flag is required.

equinix/resource_network_device.go
```diff
@@ -479,6 +481,11 @@ func createNetworkDeviceSchema() map[string]*schema.Schema {
 			Computed:    true,
 			Description: neDeviceDescriptions["ZoneCode"],
 		},
+		neDeviceSchemaNames["Connectivity"]: {
+			Type:        schema.TypeString,
+			Optional:    true,
```
rling-equinix, line 486: Default, ForceNew, and validation function are required here. You can check isBYOL field for reference.

equinix/resource_network_device.go
```diff
@@ -1100,6 +1107,9 @@ func createNetworkDevices(d *schema.ResourceData) (*ne.Device, *ne.Device) {
 	if v, ok := d.GetOk(neDeviceSchemaNames["ClusterDetails"]); ok {
 		primary.ClusterDetails = expandNetworkDeviceClusterDetails(v.([]interface{}))
 	}
+	if v, ok := d.GetOk(neDeviceSchemaNames["Connectivity"]); ok {
+		primary.Connectivity = ne.String(v.(string))
```
rling-equinix, line 1111: HA(secondary device) case needs to be updated and tested along with this one.

equinix/resource_network_device_test.go
```diff
@@ -85,6 +85,80 @@ func TestNetworkDevice_createFromResourceData(t *testing.T) {
 	assert.Equal(t, expectedPrimary, primary, "Primary device matches expected result")
 }
 
+func TestNetworkDeviceWithNoInternetAccess_createFromResourceData(t *testing.T) {
```
rling-equinix, line 88: We don't need to create new test case to cover the new field connectivity. You can simply add one more line in the previous test case.

rling-equinix (changes requested):

docs/data-sources/equinix_network_device.md
```diff
@@ -49,6 +49,7 @@ NOTE: Exactly one of either `uuid` or `name` must be specified.
   * APPLIED
   * WAITING_FOR_CLUSTER_SETUP
   * REGISTRATION_FAILED
+  * NA
```
rling-equinix, line 52: Documentation for the new field connectivity is also required in both data source and resource.

equinix/resource_network_device.go
rling-equinix: Connectivity field will be applied to both primary and secondary device.

rling-equinix (dismissed):

docs/resources/equinix_network_device.md
```diff
@@ -43,6 +43,7 @@ resource "equinix_network_device" "csr1000v-ha" {
   metro_code      = data.equinix_network_account.dc.metro_code
   type_code       = "CSR1000V"
   self_managed    = false
+  connectivity    = INTERNET-ACCESS
```
ctreatma, line 46: The values for `connectivity` in examples should be wrapped in quotes, since the field is a string.
```suggestion
  connectivity    = "INTERNET-ACCESS"
```

docs/resources/equinix_network_device.md
```diff
@@ -190,6 +191,7 @@ resource "equinix_network_device" "arista-ha" {
   metro_code      = data.equinix_network_account.sv.metro_code
   type_code       = "ARISTA-ROUTER"
   self_managed    = true
+  connectivity    = PRIVATE
```
ctreatma, line 194:
```suggestion
  connectivity    = "PRIVATE"
```

equinix/resource_network_device.go
ctreatma: @rling-equinix just to confirm, you're saying that we only have to set `primary.Connectivity` because the API will automatically apply that value to both primary and secondary device, and there is no way via the API to set a different value for primary and secondary?

rling-equinix: Yes, correct.

rling-equinix (approved):

PR #362, opened 2023-08-21
Add Fabric Cloud Router and routing protocol terraform provider

manu-equinix:
- Adds new Fabric resources and datasources:
  - equinix_fabric_cloud_router
  - equinix_fabric_routingprotocol

- Updates Fabric-go SDK version from v0.4.0(Fabric API v4.4) to v0.6.0(Fabric API v4.9)
- Fabric API spec v.49 can be found at - https://app.swaggerhub.com/apis/equinix-api/fabric/4.9
- Replaces Datasource equinix_fabric_connection type attribute value GW_VC with IP_VC
- Replaces Equinix Fabric Gateway with Equinix Fabric Cloud router
- Includes terraform examples for the above 2 providers and all the related documentation updated
- All related changes are listed on the release notes documented on the below link as part of June 2023 and Aug 2023 releases - https://developer.equinix.com/dev-docs/fabric/release-notes


ctreatma (commented):
As I understand it, these new resources are replacing the old `fabric_gateway_*` resources?  If so, the old resources should be updated according to note that they have been renamed and will be removed in a future release.  Hashicorp provides docs on how to do that here:
- https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-rename
- https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal

manu-equinix:
> As I understand it, these new resources are replacing the old `fabric_gateway_*` resources? If so, the old resources should be updated according to note that they have been renamed and will be removed in a future release. Hashicorp provides docs on how to do that here:
> 
> * https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-rename
> * https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal

@ctreatma From the terraform provider point of view, there are no changes to fabric_gateway_* prvider since this has not been released. This change adds a new provider which we call is "equinix_fabric_cloud_router".

docs/resources/equinix_fabric_routingprotocol.md
```diff
@@ -0,0 +1,178 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "equinix_fabric_routingprotocol Resource - terraform-provider-equinix"
+subcategory: "Fabric"
+description: |-
+  Fabric V4 API compatible resource allows creation and management of Equinix Fabric connection
+  ~> Note Equinix Fabric v4 resources and datasources are currently in Beta. The interfaces related to equinix_fabric_ resources and datasources may change ahead of general availability. Please, do not hesitate to report any problems that you experience by opening a new issue https://github.com/equinix/terraform-provider-equinix/issues/new?template=bug.md
+---
+
+# equinix_fabric_routingprotocol (Resource)
```
ctreatma, line 10: Elsewhere, the naming convention is to separate words with underscores.  For this resource & corresponding data sources, why is it `routingprotocol` rather than `routing_protocol`?

equinix/fabric_connection_read_schema.go
```diff
@@ -419,10 +419,10 @@ func readConnectionSideAccessPointSch() map[string]*schema.Schema {
 				Schema: readServiceProfileSch(),
 			},
 		},
-		"gateway": {
+		"router": {
```
ctreatma, line 422: This is a breaking change; the process laid out here will enable a smoother upgrade for users: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-a-computed-attribute

equinix/fabric_connection_schema.go
```diff
@@ -340,10 +341,10 @@ func createConnectionSideAccessPointRes() *schema.Resource {
 					Schema: createServiceProfileSch(),
 				},
 			},
-			"gateway": {
+			"router": {
```
ctreatma, line 344: Similar to the above comment, this should be done in a way that does not break existing configurations: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-a-computed-attribute

equinix/fabric_port_read_schema.go
```diff
@@ -307,13 +307,10 @@ func readFabricPortResourceSchema() map[string]*schema.Schema {
 				Schema: readFabricPortEncapsulation(),
 			},
 		},
-		"lag": {
-			Type:        schema.TypeSet,
+		"lag_enabled": {
```
ctreatma, line 310: This is also a breaking change and we should deprecate the old field before removing it: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-a-computed-attribute

equinix/fabric_connection_schema.go
```diff
@@ -296,8 +296,8 @@ func createConnectionSideAccessPointRes() *schema.Resource {
 			"type": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringInSlice([]string{"COLO", "VD", "VG", "SP", "IGW", "SUBNET", "GW"}, true),
```
ctreatma, line 299: Is the "GW" value still supported by the API, or would existing configs that have `type: "GW"` already be broken because the API no longer supports that value?

docs/resources/equinix_fabric_routingprotocol.md
manu-equinix: @sabwong Can you please take care of this?

equinix/fabric_connection_schema.go
manu-equinix: Yes, this has been changed to CLOUD_ROUTER on the API side as well and no longer supports GW.

docs/resources/equinix_fabric_routingprotocol.md
sabwong: addressed in code and docs.

equinix/fabric_port_mapping_helper.go
```diff
@@ -104,27 +104,6 @@ func encapsulationToTerra(portEncapsulation *v4.PortEncapsulation) *schema.Set {
 	return portEncapsulationSet
 }
 
-func lagToTerra(portLag *v4.PortLag) *schema.Set {
-	if portLag == nil {
```
ctreatma, line 108: Does the API still return the fields that are needed for this function?  I raised a concern about breaking changes for the `lag` attribute elsewhere on this PR, but if the old attribute can't work anyway because the API already removed these fields, then my concern is moot.

manu-equinix: Yes, I believe the API is fixed now. cc @anchala71

equinix/fabric_port_read_schema.go
manu-equinix: @thogarty Added a fix for all the deprecated fields you pointed out. However, the tests are failing now and we are looking into it.

equinix/fabric_connection_read_schema.go
manu-equinix: This is now fixed with the latest commit. Please check and approve.

equinix/fabric_connection_schema.go
manu-equinix: This is now fixed with the latest commit. Please check and approve.

equinix/fabric_port_read_schema.go
manu-equinix: This is now fixed with the latest commit. Please check and approve.

ctreatma (approved):
Approved from a general provider perspective; please have someone from @equinix/governor-digin-fabric review this work from a product/testing perspective before it is merged.

manu-equinix:
@ctreatma We have tested this change and most members from the https://github.com/orgs/equinix/teams/governor-digin-fabric have contributed to this PR.

Issue #357, opened 2023-08-15
Timeouts are not honored on Metal resources/data sources


aayushrangwala:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Terraform Version

`v1.5.4`

### Affected Resource(s)
- Metal Devices resource and datasource
- Metal ports
- Metal VRF
- Network Device
- SSH Key and a few more

### Terraform Configuration Files

```tf
resource "equinix_metal_device" "device" {
  hostname         = "tf-device"
  plan             = var.plan
  metro            = var.metro
  operating_system = var.os
  billing_cycle    = "hourly"
  project_id       = var.project_id
#
  depends_on = [ equinix_metal_project_ssh_key.public_key ]

  timeouts {
    create = "5s"
  }
}
```

### References
Some of the Metal resources/data sources (for example, the equinix_metal_device resource) indicate support for timeouts in their docs and schema, but do not actually support those timeouts.

Per terraform-plugin-sdk v2 docs, we should be using CreateContext, DeleteContext, etc. functions in order to have terraform automatically handle timeouts on our behalf: https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context 

Also, for the resources and datasources when the `context aware` timeouts would not be suitable, use Hashicorp's [WaitForStateContext](https://github.com/hashicorp/terraform-plugin-sdk/blob/8bde488a9529d281bc6858ca2b0a3708c2f93658/helper/retry/state.go#L57)

We should find any Metal resources & data sources that have a Timeouts schema property and ensure that those resources & data sources use context-aware CRUD functions.

### Tasks

Upgrade resources and data sources to support context aware or without context CRUD timeouts

- [x] [DataSource and Resource Metal Device support](https://github.com/equinix/terraform-provider-equinix/pull/364)
- [ ] [DataSource and Resource Metal Port support](https://github.com/equinix/terraform-provider-equinix/pull/377)
- [ ] [DataSource and Resource Metal Sport Market Request support](https://github.com/equinix/terraform-provider-equinix/pull/379)
[...]

PR #364, opened 2023-08-23
Upgrade metal device resource and data source to use timeouts

aayushrangwala:
Provides support for timeouts in Create, Update and Delete for Metal Device as already mentioned in the documents
#357

### DIFF
--- a/equinix/helpers_device.go
+++ b/equinix/helpers_device.go
@@ -1,6 +1,8 @@
 package equinix
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -8,7 +10,6 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/packethost/packngo"
@@ -142,7 +143,7 @@ func hwReservationStateRefreshFunc(client *packngo.Client, reservationId, instan
 	}
 }
 
-func waitUntilReservationProvisionable(client *packngo.Client, reservationId, instanceId string, delay, timeout, minTimeout time.Duration) error {
+func waitUntilReservationProvisionable(ctx context.Context, client *packngo.Client, reservationId, instanceId string, delay, timeout, minTimeout time.Duration) error {
 	stateConf := &retry.StateChangeConf{
 		Pending:    []string{deprovisioning},
 		Target:     []string{provisionable, reprovisioned},
@@ -151,7 +152,7 @@ func waitUntilReservationProvisionable(client *packngo.Client, reservationId, in
 		Delay:      delay,
 		MinTimeout: minTimeout,
 	}
-	_, err := stateConf.WaitForState()
+	_, err := stateConf.WaitForStateContext(ctx)
 	return err
 }
 
@@ -166,7 +167,7 @@ func getWaitForDeviceLock(deviceID string) *sync.WaitGroup {
 	return wg
 }
 
-func waitForDeviceAttribute(d *schema.ResourceData, targets []string, pending []string, attribute string, meta interface{}) (string, error) {
+func waitForDeviceAttribute(ctx context.Context, d *schema.ResourceData, stateConf *retry.StateChangeConf) (string, error) {
 	wg := getWaitForDeviceLock(d.Id())
 	wg.Wait()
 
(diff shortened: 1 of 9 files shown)


equinix/helpers_device.go
```diff
@@ -216,34 +194,6 @@ func waitForDeviceAttribute(d *schema.ResourceData, targets []string, pending []
 	return "", err
 }
 
-// powerOnAndWait Powers on the device and waits for it to be active.
-func powerOnAndWait(d *schema.ResourceData, meta interface{}) error {
```
ctreatma, line 220: Just calling out that this is being removed because it's not used anywhere.  Nice find!

equinix/provider_test.go
```diff
@@ -127,6 +128,11 @@ func init() {
 	testAccProviders = map[string]*schema.Provider{
 		"equinix": testAccProvider,
 	}
+	testAccProviderFactories = map[string]func() (*schema.Provider, error){
```
ctreatma, line 131: Can you link to docs about what this change does?  Will we eventually have only one of `testAccProviders` or `testAccProviderFactories`, or will we always need both?

aayushrangwala: From the official testing lib doc, providers is deprecated and providerFactories needs to be use from now on.
https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCase

We need a bug/enhancement task to use`testAccProviderFactores` instead of `testAccProviders`. `goodFirstIssue`

equinix/resource_metal_device_acc_test.go
```diff
@@ -1121,9 +1144,75 @@ func TestAccMetalDevice_readErrorHandling(t *testing.T) {
...
+
+func TestAccMetalDevice_timeouts(t *testing.T) {
+	d1 := &packngo.Device{
+		State: "active",
+	}
+	projSuffix := acctest.RandString(10)
+
+	mockMetalDeviceService := &mockDeviceService{
+		CreateFn: func(device *packngo.DeviceCreateRequest) (*packngo.Device, *packngo.Response, error) {
+			return d1, &packngo.Response{Response: &http.Response{Status: matchErrDeviceReadyTimeout.String(), StatusCode: 500}},
```
ctreatma, line 1162: I don't think we should mock packngo in order to return terraform provider error messages.  A mock that waits a certain amount of time before responding would be more accurate for the test case, but an integration test would be preferable; with a short timeout we should be able to guarantee a timeout error in that test.

displague: Shall we rename this to `TestMetalDevice_timeouts` (no `Acc`) so the offline test can run in our unit tests?

displague: I was overlooking that this adds more dependency on packngo (vs metal-go).

equinix/resource_network_bgp_test.go
```diff
@@ -100,6 +101,20 @@ func (r *mockedBGPUpdateRequest) Execute() error {
 	return nil
 }
 
+var _ packngo.BGPConfigService = &mockBgpConfigService{}
```
ctreatma, line 104: If we do stick with mocking the SDK (I still think we shouldn't, but it sounds like maybe we will), this mock should be in a different file.  The `resource_network_bgp_test.go` file is a test file for a Network Edge resource, not a Metal resource.

equinix/resource_metal_device_acc_test.go
ctreatma: If we want to stick with mocking _something_, this article lays out a few options that would allow us to mock API behavior rather than SDK behavior:

https://medium.com/zus-health/mocking-outbound-http-requests-in-go-youre-probably-doing-it-wrong-60373a38d2aa#fb9e

In particular I've linked to the section on [`httptest`](https://pkg.go.dev/net/http/httptest) since it's built-in, but there are other approaches to consider.

aayushrangwala: The overall process to get the device successfully created is when we the device is  provisioned and active which takes more time. For any resource create via terraform provider, which take time to be provisioned, we need to consider the actual `wait` time for that resource to be in desired state and not the API call itself. Most of the time API calls will return in no time. Reference for discussion [here](https://github.com/equinix/terraform-provider-equinix/pull/356#issuecomment-1679316443)

Checked for aws provider [here](https://github.com/hashicorp/terraform-provider-aws/blob/a4393f91f2b25b915a0780ed532b83f2bc68ff2b/internal/service/ec2/ec2_instance.go#L1007) and they have been they consume the timeout while waiting the aws instance to be created. Also checked k8s provider where the api itself is taking and respecting the context with timeout

We can use the terraform helper provided pattern for wait methods as this. As what most of the other providers are also using
https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/retry/state.go#L57

aayushrangwala: The main problem happening with this timeout test is, in the test we provide a config to terraform to create a device with a timeout of very small `5s`. The test return timeout error which is expected and then test tries to destroy the resource. But the issue is equinix apis doesnt allow the device to be destroyed when they are in provisioning state. There's no way to tell terraform to wait and then destroy.

Also, since we are anyway not considering the API call time in our timeout waits, it can be mocked

equinix/resource_metal_device_acc_test.go
```diff
@@ -747,30 +749,6 @@ resource "equinix_metal_device" "test" {
 `, confAccMetalDevice_base(preferable_plans, preferable_metros, preferable_os), projSuffix, rInt, rInt, rInt, testDeviceTerminationTime())
 }
 
-func testAccMetalDeviceConfig_varname_pxe(rInt int, projSuffix string) string {
```
ctreatma, line 750: Was this test deleted by accident or does it need to be removed for some reason?

aayushrangwala: Yes, this test helper was not being used anywhere. Should have called that out

equinix/resource_metal_device_acc_test.go
```diff
@@ -938,6 +916,30 @@ resource "equinix_metal_device" "test_ipxe_missing" {
...
+	return fmt.Sprintf(`
+resource "equinix_metal_project" "test" {
+    name = "tfacc-device-%s"
+}
+
+resource "equinix_metal_device" "test" {
+  hostname         = "tfacc-test-device"
+  description      = "test-desc"
+  plan             = "c3.small.x86"
+  metro            = "DA"
```
ctreatma, line 929: For end-to-end tests, the plan, metro, and OS should be parameterized so that the tests are more flexible:
- see https://github.com/equinix/terraform-provider-equinix/pull/364/files#diff-b5bc3c33d7df2c76175fac51f0b7b4a5d4e2fd0268430d3b28091961885d775fL761-L763 for an example
- and note that some extra stuff needs to be added in to the test config to provide the values for those fields: https://github.com/equinix/terraform-provider-equinix/pull/364/files#diff-b5bc3c33d7df2c76175fac51f0b7b4a5d4e2fd0268430d3b28091961885d775fL771

ctreatma (approved):
I see the new tests passing in CI:
- [TestAccMetalDeviceCreate_timeout](https://github.com/equinix/terraform-provider-equinix/actions/runs/6026843556/job/16350709616?pr=364#step:5:2376)
- [TestAccMetalDeviceUpdate_timeout](https://github.com/equinix/terraform-provider-equinix/actions/runs/6026843556/job/16350709616?pr=364#step:5:2297)

PR #366, opened 2023-08-25
feat: metal_devices datasource

### LINKED ISSUE #288: Add an equinix_metal_devices data source
### Community Note

### Description

Currently, the Equinix provider only includes an `equinix_metal_device` data source, which is only capable of returning exactly one device.  In some cases, it may be useful for customers to look up multiple similar devices; for example, a customer may want to look up a group of devices with similar hostnames.  To support these cases, we could introduce an `equinix_metal_devices` data source that is capable of returning data for any number of devices.

Relevant API endpoints:
- https://deploy.equinix.com/developers/api/metal/#tag/Devices/operation/findProjectDevices
- https://deploy.equinix.com/developers/api/metal/#tag/Devices/operation/findOrganizationDevices

* equinix_metal_devices (data source)

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
data "equinix_metal_devices" "my_devices" {
  project_id = "<a_project_id>"
  filter {
    attribute = "hostname"
    value = "my-prefix"
  }
}
```

### References

* #0000

t0mk:
This PR adds `metal_devices` datasource, which fetches lists of device, similarly to [metal_plans](https://github.com/equinix/terraform-provider-equinix/blob/main/equinix/data_source_metal_plans.go) ([doc](https://registry.terraform.io/providers/equinix/equinix/latest/docs/data-sources/equinix_metal_plans))

fixes #288 

The datasource is using metal-go.

### DIFF
--- a/docs/data-sources/equinix_metal_devices.md
+++ b/docs/data-sources/equinix_metal_devices.md
@@ -43,7 +43,7 @@ output "devices" {
 
 ## search vs filter
 
-The difference between `search` and `filter` is that `search` is API parameter and the "filtering" will happen in the API side. The `filter` attribute is used in the terraform provider to fiter resoruces returned by the API.
+The difference between `search` and `filter` is that `search` is an API parameter, interpreted by the Equinix Metal service. The "filter" arguments will reduce the API list (or search) results by applying client-side filtering, within this provider.
 
 ## Argument Reference
 
--- a/docs/data-sources/equinix_metal_devices.md
+++ b/docs/data-sources/equinix_metal_devices.md
@@ -51,7 +51,7 @@ The following arguments are supported:
 
 * `project_id` - (Optional) ID of project containing the devices. Exactly one of `project_id` and `organization_id` must be set.
 * `organization_id` - (Optional) ID of organization containing the devices.
-* `search` - (Optional) - Search string to filter devices by hostname, description, short_id, reservation short_id, tags, plan name, plan slug, facility code, facility name, operating system name, operating system slug, IP addresses.",
+* `search` - (Optional) - Search string to filter devices by hostname, description, short_id, reservation short_id, tags, plan name, plan slug, facility code, facility name, operating system name, operating system slug, IP addresses.
 * `filter` - (Optional) One or more attribute/values pairs to filter off of
   - `attribute` - (Required) The attribute used to filter. Filter attributes are case-sensitive
   - `values` - (Required) The filter values. Filter values are case-sensitive. If you specify multiple values for a filter, the values are joined with an OR by default, and the request returns all results that match any of the specified values
--- a/docs/data-sources/equinix_metal_devices.md
+++ b/docs/data-sources/equinix_metal_devices.md
@@ -1,0 +1,67 @@
+---
+subcategory: "Metal"
+---
+
+# equinix_metal_devices
+
+Provides an Equinix Metal devices datasource. This can be used to find devices that meet a filter criteria.
+
+## Example Usage
+
+```hcl
+# Following example will select c3.small.x86 devices which are deplyed in metro 'da' (Dallas)
+# OR 'sv' (Sillicon Valley).
+data "equinix_metal_devices" "example" {
+    project_id = local.project_id
+    filter {
+        attribute = "plan"
+        values    = ["c3.small.x86"]
+    }
+    filter {
+        attribute = "metro"
+        values    = ["da", "sv"]
+    }
+}
+
+output "devices" {
+    organization_id = local.org_id
+    value = data.equinix_metal_devices.example.devices
+}
+```
+
+```hcl
+# Following example takes advantage of the `search` field in the API request, and will select devices with
+# string "database" in one of the searched attributes. See `search` in argument reference.
+data "equinix_metal_devices" "example" {
+    search = "database"
+}
+
+output "devices" {
+    value = data.equinix_metal_devices.example.devices
+}
+```
+
+## search vs filter
+
+The difference between `search` and `filter` is that `search` is API parameter and the "filtering" will happen in the API side. The `filter` attribute is used in the terraform provider to fiter resoruces returned by the API.
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Optional) ID of project containing the devices. Exactly one of `project_id` and `organization_id` must be set.
+* `organization_id` - (Optional) ID of organization containing the devices.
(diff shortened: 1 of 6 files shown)


t0mk:
I still need to write docs

equinix/helpers_device.go
```diff
@@ -271,3 +319,42 @@ func ipAddressSchema() *schema.Resource {
 		},
 	}
 }
+
+func getDeviceMap(device metalv1.Device) map[string]interface{} {
```
t0mk, line 323: this function can be used in the singular datasource, and maybe also in device resources

t0mk:
Failing tests are some L2 tests and metal_device datasource test which hit metro capacity.

docs/data-sources/equinix_metal_devices.md
```diff
@@ -0,0 +1,67 @@
...
+}
+
+output "devices" {
+    value = data.equinix_metal_devices.example.devices
+}
+```
+
+## search vs filter
+
+The difference between `search` and `filter` is that `search` is API parameter and the "filtering" will happen in the API side. The `filter` attribute is used in the terraform provider to fiter resoruces returned by the API.
```
displague, line 46:
```suggestion
The difference between `search` and `filter` is that `search` is an API parameter, interpreted by the Equinix Metal service. The "filter" arguments will be reduce the API list (or search) results by applying client-side filtering, within this provider.
```

docs/data-sources/equinix_metal_devices.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Optional) ID of project containing the devices. Exactly one of `project_id` and `organization_id` must be set.
+* `organization_id` - (Optional) ID of organization containing the devices.
+* `search` - (Optional) - Search string to filter devices by hostname, description, short_id, reservation short_id, tags, plan name, plan slug, facility code, facility name, operating system name, operating system slug, IP addresses.",
+* `filter` - (Optional) One or more attribute/values pairs to filter off of
+  - `attribute` - (Required) The attribute used to filter. Filter attributes are case-sensitive
```
displague, line 56: Which attributes are filterable?

equinix/helpers_device.go
```diff
@@ -271,3 +319,42 @@ func ipAddressSchema() *schema.Resource {
...
+	})
+	keyIDs := []string{}
+	for _, k := range device.SshKeys {
+		keyIDs = append(keyIDs, path.Base(k.GetHref()))
+	}
+	ports := getPortsMetalGo(device.NetworkPorts)
+
+	return map[string]interface{}{
+		"hostname":            device.GetHostname(),
+		"project_id":          *device.GetProject().Id,
```
displague, line 340: Is this possible, for this line and the others, as a way to have safer dereferencing:

```suggestion
		"project_id":          device.GetProject().GetId(),
```

docs/data-sources/equinix_metal_devices.md
```diff
@@ -0,0 +1,67 @@
+---
+subcategory: "Metal"
+---
+
+# equinix_metal_devices
+
+Provides an Equinix Metal devices datasource. This can be used to find devices that meet a filter criteria.
```
displague, line 7: Perhaps we should mention that `equinix_metal_device` is also available for fetching a single device.  Likewise, perhaps we should add the reverse note in `equinix_metal_device` docs.

displague:
looks like there is conflict in need of a rebase here.

equinix/helpers_device.go
ctreatma: Chained lookups should look like this:
```suggestion
		"project_id":          device.Project.GetId(),
```

`device.Project` is a `*Project`, which has [`GetId()` with the necessary nil checks](https://github.com/equinix-labs/metal-go/blob/main/metal/v1/model_project.go#L226-L232)

`device.GetProject()` returns a `Project`, which doesn't have `GetId()`

t0mk:
@displague I forgot about this PR, but I fixed your comments.

There's something wrong with the tests, I can't see the new test even being run. I think it's because the CI never gets to it, some tests time out and then fail, and the new metal_devices datasource test is never executed.

equinix/helpers_device.go
displague: the `Project` value can use the `*Pointer` pointer receivers: https://go.dev/tour/methods/4

ctreatma: I haven't tried it specifically in this PR, but when I tried [the similar suggestion on a different PR](https://github.com/equinix/terraform-provider-equinix/pull/369#discussion_r1320210247) I got a build error `equinix/helpers_device.go:176:26: cannot call pointer method GetMac on v1.PortData`.  What am I doing wrong?

ctreatma:
I missed a change in the acceptance test workflow when converting it from the `pull_request` trigger to the `pull_request_target` trigger.  PR to fix that is here: https://github.com/equinix/terraform-provider-equinix/pull/381

Once that's merged you can rebase this PR and your new test should run in CI

equinix/helpers_device.go
ctreatma: It looks like I'm not doing anything wrong, and the error is expected in this scenario: https://stackoverflow.com/a/44543748

Summarizing from there, the return value of `GetProject()` isn't addressable, so Go can't automatically pointerize it to look up the pointer receiver as a convenience.

Since the `GetX()` functions are written to handle a nil receiver, and since this function receives a `Device` and not a `*Device`, the `device.Project.GetId()` call here is nil-safe.  IMO there's no benefit to doing the extra work we would need to do in order to get a pointer for the value returned by `device.GetProject()`.

displague: Interesting and well enough. I tried a few combinations myself: https://go.dev/play/p/YcDgg2oELju (go.dev won't download metal-go successfully)

t0mk:
> I missed a change in the acceptance test workflow when converting it from the `pull_request` trigger to the `pull_request_target` trigger. PR to fix that is here: #381
> 
> Once that's merged you can rebase this PR and your new test should run in CI

thank you for letting me know

t0mk:
@displague The DataSourceMetalDevices test passed now. I resolved all your review comments. Please approve, so we can move on with this.


https://pipelinesghubeus2.actions.githubusercontent.com/HzLgMX4ejGF2QGyYnimawHYb58Or3PZJTkdVvJoe2T1r538sTo/_apis/pipelines/1/runs/1027/signedlogcontent/17?urlExpires=2023-09-21T14%3A13%3A55.5264932Z&urlSigningMethod=HMACV1&urlSignature=LONF1Mea8aYanOefVgaqy11Y1QE5W4VCL%2FczMR2E1no%3D

displague (approved):

PR #369, opened 2023-08-30
[WIP] Use metal-go for device resource

ctreatma:


equinix/resource_metal_device_acc_test.go
```diff
@@ -1060,6 +1061,16 @@ func mockFuncNotImplemented(f string) error {
 
 var _ packngo.DeviceService = (*mockDeviceService)(nil)
 
+type mockDevicesApi struct {
```
ctreatma, line 1064: Not sure exactly where I left this conversion, but I think it'd be better to update the tests to use a mocked API rather than mock whatever SDK is currently in use.

That should be doable with `httptest`, based on this: https://medium.com/zus-health/mocking-outbound-http-requests-in-go-youre-probably-doing-it-wrong-60373a38d2aa#fb9e

Based on that post I think we'd need to set up a mock API with `httptest`, configure the test server to handle the expected API requests, and include a `provider` block in the test config that points the provider at the `httptest` server.

ctreatma: See #371 for the `httptest` change

equinix/resource_metal_device.go
```diff
@@ -607,9 +612,9 @@ func resourceMetalDeviceCreate(ctx context.Context, d *schema.ResourceData, meta
 
 func resourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta interface{}) error {
 	meta.(*Config).addModuleToMetalUserAgent(d)
-	client := meta.(*Config).metal
+	client := meta.(*Config).metalgo
```
ctreatma, line 615: I haven't converted the create/update/delete functions yet; we could _potentially_ migrate this resource (and resources in general) from `packngo` to `metal-go` one function at a time in order to have smaller PRs that are easier to review.  The major complication with that is if we have tests that expect a specific SDK to be used, but that complication is already mentioned elsewhere in this PR.

equinix/helpers_device.go
```diff
@@ -95,6 +97,48 @@ func getNetworkInfo(ips []*packngo.IPAddressAssignment) NetworkInfo {
...
+			"address": *ip.Address,
+			"gateway": *ip.Gateway,
+			"family":  *ip.AddressFamily,
+			"cidr":    *ip.Cidr,
+			"public":  *ip.Public,
+		}
+		ni.Networks = append(ni.Networks, network)
+
+		// Initial device IPs are fixed and marked as "Management"
+		if !ip.HasManagement() || *ip.Management {
```
displague, line 125: I'm not 100% sure what the goal of this logic is: `management == null || management == true` ?

We have additional nil-safe helpers generated that may be able to simplify these conditions (as with the HasPublic check): `ip.GetManagement()` https://pkg.go.dev/github.com/equinix-labs/metal-go@v0.19.1/metal/v1#IPReservation.GetManagement

If I am misinterpreting the logic, GetManagementOk() may be more fitting.

`*ip.AddressFamily == int32(4)` could be `ip.GetAddressFamily() == 4`

displague: Same concern:

`!ip.HasPublic() || *ip.Public` sounds like `ip.public == nil || ip.public == true`, I think the intention is (pseudo-code) `ip.public != nil && ip.public == true`.

`if ip.GetPublic() {}` should handle that case.

equinix/helpers_device.go
```diff
@@ -122,6 +166,21 @@ func getPorts(ps []packngo.Port) []map[string]interface{} {
...
 }
 
+func getMetalGoPorts(ps []metalv1.Port) []map[string]interface{} {
+	ret := make([]map[string]interface{}, 0, 1)
+	for _, p := range ps {
+		port := map[string]interface{}{
+			"name":   p.GetName(),
+			"id":     p.GetId(),
+			"type":   p.GetType(),
+			"mac":    p.Data.GetMac(),
```
displague, line 176: Data is `*PortData`, it could technically be null.
`p.GetData.GetMac()` would be nil-safe, at the risk of returning "" for mac and false for bonded if Data was not included in the response (or testing Port object). 

These zero values seem like they would be appropriate if `data` was not included in the response, for whatever reason.

equinix/resource_metal_device.go
```diff
@@ -410,6 +410,11 @@ func resourceMetalDevice() *schema.Resource {
 					},
 				},
 			},
+			"sos_hostname": {
```
displague, line 413: nice! we'll want to include this in the datasource and the docs/ for both resources.

equinix/resource_metal_device.go
displague: After the conversion, I vote we strip out the "Go" in these function names

equinix/resource_metal_device.go
```diff
@@ -625,23 +630,24 @@ func resourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta i
...
-	d.Set("state", device.State)
-	d.Set("billing_cycle", device.BillingCycle)
-	d.Set("locked", device.Locked)
-	d.Set("created", device.Created)
-	d.Set("updated", device.Updated)
-	d.Set("ipxe_script_url", device.IPXEScriptURL)
-	d.Set("always_pxe", device.AlwaysPXE)
-	d.Set("root_password", device.RootPassword)
-	d.Set("project_id", device.Project.ID)
+		d.Set("metro", device.Metro.GetCode())
```
displague, line 638: Should be safe and equivalent to use `d.Set("metro", device.GetMetro().GetCode())` without the `if` check.

equinix/resource_metal_device.go
```diff
@@ -655,10 +661,13 @@ func resourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta i
 		d.Set("storage", storageString)
 	}
 	if device.HardwareReservation != nil {
-		d.Set("deployed_hardware_reservation_id", device.HardwareReservation.ID)
+		d.Set("deployed_hardware_reservation_id", device.HardwareReservation.GetId())
```
displague, line 664: This may also be a case to simplify with `GetHardwareReservation().GetId()` (removing the `if` clause.

equinix/helpers_device.go
ctreatma: `.GetMac()` and `.GetBonded()` have pointer receivers, so they cannot be chained on `GetData()`.  We'd have to do `GetData().Mac` and `GetData().Bonded`, which are `*string` and `*bool` respectively, so we'll need to break these calls out of the map declaration so we can check for nil and use an appropriate fallback.

equinix/resource_metal_device.go
ctreatma: Yeah, the suffix is only here to keep the code paths separate and should be removed once it's no longer needed.

equinix/resource_metal_device.go
ctreatma: The datasource gains this attribute in #291; I broke these apart to keep smaller changes, but that was because I was worried about fixing the mock tests as part of the conversion.  Now that #371 is in, this could probably be collapsed back into #291

equinix/helpers_device.go
ctreatma: Or...actually, I think we don't need a fallback, and `p.Data.Get<Blah>()` is doing what we want it to, because if `Data` is a nil `*PortData`, then e.g. [`p.Data.GetMac()` will return `""`](https://github.com/equinix-labs/metal-go/blob/main/metal/v1/model_port_data.go#L51).

equinix/helpers_device.go
ctreatma: Yeah no idea what the aim was with these conditions but the `Get` helpers should be analogous to what the packngo version of this function was doing.

equinix/resource_metal_device.go
ctreatma: Oh right, the other reason this PR is small is because so far it only converts the read function.  If we're good with using 2 clients in a single resource, though, maybe it's worth combining this with the device data source PR and then follow up in a separate PR to convert the create/update/delete functions.

equinix/resource_metal_device.go
ctreatma: The `if` clause here is not specific to the SDK conversion

ctreatma:
Closing in favor of #291; that PR now updates both the datasource and the device read pathway, so it covers everything that was in this PR so far.

PR #371, opened 2023-08-30
test: refactor device resource tests to mock API instead of SDK

ctreatma:
We have a few tests for the device resource that cover scenarios that are impossible to reliably recreate in an end-to-end test.  These tests were originally based on a mock packngo client, but mocking the SDK hides issues that may come in due to SDK upgrades and requires changing tests when we switch to a different SDK.

This updates the device mock tests to use a mock API created with the built-in `httptest` module so that we can more easily validate that the tests continue to pass after upgrading or changing the SDK.

### DIFF
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -6,11 +6,13 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"regexp"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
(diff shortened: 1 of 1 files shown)


displague (approved):
lgtm. I was a little surprised that we didn't have a test that included an `{"errors":[...]}` body (likely just not needed with failing headers).
```
--- PASS: TestAccMetalDevice_readErrorHandling (0.00s)
    --- PASS: TestAccMetalDevice_readErrorHandling/forbiddenAfterProvision (0.00s)
    --- PASS: TestAccMetalDevice_readErrorHandling/notFoundAfterProvision (0.00s)
    --- PASS: TestAccMetalDevice_readErrorHandling/forbiddenWaitForActiveDeviceProvision (0.00s)
    --- PASS: TestAccMetalDevice_readErrorHandling/notFoundProvision (0.00s)
    --- PASS: TestAccMetalDevice_readErrorHandling/errorProvision (0.00s)
    ```

PR #374, opened 2023-09-01
ci: run acctest workflows 1 at a time and support external PRs

ctreatma:
We don't need to hammer the API with end-to-end tests when the repo gets a flurry of activity.  This adds a concurrency setting to our acceptance test workflow with a hard-coded string value so that only one run of the acceptance tests can execute at any time.

This also updates the acceptance test workflow to replace the `pull` trigger with a `pull_request_target` trigger.  We've been using this for `metal-cli` and it has been a good experience thus far.  The new trigger allows us to manually run end-to-end tests for PRs from forks in order to build confidence in code changes before they are merged.

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -1,6 +1,14 @@
 name: Acceptance Tests
+# This workflow determines whether a PR comes from an external fork
+# (which requires approval from us) or from a branch on this repository
+# (which means it was made by us and can run immediately).  Once a PR
+# is approved, the PR code gains access to secrets referenced in this
+# workflow.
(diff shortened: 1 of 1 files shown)


ocobles (dismissed):
lgtm!

ctreatma:
@ocobles (no rush!  can wait 'til next week): I missed a duplicate `concurrency` setting in the workflow file and had to update the PR, which dismissed your review.  When you get a chance could you take another look?

"Successful" workflow run is here (the workflow works but the tests, as usual, have some flakes): https://github.com/equinix/terraform-provider-equinix/actions/runs/6052317510

ocobles (approved):

PR #380, opened 2023-09-14
fix: Timeout for metal virtual circuits

aayushrangwala:
https://github.com/equinix/terraform-provider-equinix/issues/357

### DIFF
--- a/equinix/data_source_metal_virtual_circuit.go
+++ b/equinix/data_source_metal_virtual_circuit.go
@@ -1,12 +1,14 @@
 package equinix
 
 import (
+	"context"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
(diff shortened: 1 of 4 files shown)


ctreatma:
@aayushrangwala could you add a Conventional Commits tag to the PR title?  Tags are described in [the comment above](https://github.com/equinix/terraform-provider-equinix/pull/380#issuecomment-1719733257).  For this I think `fix:` would make sense but a different one might work.

aayushrangwala:
> @aayushrangwala could you add a Conventional Commits tag to the PR title? Tags are described in [the comment above](https://github.com/equinix/terraform-provider-equinix/pull/380#issuecomment-1719733257). For this I think `fix:` would make sense but a different one might work.

yes added fix for all the related PRs

ctreatma (approved):

PR #386, opened 2023-09-20
fix: Remove timeout for metal vrf

aayushrangwala:
As part of fixing the timeouts for metal resources to actually comply with what is documented, it was found out that we donot document any timeout support for metal VRF whereas in the code we do have support for timeouts.

Since it has never been working due to the below issue, also not documented, it'll be ideal to remove it directly considering a code smell

Relates to: #357

### DIFF
--- a/equinix/resource_metal_vrf.go
+++ b/equinix/resource_metal_vrf.go
@@ -1,26 +1,20 @@
 package equinix
 
 import (
-	"log"
-	"time"
-
+	"context"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/packethost/packngo"
+	"log"
 )
 
 func resourceMetalVRF() *schema.Resource {
 	return &schema.Resource{
-		Timeouts: &schema.ResourceTimeout{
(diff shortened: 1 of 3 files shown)


equinix/resource_metal_vrf.go
```diff
@@ -1,26 +1,20 @@
...
-
+	"context"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/packethost/packngo"
+	"log"
 )
 
 func resourceMetalVRF() *schema.Resource {
 	return &schema.Resource{
-		Timeouts: &schema.ResourceTimeout{
```
ctreatma, line 13: This is removing timeouts entirely rather than fixing them.  I don't necessarily think that's an issue:  as I understand it, they didn't work and they weren't documented, so IMO it's just as valid to remove them as it is to make them work.  However, if we're removing them, then the PR title should be updated to reflect that.

aayushrangwala: Yes thats exactly correct, added the description in PR description as well

ctreatma (approved):

PR #387, opened 2023-09-20
Timeout fix metal network device

aayushrangwala:
#357


ctreatma (commented):
These are Network Edge resources, not Metal resources, and they're not covered by tests, so I think it would be preferable to leave them as-is and work with the Network Edge team to fix them when they have capacity to work with us.

PR #388, opened 2023-09-20
fix: Timeout for metal ipblock

aayushrangwala:
#357

### DIFF
--- a/equinix/resource_metal_reserved_ip_block.go
+++ b/equinix/resource_metal_reserved_ip_block.go
@@ -1,6 +1,7 @@
 package equinix
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"log"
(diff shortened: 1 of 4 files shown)


equinix/resource_metal_reserved_ip_block.go
```diff
@@ -215,12 +216,12 @@ func resourceMetalReservedIPBlock() *schema.Resource {
 	}
 	// TODO: add comments field, used for reservations that are not automatically approved
 	return &schema.Resource{
-		Create: resourceMetalReservedIPBlockCreate,
-		Read:   resourceMetalReservedIPBlockRead,
-		Update: resourceMetalReservedIPBlockUpdate,
-		Delete: resourceMetalReservedIPBlockDelete,
+		CreateContext:        diagnosticsWrapper(resourceMetalReservedIPBlockCreate),
+		ReadWithoutTimeout:   diagnosticsWrapper(resourceMetalReservedIPBlockRead),
+		UpdateWithoutTimeout: diagnosticsWrapper(resourceMetalReservedIPBlockUpdate),
```
ctreatma, line 221: I think it would be good to have a timeout on update and delete; is there a platform limitation or other reason why we can't/shouldn't support timeouts on those operations?

aayushrangwala: Since the default timeout has only been configured/exposed for `create` the create with context is used
`Timeouts: &schema.ResourceTimeout{
    Create: schema.DefaultTimeout(ReservedIPCreateTimeout),
},`

ctreatma (approved):
👍 I see all tests starting with `TestAccMetalReservedIPBlock` (including the new one) are passing in CI.

Issue #384, opened 2023-09-18
Improve discoverability and usability of examples/


displague:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

In #100, we moved the Fabric examples/ to their own GitHub repos. We've slid back from this position in recent updates. This can be seen in the project today with some inconsistent between resource families:
- Network Edge: examples/ directory contains a single README.md that refers to Terraform modules in the registry
   https://github.com/equinix/terraform-provider-equinix/tree/main/examples/edge-networking#readme
- Fabric: Documentation (Guide) includes a module example section that refers to Terraform modules in the registry 
   https://registry.terraform.io/providers/equinix/equinix/latest/docs/guides/equinix_fabric_cloud_providers#terraform-modules---the-easiest-way
- Fabric: examples/ nested directory tree with per resource, use-case, examples. This pattern is evidenced in AzureRM, AWS, and (to a lesser-extent) Google providers maintained by Hashicorp.
   https://github.com/equinix/terraform-provider-equinix/tree/main/examples/fabric
- Metal: none (there are some guides that walk through use-cases, however there are few if any single resource example modules found in the two GitHub orgs that contain the most Equinix Metal modules ([Equinix Labs](https://github.com/orgs/equinix-labs/repositories?q=terraform-&type=all&language=&sort=), [Equinix](https://github.com/orgs/equinix/repositories?q=terraform-&type=all&language=&sort=))

The main advantage of having the examples in the terraform-provider-equinix examples/ directory is that they are easier to maintain when making provider changes.

The main advantage of having examples included or referenced in the docs/ directory is that they are discoverable and alongside the primary reference for the resource.

The main advantage of having examples published in modules is that they can be consumed by Terraform directly. Their use can be measured. They can be modified to continue working even when the underlying resources have changed.

In order to make our examples more discoverable, usage measurable, reusable, and well maintained, the community would benefit from:
- [x] the examples/fabric/ paths being moved to distinct Terraform Modules (derived from the template: https://github.com/equinix-labs/terraform-equinix-template)
[...]

PR #397, opened 2023-09-29
feat: Add Port2Azure Redundant Connection Example

thogarty:
* Add redundant connection example
* Update README.md for both single and redundant connections
* Update unused variables for both connections

In regards to https://github.com/equinix/terraform-provider-equinix/issues/384, we will start to use the `_e2e_` tests for regression testing for Fabric resources. At first locally but hopefully in GHA if we can get connecting to the lower environments from there. 

@manu-equinix was of the mind that we could access UAT environments from GHA.


examples/fabric/v4/portConnectivity/azure/twoRedundantConnections/terraform.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_providers {
+    equinix = {
+      source = "equinix/equinix"
+    }
+  }
+}
```
srushti-patl, line 7: Trailing space is missing

srushti-patl (approved):
LGTM

PR #398, opened 2023-10-02
chore: use terraform_data resource to stabilize test selections

ctreatma:
Occasionally, we see metro or facility selections change as a result of capacity changes on the Metal platform side; for example, step 1 of a test deploys to `da`, but when step 2 runs, there are no longer any available servers in `da`, so a different metro is selected and the test generates an unexpected plan.

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -71,7 +71,7 @@ jobs:
         version:
           - stable
         terraform:
-          - '1.1.5'
+          - '1.5'
     steps:
 
     - name: Check out code into the Go module directory
--- a/equinix/resource_metal_device_acc_test.go
+++ b/equinix/resource_metal_device_acc_test.go
@@ -118,12 +118,28 @@ resource "random_integer" "plan_idx" {
   max = length(data.equinix_metal_plans.test.plans) - 1
 }
 
+resource "terraform_data" "plan" {
+  input = data.equinix_metal_plans.test.plans[random_integer.plan_idx.result]
+
+  lifecycle {
+	ignore_changes = ["input"]
+  }
+}
+
 // Select a metal facility randomly and lock it in
 // so that we don't pick a different one for
 // every subsequent terraform plan
 resource "random_integer" "facility_idx" {
   min = 0
   max = length(local.facilities) - 1
+}
+
+resource "terraform_data" "facility" {
+  input = local.facilities[random_integer.facility_idx.result]
+
+  lifecycle {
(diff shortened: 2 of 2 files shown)


.github/workflows/acctest.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         version:
           - stable
         terraform:
-          - '1.1.5'
+          - '1.5'
```
ctreatma, line 74: The `terraform_data` resource was introduced in 1.4, so we need to run tests in 1.4+ in order to use that resource.  I _think_ this might need to go in a separate PR so that it can be merged to `main` separately because we use the `pull_request_target` trigger.

equinix/resource_metal_device_acc_test.go
```diff
@@ -118,6 +118,14 @@ resource "random_integer" "plan_idx" {
   max = length(data.equinix_metal_plans.test.plans) - 1
 }
 
+resource "terraform_data" "plan" {
+  input = data.equinix_metal_plans.test.plans[random_integer.plan_idx.result]
+
+  lifecycle {
+	ignore_changes = ["input"]
```
displague, line 125: This is a nice trick to effectively cache the result of the data source. Any good links to check out (curious if there are other tricks like this given).

displague (approved):

.github/workflows/acctest.yml
displague: Do you want to merge a PR with that change first so we can see these tests pass?

ctreatma: I kicked off a manual run of the updated job, so we can wait to see what that does: https://github.com/equinix/terraform-provider-equinix/actions/runs/6411225404/job/17406201790

ctreatma: Since the tests are flaky even without the caching introduced here, I re-ran them a couple times to build confidence that these changes are not causing new test failures.  I don't see anything to indicate that this change is bringing in _new_ problems, and it does appear to stabilize the tests in terms of avoiding unnecessary plan changes due to changes in plans API responses due to platform capacity.

equinix/resource_metal_device_acc_test.go
ctreatma: The only reference I found for this resource was [the docs page on the terraform developer site](https://developer.hashicorp.com/terraform/language/resources/terraform-data).  It doesn't specifically cover this use case of caching values, but it does include a similar trick for using `lifecycle` to trigger replacement of a different resource.

Issue #365, opened 2023-08-24
Migrate to terraform plugin framework


ctreatma:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

Our Terraform provider is currently based on terraform-plugin-sdk which is no longer prioritized by Hashicorp and may not receive new features going forward.

The new hotness is the terraform plugin framework, which is getting the bulk of Hashicorp’s attention and has some useful things like built-in attribute validation so we can, for example, validate that strings are a certain length or match a regex without writing our own validation functions.

There is a migration guide here: https://developer.hashicorp.com/terraform/plugin/framework/migrating 

Given the number of resources we have, we may want to use muxing (documented in the link above) to gradually convert the provider.  Testing will be an issue because (1) Metal tests can be flaky and (2) Fabric and NE tests don’t exist (or at least, not in a way that we can run them ourselves on demand).

displague:
I think we need to do #106 as a requirement of this work because sub-directory packages are part of Framework's best-practices. If Framework can be introduced without moving the resources to separate directories, this is not a concern.

Example of the cyclic dependencies that we'll encounter if we don't move helpers first:
- internal/services/metal/devices/resource.go import /equinix to get SomeHelper
- equinix/provider.go import internal/services/metal/devices to get device.Resource

ocobles:
I have created an [issue](https://github.com/equinix/terraform-provider-equinix/issues/612) to track the migrated resources, can we close this one?

ctreatma:
As originally written, this issue has been completed: we've introduced muxing to support gradual migration of SDKv2 resources to the terraform-plugin-framework provider and have migrated a few resources to that provider.  The process of migrating the remaining resources out of SDKv2 is captured in the issue Oscar linked above.

PR #406, opened 2023-10-10
feat: stub for framework provider, mux, frameworking resource and datasource for metal_ssh_key

t0mk:
This PR is a first step towards migration from TF SDKv2 to the plugin Framework. It's along way, and Hashicorp documents it at [0].

We want to have both types of providers exist in the repo until we can totally migrate to framework. We need to use "muxing" [1].

I base this on the continuous migration of the Linode TF provider [2]. [3] is a first PR where they introduced the framework provider and migrated `token` resource to the framework format.

For our case, I selected `metal_ssh_key` resource and datasource to migrate here, because those are simple resource types.

[0] https://developer.hashicorp.com/terraform/plugin/framework/migrating
[1] https://developer.hashicorp.com/terraform/plugin/framework/migrating/mux
[2] https://github.com/linode/terraform-provider-linode
[3] https://github.com/linode/terraform-provider-linode/pull/791

Towards #365


t0mk:
Thsi is a WIP. I wonder if there's a way not to have acceptance tests run on every commit.

ocobles:
> Thsi is a WIP. I wonder if there's a way not to have acceptance tests run on every commit.

I think we can skip test exuction for draft PRs. As I understand it, defining all required types in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target without `synchronize` type, won't run the test till you change it to `ready_for_review`. A similar approach with more control could be adding the condition in the job, something like:

`if: ${{ github.event_name == 'pull_request_target' && github.event.action == 'synchronize' && !github.event.pull_request.draft }}`

Thoughts on this @ctreatma ?

displague:
I do have, from time to time, drafts where I want to run E2E tests. Perhaps a label (No E2E Testing) or the title (WIP or NOE2E) could signal this?

ocobles:
you can still run it manually since that condition is only for `github.event_name == 'pull_request_target'` and not for `workflow_dispatch`

ctreatma:
Yeah, @ocobles I think your suggestion is good.  We should go with the second approach, adding a condition to the job (either by updating the `authorize` check or adding a separate check job before it).  I don't think we should check for the `synchronize` action in the condition, though; if we want to limit the job to specific action names lets add that to the trigger config (`synchronize` and `opened` might be the only ones we really need?).

t0mk:
I'm now doing the migration of the provider, and I also want to move resources and datasources to subdirectories, e.g `equinix/metal_ssh_key/{resource,datasource}.go`

The framework provider needs to refer to owned resource [0], and if we have resource in subdir, it needs to import the subdir (see top of `framework_provider.go`).

I run into issues with the fact that we have a lot of helper code straight in the `github.com/equnix/terraform-provider-equinix/equinix` namespace. The code of a resource, for example in `equnix/metal_ssh_key/resource.go` needs to import the Config struct which is in `github.com/equnix/terraform-provider-equinix/equinix` and since the provider is importing the subdir package, there's an import cycle.

I am not sure how to resolve it. The natural solution would be to move the config (and all the helpers that resources and Config are using) to `equinix/internal`. Then both provider and resource code could use them without cycles. But that is a huge change, affecting every reference to Config or the helper functions, and should be agreed with other contributors. Also I think it should be done separately from the migration work.

We need to break the cycle provider@equinix -> resource@equinix/metal_ssh_key/ -> config_struct,helpers@equinix

It could be done without breaking things, if we **copy** all the necessary code to `equinix/internal/`. Then, existing unrelated code will not have to be changed along the migration, but if there will be work done in parallel (to the migration) on helpers in original location, it will be hard to merge once we get rid of the SDKv2 resources.

I think the helpers and config must be moved to a subdir eventually if we want to have resource code in subdirs. Or is there another solution in Go that I can't see? My suggestion is to move the helpers and config first, and continue with migration later.

Also, it's important to do the move of the config and helpers **properly**, so that really all the functionality (and its dependencies) that resources need goes to a subdir. It should be done somehow aware of the framework provider and resources in subdir. Along with the move to `internal/`, we should move one existing SDKv2 resource to a subdir so that we are sure that we restructured things properly.

I need your feedback @ctreatma @displague .

Thanks for reading this.

[0] https://github.com/equinix/terraform-provider-equinix/blob/framework_migr_base_plus_ssh_key/equinix/framework_provider.go#L93

t0mk:
@displague I see, you moved everything but provider.go provider_test.go and equinix_sweeper_test.go from `<repo_root>/equinix` to `<repo_root>/internal` and structured it a lot in there. Is it possible you wouldn't run into the cycle in tests if you'd move it to `<repo_root>/equinix/internal`? That's how I see it in Linode TP.

Anyway, it looks like a way to go, and too bad you didn't manage to merge it last year.

I don't think we need to move all resources to subdirs at this point, since we will migrate them to framework plugin ecventually. We just need one as a tracing bullet, to see that most of the helper code is separated from `equinix` namespace.

Give me green light on the code move and I will do a PR, restructure it in a reasonable way, and then we will discuss the paths and directory structure there, as that's not too relevant to the migration.

ctreatma:
@t0mk, yes, you can go ahead with the code move.  If you're looking for a resource / datasource to use as a testbed, the `metal_ssh_key` resource & datasource used here would be good candidates; step 1 would be refactoring within the SDKv2 implementation, and then step 2 would be migrating the refactored implementation to the plugin framework.

t0mk:
@ctreatma , great! I will do so. I was thinking about metal_ssh_key myself. Actually we don't even have a datasource for the user key.

displague:
My PR got too big too fast.
One piece that would be a nice chore that could be applied independently of resource moves and framework conversion is to move the helpers to various type specific packages.

equinix/framework_provider.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+
+	"github.com/equinix/terraform-provider-equinix/equinix/metal_ssh_key"
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/eqinix/terraform-provider-equinix/equinix/internal"
```
displague, line 14:
```suggestion
	"github.com/equinix/terraform-provider-equinix/equinix/internal"
```

ocobles:
Are there any disadvantages to moving into subfolders by service one level above the split by resource? I mean, `internal/metal/ssh_key/` or `internal/fabric/connection/`. That approach keeps the main directory cleaner and helps limit the workspace for teams focused on a single service

ocobles:
Currently, we have a datasource for `metal_project_ssh_key` and resources for `metal_project_ssh_key` and `metal_ssh_key`. I see you have created a folder `metal_ssh_key` with both resource and datasource files. Is that datasource a new one to cover the GET user ssh key API request, or did you accidentally mix the resource for user ssh key and datasource for project ssh key?

ocobles:
Hi @t0mk, there are a few things I found while migrating resources that need to be addressed here:

- While they coexist, both provider schemas must produce the same output. So if you add a description to `max_retries` or `max_retry_wait_seconds` make sure it is included also in the old sdkv2 provider schema
- `tfsdk` struct tag doesn't have `omitempty` augmentation
- `id` attribute must be defined in the resource schemas
- A solution to add module to user agent:

//resource.go
```
func (r *Resource) Create(
	ctx context.Context,
	req resource.CreateRequest,
	resp *resource.CreateResponse,
) {

	r.Meta.AddFwModuleToMetalUserAgent(ctx, req.ProviderMeta)
	client := r.Meta.Metal
```

//config.go
```
func (c *Config) AddFwModuleToMetalUserAgent(ctx context.Context, meta tfsdk.Config) {
	c.Metal.UserAgent = generateFwModuleUserAgentString(ctx, meta, c.metalUserAgent)
}

func (c *Config) AddFwModuleToMetaGolUserAgent(ctx context.Context, meta tfsdk.Config) {
	c.Metalgo.GetConfig().UserAgent = generateFwModuleUserAgentString(ctx, meta, c.metalGoUserAgent)
}

func generateFwModuleUserAgentString(ctx context.Context, meta tfsdk.Config, baseUserAgent string) string {
	var m ProviderMeta
	diags := meta.Get(ctx, &m)
	if diags.HasError() {
		log.Printf("[WARN] error retrieving provider_meta")
		return baseUserAgent
	}
	if m.ModuleName != "" {
		return strings.Join([]string{m.ModuleName, baseUserAgent}, " ")
	}
	return baseUserAgent
}

```

ocobles:
more findings @t0mk : 

- I think there's no schema URL validator in the framework and it seems documentation https://developer.hashicorp.com/terraform/plugin/framework/migrating/providers#framework-4 is wrong. You can create a custom validator but I would rather use RegMatches for the `endpoint` field

```
import "github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"

...

var UrlRE = regexp.MustCompile(`^https?://(?:www\.)?[a-zA-Z0-9./]+$`)

...

"endpoint": schema.StringAttribute{
  Optional:    true,
  Description: "The Equinix API base URL to point out desired environment. Defaults to " + config.DefaultBaseURL,
  Validators: []validator.String{
    stringvalidator.RegexMatches(urlRE, "must be a valid URL with http or https schema"),
  },
},
```

ocobles:
that go test error after my last commit is related to https://developer.hashicorp.com/terraform/plugin/testing/migrating#flag-redefined-panic

I'm going to open a different PR to fix the code before continuing with the work here. So basically we need to replace any resource.StateChangeConf references with helper/retry package

internal/resources/metal/metal_project_ssh_key/framework_datasource.go
```diff
@@ -0,0 +1,99 @@
+package metal_project_ssh_key
```
displague, line 1: I think the directory and package name could be simplified to not repeat the parent directory, project_ssh_key, in this case.

Similarly, I don't see added value in the `framework_` prefixes. It may make it easier to find framework vs sdkv2 files, but the contents, types and methods, of those files are just as telling.

ocobles: addressed in https://github.com/equinix/terraform-provider-equinix/pull/406/commits/b8cc18d3e9e193006d8a6cbadbebd7f28a565bad

ctreatma:
> I replicated that logic in the new framework update function, but I'm wondering if that `ForceNew` is a bug or added on purpose. If it's a bug, should we fix it before/after merging this PR or even in this PR?

Since this behavior has been in place for a while, I say we keep it as-is until after the resource is migrated.

internal/resources/metal/project_ssh_key/models.go
```diff
@@ -0,0 +1,59 @@
...
+	m.Created = types.StringValue(key.Created)
+	m.Updated = types.StringValue(key.Updated)
+	m.OwnerID = types.StringValue(path.Base(key.Owner.Href))
+	m.ProjectID = m.OwnerID
+	return nil
+}
+
+// TODO (ocobles) ideally we would embed ResourceModel instead of
+// explicitly define all the ResourceModel fields again in DataSourceModel
+// https://github.com/hashicorp/terraform-plugin-framework/issues/242
```
ctreatma, line 36: I'm not clear why we need a separate `DataSourceModel`; it looks like these model types are only known to us and the framework doesn't care how we name or structure them.  Could the data source simply use the `ResourceModel`?

ctreatma: @ocobles pointed out that the `DataSourceModel` has a `search` property, and the `tfsdk` tag doesn't support optional fields, so we _do_ need these separate models for the resource and data source

main.go
```diff
@@ -1,29 +1,54 @@
...
-			log.Fatal(err.Error())
-		}
-		return
+			serveOpts = append(serveOpts, tf5server.WithManagedDebug())
 	}
 
-	plugin.Serve(opts)
-}
+	err = tf5server.Serve(
+			"registry.terraform.io/equnix/equinix",
```
ctreatma, line 46:
```suggestion
			"registry.terraform.io/equinix/equinix",
```

displague: If tests didn't / can't uncover this being wrong, we may want to see some manually testing to confirm that this provider works with a combination of framework and sdkv2 plugins.

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -0,0 +1,197 @@
...
+	ctx context.Context,
+	req resource.ReadRequest,
+	resp *resource.ReadResponse,
+) {
+	r.Meta.AddFwModuleToMetalUserAgent(ctx, req.ProviderMeta)
+	client := r.Meta.Metal
+
+	// Retrieve values from state
+	var state ResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
```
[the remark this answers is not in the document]
ocobles: `req.State.Get` retrieves current state and once populated info from the API we update the state with the retrieved information using `resp.State.Set`

internal/resources/metal/ssh_key/resource.go
```diff
@@ -1,4 +1,4 @@
-package metal_ssh_key
+package ssh_key
```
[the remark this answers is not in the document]
ocobles: added in [421d41d](https://github.com/equinix/terraform-provider-equinix/commit/421d41d62b7068fa97ddd66fbf9c6c20bfa77b9a)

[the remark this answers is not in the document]
ocobles: https://github.com/equinix/terraform-provider-equinix/pull/406/commits/469ee17d9bde408b69653e3f8ceabf21ba3fca17

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -0,0 +1,197 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Extract the ID of the resource from the state
+    id := state.ID.ValueString()
```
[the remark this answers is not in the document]
ocobles: https://github.com/equinix/terraform-provider-equinix/pull/406/commits/469ee17d9bde408b69653e3f8ceabf21ba3fca17

main.go
ocobles:
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create
 <= read (data resources)

Terraform will perform the following actions:

  # data.equinix_metal_project_ssh_key.foobar will be read during apply
  # (config refers to values not yet known)
 <= data "equinix_metal_project_ssh_key" "foobar" {
      + created     = (known after apply)
      + fingerprint = (known after apply)
      + id          = (known after apply)
      + name        = (known after apply)
      + owner_id    = (known after apply)
      + project_id  = "b3b111d4-a313-4132-a824-XXXX"
      + public_key  = (known after apply)
      + search      = (known after apply)
      + updated     = (known after apply)
    }

  # equinix_metal_project_ssh_key.foobar will be created
  + resource "equinix_metal_project_ssh_key" "foobar" {
      + created     = (known after apply)
      + fingerprint = (known after apply)
      + id          = (known after apply)
      + name        = "ocm-tf-test"
      + owner_id    = (known after apply)
      + project_id  = "b3b111d4-a313-4132-a824-f48a06cf7ca3"
      + public_key  = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDM/unxJeFq..."
      + updated     = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + test = (known after apply)

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

equinix_metal_project_ssh_key.foobar: Creating...
equinix_metal_project_ssh_key.foobar: Creation complete after 0s [id=651002ea-fbeb-4969-a3e6-XXXX]
data.equinix_metal_project_ssh_key.foobar: Reading...
data.equinix_metal_project_ssh_key.foobar: Read complete after 0s [id=651002ea-fbeb-4969-a3e6-XXXX]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

Outputs:

test = "48:b8:72:58:d5:32:36:45:f3:4d:b5:17:ef:82:50:e3"
```

here's some references in the output log to the tmux provider 
image

ocobles:
@ctreatma definitely #551 worked as expected, I don't see that error anymore. Here's is the specific tests for both migrated resources https://github.com/equinix/terraform-provider-equinix/actions/runs/7744881780/job/21119662842#step:6:4306 https://github.com/equinix/terraform-provider-equinix/actions/runs/7744881780/job/21119662842#step:6:4405

go.work.sum
```diff
@@ -0,0 +1,711 @@
+cloud.google.com/go v0.110.2/go.mod h1:k04UEeEtb6ZBRTv3dZz4CeJC3jKGxyhl0sAiVVquxiw=
```
ctreatma, line 1: Could you remove this file?

examples/fabric/v4/cloudRouterConnectivity/cloudRouter2serviceprofile/terraform.tfvars.example
```diff
@@ -1,4 +1,4 @@
-equinix_client_id      = "Equnix_Client_Id"
+equinix_client_id      = "Equinix_Client_Id"
```
ctreatma, line 1: Code owner review has been enabled on this repo so we'll either need to revert this or get a 👍 from the Fabric team

go.work.sum
ocobles: removed in https://github.com/equinix/terraform-provider-equinix/pull/406/commits/42f2da654a27f8c93a6220fc4a2fa8bb8b0107f1

examples/fabric/v4/cloudRouterConnectivity/cloudRouter2serviceprofile/terraform.tfvars.example
ocobles: reverted in https://github.com/equinix/terraform-provider-equinix/pull/406/commits/42f2da654a27f8c93a6220fc4a2fa8bb8b0107f1 since is a typo in an example value it could be addressed later in a different PR

ctreatma (approved):
Looks good!  I see all of the SSH key tests passing in CI (including the new upgrade test).

PR #415, opened 2023-10-16
chore(ci): avoid test changes due to plan facilities

ctreatma:
We still had some Metal test failures due to changes in the `facilities` list for a selected plan.  In these cases, the tests were directly referencing the full `facilities` list rather than the single, cached, `facility` value.

This updates the tests to also cache the `facilities` list in an attempt to further reduce flakiness.  In addition, some tests that were using `facilities` have been updated to use `metro` instead, since none of the functionality under test was reliant on using `facilities`.


displague (approved):
test failures don't appear related to this change

PR #420, opened 2023-10-18
refactor: Moving utility code to `internal/`

### LINKED ISSUE #106: adopt internal/ layout structure

t0mk:
Fixes #106


t0mk:
To move code for resource `metal_ssh_key` to a subdirectory, I had to
- create subdirs `metal_ssh_key/` and `metal_project_ssh_key/` and moved code of respective resources and datasources to the subdirectories. `provider.go` then imports the `metal_ssh_key` resource code from the sub-package. Since the module code uses the Config struct, I had to move the Config struct out of the `equinix/` namespace to avoid import cycle.
- I moved the Config struct to `internal/`. I had to move more code along to `internal/`.
- All around the provider source files, I had to fix a lot of references to the Config struct and also to other utilities which now live in `internal/`.
- At this point the code in `metal_ssh_key` subdir compiled OK.

Then I wanted to run tests in the subdir `metal_ssh_key`. I realized that if the tests live in the same package as the module code, it's not straightforward because the tests need to instantiate the TF Provider object defined in `equinix`, and thus close the import cycle that I worked so hard to break..

I found out yesterday (after 6 years of Golang programming) that it's possible to have packages `foo` and `foo_test` live in the same directory `foo/`. Then, `import foo` will not import `foo_test`. So I just put `package  metal_ssh_key_test` to the top of source files which contain the tests. Then, the main package didn't import them, and I could instantiate the provider object for acceptance tests. It's not necessary to import the `metal_ssh_key` package from the tests because the tests only use TF templates and don't call resource code directly.

There is a lot of shared functionality for acceptance tests, and that utility code can't live in `internal/`, because the main package imports `internal/`. I created `acceptance/` subdir where I placed the test helpers.

I had a feedback from @ctreatma not to touch the `_fabric_` code and resources, but it was not possible to make all this without changing the references to Config and some helper functions in most of the source files, including fabric stuff.

We do all this in order to migrate to framework plugin eventually [0].

Let's discuss the format! I will keep this up to date (rebasing) with main.

[0] https://github.com/equinix/terraform-provider-equinix/pull/406#issuecomment-1764477274

equinix/fabric_service_profile_schema.go
```diff
@@ -1,8 +1,6 @@
 package equinix
 
-import (
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-)
+import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
```
ctreatma, line 3: There are a few stylistic changes like this that distract from the refactoring; could you revert them?

equinix/fabric_mapping_helper.go
```diff
@@ -1209,3 +1209,19 @@ func CorrelationIdWithCharset(length int, charset string) string {
 func CorrelationId(length int) string {
 	return CorrelationIdWithCharset(length, allowed_charset)
 }
+
+func additionalInfoContainsAWSSecrets(info []interface{}) ([]interface{}, bool) {
```
ctreatma, line 1213: It looks like this was moved even though it's staying in the `equinix` package; we should only move things if it's necessary to complete the refactor.

equinix/fabric_connection_schema.go
ctreatma, line 1: There are a bunch of file mode changes; can you roll those back?

ctreatma (commented):
We have a big ball of mud right now, so it probably won't be possible to avoid touching NE / Fabric stuff _at all_ during the refactor.  So far, this looks targeted enough in terms of those changes; I did comment on a couple places where the changes aren't strictly needed for refactoring.

We should add more package isolation as part of this refactor so that we don't end up with a new ball of mud in a different directory: config in a `config` package, errors in an `error` package, etc..

equinix/fabric_connection_schema.go
t0mk: Nice catch, I didn't notice that. But I'm reluctant to set executable flag on 18 Golang source code files. I'd prefer to fix the wrong mode in `main` and then rebase. I did a PR to fix the modes:
https://github.com/equinix/terraform-provider-equinix/pull/450

equinix/fabric_service_profile_schema.go
t0mk: I reverted the stylistic changes.

equinix/fabric_mapping_helper.go
t0mk: I moved the function to it's original place, and not it doesn't show in the diff.

equinix/fabric_connection_schema.go
t0mk: This was resolved by merging https://github.com/equinix/terraform-provider-equinix/pull/450

equinix/internal/fabric_cloud_router_read_schema.go
```diff
@@ -1,4 +1,4 @@
-package equinix
+package internal
```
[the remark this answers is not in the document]
displague: Looking at the AWS structure (which is not adapted to Framework), we see they chose the paths: internal/service/foo/: https://github.com/hashicorp/terraform-provider-aws/tree/main/internal/service/ec2/ec2_instance.go

Linode took another approach. they didn't use `internal/` and they didn't create namespaces for the resources: https://github.com/linode/terraform-provider-linode/tree/dev/linode/instance

In my old PR, I went weith internal/resources/{service:metal}/{resource:device}/resource.go

Namespaces (metal, fabric, ne)  will be important for how we collaborate with peer teams on these resources. 
We'll also want per resource directories.
We'll also want to use `internal/` to keep the resources from becoming the dependency of other projects.

Notably, the equinix/Provider can be exported since its Go signature remains consistent and doesn't break compatibility. This package has been a dependency for Crossplane and was previously relied upon by Pulumi before Pulumi transitioned to a git submodule pattern.

displague:
Let's isolate some of these changes and introduce them one at a time in separate PRs:

internal/config, internal/errors etc (in #142 there was some suggestion in the packages proposed on how to reorganize these helpers)

Then we can move the resources around - again separate PRs, first with Metal, then NE, then Fabric.

From there it may be easy to convert a single resource to Framework and continue from there.

ctreatma:
@t0mk when you have a moment, could you rebase this PR on the latest from `main` so we can get a better idea of what remains for unblocking framework adoption?

internal/utils.go
```diff
@@ -0,0 +1,28 @@
+package internal
```
ctreatma, line 1: It looks like the functions in this file aren't used by anything related to the Metal SSH key refactor; we should still break this change out into a separate refactor, but if it's not necessary for the Metal SSH key refactor, we can remove these changes from the PR now with no impact.

internal/mutexkv.go
```diff
@@ -1,4 +1,4 @@
-package equinix
+package internal
```
ctreatma, line 1: This move also seems to not impact the SSH key refactor and could be removed from this PR without waiting for a separate mutexkv refactor PR to be merged.

internal/port_helpers.go
```diff
@@ -1,4 +1,4 @@
-package equinix
+package internal
```
ctreatma, line 1: This one also seems to not be needed to support the SSH key changes

internal/config/config.go
```diff
@@ -28,11 +28,6 @@ import (
 	xoauth2 "golang.org/x/oauth2"
 )
 
-var (
-	UuidRE         = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
```
ctreatma, line 32: This, too, looks like it's not technically required in order to move the Metal SSH key code.

ctreatma:
Closing again since the resources affected here were covered in #479 and the unrelated refactorings will be covered in separate PRs.

PR #421, opened 2023-10-18
chore(ci): execute acctest only if pull request is not draft

ocobles:
in response to this https://github.com/equinix/terraform-provider-equinix/pull/406#issuecomment-1757739775


ctreatma (approved):
The updated workflow ran as expected via manual workflow dispatch: https://github.com/equinix/terraform-provider-equinix/actions/runs/6616762756

@ocobles reports that the other trigger configuration works on the fork, and it looks correct, so this is good to merge.

PR #426, opened 2023-10-25
feat: Adding Port2IBM2.0 example and test scripts

srushti-patl:
Port 2 IBM2.0 Connection example Scripts
Added test script for FCR Multi Cloud Connection


thogarty (changes requested):
I'm curious if naming the directory for the example ibm2.0 is going to cause issues somewhere; potentially in the modules. I don't think all tools will have an issue with it but I also don't know which tools might. Other than that there's just one small change requested and the rest looked good.

equinix/resource_fabric_connection.go
```diff
@@ -231,13 +233,15 @@ func resourceFabricConnectionUpdate(ctx context.Context, d *schema.ResourceData,
...
+		conn, err := waitFunction(d.Id(), meta, ctx)
 
 		if err != nil {
 			if !strings.Contains(err.Error(), "500") {
 				d.SetId("")
 			}
 			diags = append(diags, diag.Diagnostic{Severity: 2, Summary: fmt.Sprintf("connection property update completion timeout error: %v [update payload: %v] (other updates will be successful if the payload is not shown)", err, update)})
+		} else {
+			updatedConn = conn
 		}
```
ocobles, line 245: doing it this way, in case of error it will not save the last status

thogarty: Yes, this is intentional; some of the updates might fail but we want to save the status of the last successful update to the connection. We're not worried about the status of the failed update. The response won't have any details we're concerned about saving to Terraform state.

thogarty (approved):
LGTM!

PR #435, opened 2023-10-31
fix: Fixing VD to Azure connection resource

srushti-patl:
Fixed resource_fabric_connection by adding Type and Interface parameters to access_point. 
Updated equinix_fabric_connection.md file

### DIFF
--- a/docs/data-sources/equinix_fabric_connection.md
+++ b/docs/data-sources/equinix_fabric_connection.md
@@ -261,7 +261,6 @@ Read-Only:
 - `key` (String)
 - `value` (String)
 
-~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
 
 <a id="nestedatt--change_log"></a>
 ### Nested Schema for `change_log`
--- a/docs/data-sources/equinix_fabric_connection.md
+++ b/docs/data-sources/equinix_fabric_connection.md
@@ -261,6 +261,7 @@ Read-Only:
 - `key` (String)
 - `value` (String)
 
+~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
 
 <a id="nestedatt--change_log"></a>
 ### Nested Schema for `change_log`
--- a/docs/resources/equinix_ecx_l2_connection.md
+++ b/docs/resources/equinix_ecx_l2_connection.md
@@ -61,7 +61,7 @@ resource "equinix_ecx_l2_connection" "ports-2-azure" {
   vlan_ctag         = 2512
   seller_metro_code = "SV"
   named_tag         = "PRIVATE"
-  authorization_key = "c4dff8e8-b52f-4b34-b0d4-c4588f7338f3
+  authorization_key = "c4dff8e8-b52f-4b34-b0d4-c4588f7338f3"
   secondary_connection {
     name      = "tf-azure-sec"
     port_uuid = data.equinix_ecx_port.sv-qinq-sec.id
@@ -184,8 +184,11 @@ for more details.
 docs.
 
 * `additional_info` - (Optional) one or more additional information key-value objects
-  * `name` - (Required) additional information key
+  * `key` - (Required) additional information key
   * `value` - (Required) additional information value
+
+~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
+
 * `zside_port_uuid` - (Optional) Unique identifier of the port on the remote/destination side
 (z-side). Allows you to connect between your own ports or virtual devices across your company's
 Equinix Fabric deployment, with no need for a private service profile.
--- a/docs/resources/equinix_ecx_l2_connection.md
+++ b/docs/resources/equinix_ecx_l2_connection.md
@@ -184,11 +184,8 @@ for more details.
 docs.
 
 * `additional_info` - (Optional) one or more additional information key-value objects
-  * `key` - (Required) additional information key
+  * `name` - (Required) additional information key
   * `value` - (Required) additional information value
-
-~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
-
 * `zside_port_uuid` - (Optional) Unique identifier of the port on the remote/destination side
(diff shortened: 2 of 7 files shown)


docs/resources/equinix_fabric_connection.md
```diff
@@ -103,6 +104,12 @@ Read-Only:
 
 - `id` (String) id
 
+<a id="nestedblock--a_side--access_point--netwrok"></a>
```
thogarty, line 107: Typo in network here.

docs/resources/equinix_fabric_connection.md
```diff
@@ -192,10 +199,6 @@ Optional:
 - `type` (String) Virtual Device type
 - `uuid` (String) Equinix-assigned Virtual Device identifier
 
-Read-Only:
-
-- `href` (String) Unique Resource Identifier
```
thogarty, line 197: Is href not supported for VD schema? I see in the response schema that it will be returned. Might want to check into this a bit deeper.

equinix/fabric_mapping_helper.go
```diff
@@ -252,6 +264,27 @@ func locationToFabric(locationList []interface{}) v4.SimplifiedLocation {
 	return sl
 }
 
+func virtualdeviceToFabric(virtualdeviceList []interface{}) v4.VirtualDevice {
```
thogarty, line 267: I see that this method isn't parsing all the different parts of what could be sent in the virtual device according to our API spec. Just type and uuid.

equinix/fabric_mapping_helper.go
```diff
@@ -252,6 +264,27 @@ func locationToFabric(locationList []interface{}) v4.SimplifiedLocation {
...
+	for _, ll := range virtualdeviceList {
+		llMap := ll.(map[string]interface{})
+		tp := llMap["type"].(string)
+		ud := llMap["uuid"].(string)
+		vd = v4.VirtualDevice{Type_: tp, Uuid: ud}
+	}
+	return vd
+}
+
+func interfaceToFabric(interfaceList []interface{}) v4.ModelInterface {
```
thogarty, line 278: This isn't parsing all of what is in the schema for interface. You changed the type of id from String to Int but it's not being parsed here.

examples/fabric/v4/portConnectivity/ibm/ibm2/README.md
```diff
@@ -30,6 +30,12 @@ At minimum, you must set below variables in `terraform.tfvars` file:
 `seller_asn` - Seller ASN Number
 `seller_region` - Seller Region
 
+## Note
```
thogarty, line 33: Nice note here. Would you add this to the published docs as well? Just like where we will start putting the examples to customers that will be shown on the Hashicorp Documentation.

equinix/fabric_mapping_helper.go
```diff
@@ -252,6 +264,27 @@ func locationToFabric(locationList []interface{}) v4.SimplifiedLocation {
 	return sl
 }
 
+func virtualdeviceToFabric(virtualdeviceList []interface{}) v4.VirtualDevice {
+	vd := v4.VirtualDevice{}
+	for _, ll := range virtualdeviceList {
+		llMap := ll.(map[string]interface{})
+		tp := llMap["type"].(string)
```
thogarty, line 271: `name` was added to the VD schema but it isn't being parsed into the model here.

thogarty (changes requested):
Have some requested changes.

equinix/fabric_mapping_helper.go
srushti-patl: I have parsed only parameters which are required for API call for the virtual device and interface model.

docs/resources/equinix_fabric_connection.md
srushti-patl: It's supported for the schema, but it wasn't used in model so I removed it. But I will add it.

equinix/fabric_mapping_helper.go
thogarty: We have the other values in the schema and in the published documentation. Anything we have exposed as usable needs to be included.

docs/resources/equinix_fabric_connection.md
thogarty: Same goes for here; we just need the provider to be very consistent with what is being exposed in its schema and in the documentation. It doesn't always have to match the 4.10 spec exactly, but the Terraform Docs/Schema/Internal Parsing and sending in API requests needs to be consistent.

docs/data-sources/equinix_fabric_connection.md
```diff
@@ -261,6 +261,7 @@ Read-Only:
 - `key` (String)
 - `value` (String)
 
+~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
```
thogarty, line 264: This is specifically just for IBM on this one. I know it's a nit, but it should specify IBM side unless we know it applies to all providers.

docs/resources/equinix_ecx_l2_connection.md
```diff
@@ -61,7 +61,7 @@ resource "equinix_ecx_l2_connection" "ports-2-azure" {
   vlan_ctag         = 2512
   seller_metro_code = "SV"
   named_tag         = "PRIVATE"
-  authorization_key = "c4dff8e8-b52f-4b34-b0d4-c4588f7338f3
+  authorization_key = "c4dff8e8-b52f-4b34-b0d4-c4588f7338f3"
```
thogarty, line 64: Good cleanup! :)

docs/resources/equinix_ecx_l2_connection.md
```diff
@@ -184,8 +184,11 @@ for more details.
 docs.
 
 * `additional_info` - (Optional) one or more additional information key-value objects
-  * `name` - (Required) additional information key
+  * `key` - (Required) additional information key
```
thogarty, line 187: Does this apply for ecx_l2 though? Looking at the schema in it this won't apply to the Fabric v3 APIs.

docs/resources/equinix_ecx_l2_connection.md
```diff
@@ -184,8 +184,11 @@ for more details.
...
   * `value` - (Required) additional information value
+
+~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
```
thogarty, line 190: Again, need to make sure this applies to Fabric v3 API. The schema also suggests name/value instead of key/value as well.

Love the initiative to clean up but we need to check it against the correct backend schema :)

docs/resources/equinix_fabric_connection.md
```diff
@@ -220,6 +229,7 @@ Optional:
 - `key` (String) Additional information key
 - `value` (String) Additional information value
 
+~> **NOTE:** Connection could be modified from Service Provider Side by using parameters passed to additional_info field:  `{"key": "ASN", "value": "1111"}` `{"key": "Global", "value": "false"}` `{"key": "BGP_IBM_CIDR", "value": "172.16.0.18/30"}` `{"key": "BGP_CER_CIDR", "value": "172.16.0.19/30"}`
```
thogarty, line 232: Same note about IBM Provider needing to be specified.

docs/data-sources/equinix_fabric_connection.md
thogarty: This is also a data source. It isn't making any changes to the underlying connection and is read only. Don't think this information belongs here.

examples/fabric/v4/portConnectivity/ibm/ibm2/README.md
thogarty: The note here was still great. The movement of it to the docs excluded the information about it being the IBM side though.

thogarty (changes requested):
Still got some changes to go, but we're heading in the right direction :)

thogarty (approved):
Thanks for the edits, Srushti! LGTM

PR #436, opened 2023-11-02
feat: Add routing protocols resource to terraform provider

thogarty:
* Add routing protocols resource to provider
* Include acceptance tests and docs
* Tested manually with all permutations of direct and bgp routing protocol data
* Data Source tested manually with both direct and bgp


.gitignore
```diff
@@ -6,13 +6,11 @@
 # Terraform specific files
 terraform.tfplan
 terraform.tfstate
-*.terraform.lock.hcl
-./*.tfstate
```
displague, line 10: `terraform.tfstate` is the default name. the added `*.terraform*` rule would not capture this

docs/data-sources/equinix_fabric_routing_protocols.md
```diff
@@ -0,0 +1,205 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "equinix_fabric_routing_protocols Resource - terraform-provider-equinix"
+subcategory: "Fabric"
+description: |-
+Fabric V4 API compatible data resource that allows user to fetch routing protocols for a given routing protocol ids
+~> Note Equinix Fabric v4 resources and datasources are currently in Beta. The interfaces related to equinix_fabric_ resources and datasources may change ahead of general availability. Please, do not hesitate to report any problems that you experience by opening a new issue https://github.com/equinix/terraform-provider-equinix/issues/new?template=bug.md
```
displague, line 7:
```suggestion
~> Note Equinix Fabric v4 resources and datasources are currently in Beta. The interfaces related to equinix_fabric_ resources and datasources may change ahead of general availability. Please, do not hesitate to report any problems that you experience by opening a new issue https://github.com/equinix/terraform-provider-equinix/issues/new?assignees=&labels=bug&projects=&template=bug.yml&title=%5BBug%5D%3A+equinix_fabric_routing_protocols


```

displague: If we have other references to bug.md, we'll want to update those in a separate PR.

equinix/errors.go
```diff
@@ -98,7 +98,7 @@ func setMap(d *schema.ResourceData, m map[string]interface{}) error {
 			err = f(d, key)
 		} else {
 			if key == "router" {
-				d.Set("gateway", v)
+				err = d.Set("gateway", v)
```
displague, line 101: This condition shouldn't be here. Whatever is calling setMap with the expectation that `router` will be renamed to `gateway` should just use a `gateway` key instead of `gateway`.

This condition has the potential to affect any resource with a `router` field.

displague: Since this is followed by an additional `d.Set`, I'll assume the intent was to duplicate a value into both field names when only one field name was given. In that case, call setMap with both field names and the same value.

equinix/resource_fabric_routing_protocols_acc_test.go
```diff
@@ -0,0 +1,105 @@
...
+		CheckDestroy: checkRoutingProtocolsDelete,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFabricCreateRoutingProtocolsConfig(7, 3),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckTypeSetElemNestedAttrs("equinix_fabric_routing_protocols.test", "direct_routing_protocol.direct_ipv4.*", map[string]string{
+						"equinix_iface_ip": fmt.Sprintf("190.1.1.1/29"),
+					}),
+				),
+				ExpectNonEmptyPlan: true,
```
displague, line 26: It would be helpful to test imports to ensure we aren't drifting in state by executing a refresh using an import of the identifier.

```suggestion
				ExpectNonEmptyPlan: true,
				ImportState:             true,
				ImportStateVerify:       true,
				
```

displague: Assuming we include Import in the resource schema, which we should be doing for every resource.

thogarty: Agree, change made.

equinix/errors.go
thogarty: You read the intention correctly. When I originally made this change I was brand new to the provider and saw this as the fix but it didn't actually achieve the result I was looking for. Calling setMap with the additional field is the correct method and is updated in the fabricMappingHelper after this comment.

docs/data-sources/equinix_fabric_routing_protocols.md
thogarty: Agree; backlog item created.

.gitignore
thogarty: There's a line for `terraform.tfstate` on line 8. Is it possible that the state file could have a different name? 

My thought was that there was already a rule for it so the redundant rule was unnecessary.

docs/resources/equinix_fabric_routing_protocols.md
```diff
@@ -0,0 +1,230 @@
...
+
+Create Direct and BGP Routing Protocol on Some Connection UUID:
+```hcl
+provider "equinix" {
+  client_id     = "<equinix_client_id>"
+  client_secret = "<equinix_client_secret>"
+}
+
+resource "equinix_fabric_routing_protocols" "routing_protocols" {
+  connection_uuid = "b0eb3892-6404-442b-85c0-4f8fcf53d123"
```
displague, line 30: I have some questions about this resource that will be easier to answer in terms of this documentation example.

Will a single `equinix_fabric_routing_protocols` explicitly and comprehensively cover all of direct and BGP routing protocols for a specific connection?

If so:
- Is it not sufficient to use the connection id as the `SetId` for this TF resource? Can't we fetch and sync all routes from connection id alone?
- Will any routes not defined in this block be deleted?
- should we provide a warning that only a single routing_protocol resource per connection uuid should be active (count != 0) at a time
- Multiple direct and BGP routes seem to be supported (can we show that in the example)
- Is there a universal documented hard API upper bound on the number of routes? Should this be defined as MaxItems?


If not:
- I wonder if the plural resource name implies otherwise (that it should be a comprehensive resource)
- I assume we are doing so to avoid touching routes not defined in TF?
  - we could surface those routes as smaller resources (one type or setting to differentiate BGP vs direct)
  - we wouldn't be able to use the bulk API endpoint, so perhaps that is the reason we took this path?

In either case:
- How do we use SetId and ImportId in a way that reflects that there may be multiple Direct or BGP sub-resources?

I think it would be helpful to include tests steps (on the existing tests) that:
- add direct and bgp rules from a set
- remove direct and bgp rules from a set
- juggle (move) direct and BGP rules to another set in the same resource (this is what I'm most interested in)
- ^ include an import test on this step (I suspect the additional rules would be dropped today)

thogarty: @displague, only one of Direct and one of BGP are allowed on a connection resource at a time:

https://github.com/equinix-product/interconnection_ic-routing-protocol-management/blob/db8f14a862f7cf9f9deb6283e0d9bc11f119e061/src/main/java/com/equinix/interconnection/fabric/routing/handler/RoutingProtocolHandler.java#L159-L168

This resource is to manage the Single Direct and BGP routing protocol resources under a connection. There will not be more than a single of each. I should add MaxItems = 1 to those set types to reflect that.

This resource is to manage the synchronization between Direct and BGP. Direct cannot be deleted before BGP if both are applied to the connection; an error will be returned if attempted. Additionally, you cannot send a POST for Direct or BGP after they have been created to overwrite; it will return an error saying only one of each type is allowed.

This restriction I believe simplifies your queries. Let me know if that's not the case though.

displague:
Noting the relevance to #362 #571. 

I thought this may have already been merged by other PRs before realizing this is for the plural resource and datasource. The singular variations are present today.

thogarty:
Out of date. Closing. Open issue https://github.com/equinix/terraform-provider-equinix/issues/484 refers to it and if that conversation ever gets going we can revisit this.

PR #449, opened 2023-11-03
fix: Add network body response in terraform errors for equinix_fabric_connection resource

thogarty:
* Add errors method to output err and network response body together
* Update equinix_fabric_connection resource to include the errors method in response to errors on network requests

Part of https://github.com/equinix/terraform-provider-equinix/issues/124

Before:
```
equinix_fabric_connection.aws-qinq: Creating...
╷
│ Error: 400 Bad Request
│ 
│   with equinix_fabric_connection.aws-qinq,
│   on main.tf line 20, in resource "equinix_fabric_connection" "aws-qinq":
│   20: resource  "equinix_fabric_connection" "aws-qinq"{
```

After:
```
equinix_fabric_connection.aws-qinq: Creating...
╷
│ Error: 400 Bad Request; Code: EQ-3142501; Message: Invalid argument value passed; Details: Invalid notification emails; AdditionalInfo: [{Property: /notifications/emails, Reason: Not Provided}]
│ 
│   with equinix_fabric_connection.aws-qinq,
│   on main.tf line 20, in resource "equinix_fabric_connection" "aws-qinq":
│   20: resource "equinix_fabric_connection" "aws-qinq" {
│ 
```

### DIFF
--- a/equinix/resource_fabric_connection.go
+++ b/equinix/resource_fabric_connection.go
@@ -114,9 +114,9 @@ func resourceFabricConnectionCreate(ctx context.Context, d *schema.ResourceData,
 		Project:        &project,
 	}
 
-	conn, _, err := client.ConnectionsApi.CreateConnection(ctx, createRequest)
+	conn, httpResponse, err := client.ConnectionsApi.CreateConnection(ctx, createRequest)
 	if err != nil {
-		return diag.FromErr(err)
+		return networkErrorOutput(err, httpResponse)
 	}
 	d.SetId(conn.Uuid)
 
@@ -134,9 +134,9 @@ func resourceFabricConnectionCreate(ctx context.Context, d *schema.ResourceData,
 			},
 		}
 
-		_, _, patchErr := client.ConnectionsApi.UpdateConnectionByUuid(ctx, patchChangeOperation, conn.Uuid)
+		_, patchHttpResponse, patchErr := client.ConnectionsApi.UpdateConnectionByUuid(ctx, patchChangeOperation, conn.Uuid)
 		if patchErr != nil {
-			return diag.FromErr(err)
+			return networkErrorOutput(err, patchHttpResponse)
 		}
 
 		if _, statusChangeErr := waitForConnectionProviderStatusChange(d.Id(), meta, ctx); err != nil {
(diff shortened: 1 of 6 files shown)


equinix/resource_fabric_connection.go
```diff
@@ -218,9 +214,9 @@ func resourceFabricConnectionUpdate(ctx context.Context, d *schema.ResourceData,
 	updatedConn := dbConn
 
 	for _, update := range updateRequests {
-		_, _, err := client.ConnectionsApi.UpdateConnectionByUuid(ctx, update, d.Id())
+		_, httpResponse, err := client.ConnectionsApi.UpdateConnectionByUuid(ctx, update, d.Id())
 		if err != nil {
-			diags = append(diags, diag.Diagnostic{Severity: 2, Summary: fmt.Sprintf("connectionn property update request error: %v [update payload: %v] (other updates will be successful if the payload is not shown)", err, update)})
+			diags = append(diags, networkErrorOutput(fmt.Errorf("connectionn property update request error: %v [update payload: %v] (other updates will be successful if the payload is not shown)", err, update), httpResponse)...)
```
displague, line 219:
```suggestion
			diags = append(diags, networkErrorOutput(fmt.Errorf("connection property update request error: %v [update payload: %v] (other updates will be successful if the payload is not shown)", err, update), httpResponse)...)
```

equinix/errors.go
```diff
@@ -51,6 +53,17 @@ func convertToFriendlyError(errors Errors, resp *http.Response) error {
...
 }
 
+func networkErrorOutput(err error, resp *http.Response) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	responseBody, readErr := io.ReadAll(resp.Body)
+	resp.Body.Close()
+	if readErr != nil {
+		diags = append(diags, diag.Diagnostic{Severity: 2, Summary: fmt.Sprintf("error parsing network request body: %s", readErr)})
+	}
+	diags = append(diags, diag.Diagnostic{Severity: 2, Summary: fmt.Sprintf("%s; \n %s", err, responseBody)})
```
displague, line 63: Can the responseBody be parsed into an Equinix API error object, and can the messages from that be included in the error message, rather than a raw json error? (or is that happening elsewhere)

thogarty: Are you talking about this type?: 
```
type ErrorResponse struct {
	StatusCode int
	Errors
	IsAPIError bool
}
```

The error response from Fabric is pretty different:
```
[{
    "errorCode":"EQ-3142601",
    "errorMessage":"Error while validating Authorization Key",
    "details":"Authorization key provided is not valid.",
    "correlationId":"pd6cVulZP2&JMxPsfDtIoiG#3",
    "additionalInfo":[
        {"property":"/zSide/accessPoint/authenticationKey"}
    ]
}]
```

For some reason I'm not quite seeing the connection at this point, but would be happy to attempt a similar pattern once it clicks. I don't think I've seen what the Terraform Error Output difference would be, but I agree that an improvement to it would be nice so that it isn't just a string of white text underneath the error.

srushti-patl (approved):
LGTM

equinix/errors.go
thogarty: @displague I've updated the PR to align with your comments here. We don't have OpenAPIGenerator yet for Fabric-Go so we can't modify the SDK to include the formatMessaging there to do the same exact process as Metal errors. It's doing the message formatting after doing type assertion on the SDK's returned to get the details from Fabric-Go's ModelError type.

When @ctreatma is ready for other services to be added to https://github.com/equinix/equinix-sdk-go we will add Fabric to that with the OpenAPIGenerator and leave the old project on SwaggerCodegen. Then we will gradually migrate our resources to the grouped SDK during a refactor effort.

Thank you for the great feedback on getting this more aligned overall. We're still working towards convergence of development patterns but it's moving in that direction 👍

PR #451, opened 2023-11-06
feat: Adding Port2IBM1.0 example and test script

srushti-patl:
Added Port 2 IBM1.0 connection example and test script


thogarty (approved):
LGTM! Thanks Srushti

PR #452, opened 2023-11-06
chore: gh action to upgrade provider dependency in Pulumi provider

ocobles:
This action will open an Issue in equinix/pulumi-equinix when there is a new release, and this in turn will trigger https://github.com/equinix/pulumi-equinix/blob/main/.github/workflows/upgrade-provider.yml


displague (approved):
lgtm

As a next step, could we trigger a PR on Pulumi to build the update?

ocobles:
> lgtm
> 
> As a next step, could we trigger a PR on Pulumi to build the update?

Yes @displague, It is already in place, I took it from Pulumi, it is the method that is used for all the providers that they maintain. https://github.com/equinix/pulumi-equinix/blob/c73136ee0a1a8f9c4e3ccb6c9edd834aa2f6f967/.github/workflows/upgrade-provider.yml#L6-L8

PR #459, opened 2023-11-14
refactor: Move Config-related code to internal/config

t0mk:
This PR moves Config struct and related code to `internal/config`. It's part of breakdown of the refactor to `internal/`.

Part of #106


equinix/fabric_mapping_helper.go
```diff
@@ -1228,21 +1226,3 @@ func getCloudRouterUpdateRequest(conn v4.CloudRouter, d *schema.ResourceData) (v
 	}
 	return changeOps, nil
 }
-
```
t0mk, line 1231: This code is needed in Config for some user-agent magic, so I moved it to `internal/fabric_utils`

displague:
Could you run `goimports -w -l github.com/equinix/terraform-provider-equinix .` on this branch. 

The TF project imports should be in the group of non-native package or a third group dedicated to local packages (which is what the above command will do).

t0mk:
> Could you run `goimports -w -l github.com/equinix/terraform-provider-equinix .` on this branch.

Done!

displague:
Could we move `internal` to the root. While it has the same effect to use `equinix/internal`, in that we can freely change the go package signatures without worry of breaking third-party dependencies, the intent with this `internal/` package will be to park everything by default. This includes Equinix related code and generic (type juggling) or TF setup/helping code.

internal/fabric_utils/correlation_id.go
```diff
@@ -0,0 +1,24 @@
+package fabric_utils
```
displague, line 1: the package name should avoid `utils`

Without looking at how this is used.. fabric/correlation ?

If this is only used in one or two files, especially for testing, I'd be tempted to make this a private helper adjacent to the code that needs it.

t0mk: There will be more fabric utils when we migrate further:

```
fabric_mapping_helper.go
fabric_mapping_service_profile_helper.go
fabric_port_mapping_helper.go
```

The CorrelationId code was in fabric_mapping_helper.go IIRC. I didn't think about it too much and moved it to fabric_utils. However it seems that CorrelationId is not used anywhere but in Config. So I moved it straight to Config.

ocobles: You can move that CorrelationId code into Config but I would create a `helper` folder under `internal`, you can add those mapping helpers you mention there and we will also need that also to avoid import cycles https://github.com/equinix/terraform-provider-equinix/pull/406#issuecomment-1764477274 so it should look like 
image

displague: `helper` (like `utils`) is too generic of a name

- https://go.dev/doc/effective_go#names
- https://go.dev/blog/package-names#bad-package-names

displague:
There were an unusually high number of E2E tests that failed in CI for this build, but none look to be related to this change

displague (approved):

PR #465, opened 2023-11-16
fix: make correrlationI() function private

t0mk:
CorrelationId was moved to internal/config and accidental made public. This PR makes it private.


ocobles (approved):
lgtm. Merging since this change doesn't affect e2e running test

PR #467, opened 2023-11-27
refactor: move error handling code to internal/errs, move SetMap to internal/utils

t0mk:
This PR is moving error code to internal/errs, and SetMap function to internal/utils.

I'm open to change the directory/package names.


internal/utils/utils.go
```diff
@@ -0,0 +1,36 @@
+package utils
```
ctreatma, line 1: Since this is a helper for working with schemas, I think we could have `package schema` and this file could be `set_map.go`

internal/utils/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type setFn = func(d *schema.ResourceData, key string) error
+
+// setMap sets the map of values to ResourceData, checking and returning the
+// errors. Typically d.Set is not error checked. This helper makes checking
+// those errors less tedious. Because this works with a map, the order of the
+// errors would not be predictable, to avoid this the errors will be sorted.
+func SetMap(d *schema.ResourceData, m map[string]interface{}) error {
```
ctreatma, line 16: Not in scope for this PR, but looking at how and where this helper is used, I'm not convinced it's really necessary.

displague: The pattern SetMap was made to simplify was 

```
d.Set("foo", foo)
d.Set("bar", bar)
```

Which is repetitive and common, but is also bad practice.

`d.Set` can error out and those errors were seldom checked. When they are checked, you end up with resources with 10-30 lines per attributes and an extra 20-60 lines of `if {`, `}`.

internal/schema/set_map.go
```diff
@@ -0,0 +1,36 @@
+package utils
```
displague, line 1: nit: we can correct this later as minor cleanup. 
the package name should match -- schema

underscores in filenames can be interpreted as compiler hints, as in `_arm64.go`. I try to avoid them. 
In this case, maps.go could be an alternative.

PR #472, opened 2023-12-02
fix: fixing schemas for Servicetoken and accesspoint.interface.

srushti-patl:
This PR includes fixes for two defects(CXF-[81949](https://equinixjira.atlassian.net/browse/CXF-81949) & [CXF-81844)](https://equinixjira.atlassian.net/browse/CXF-81844)
- fixed service token schema
- fixed accespoint.interface schema

### DIFF
--- a/equinix/resource_fabric_connection.go
+++ b/equinix/resource_fabric_connection.go
@@ -49,8 +49,8 @@ func resourceFabricConnectionCreate(ctx context.Context, d *schema.ResourceData,
 	aside := d.Get("a_side").(*schema.Set).List()
 	projectReq := d.Get("project").(*schema.Set).List()
 	project := projectToFabric(projectReq)
-	additionalInfo := d.Get("additional_info").([]interface{})
-	additionalinfoGoToTerraState := additionalInfoToFabric(additionalInfo)
+	additionalInfoTerraConfig := d.Get("additional_info").([]interface{})
+	additionalInfo := additionalInfoTerraToGo(additionalInfoTerraConfig)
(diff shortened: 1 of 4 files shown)


equinix/resource_fabric_connection.go
```diff
@@ -50,7 +50,7 @@ func resourceFabricConnectionCreate(ctx context.Context, d *schema.ResourceData,
 	projectReq := d.Get("project").(*schema.Set).List()
 	project := projectToFabric(projectReq)
 	additionalInfo := d.Get("additional_info").([]interface{})
-	additionalinfo := additionalInfoToFabric(additionalInfo)
+	additionalinfoGoToTerraState := additionalInfoToFabric(additionalInfo)
```
thogarty, line 53: This is actually `additionalInfoTerraConfigToGo`. The i should be capitalized in info and the data flow is Terraform Config -> Go Struct.

However, the variable shouldn't be the one naming the operation, the method should be. Everywhere we have `ToFabric` should become `TerraToGo` in the methods themselves. 

I would name the variable on line 52 `additionalInfoTerraConfig` and make the variable on line 53 be `additionalInfo`.

If you want to change the method name with this PR you can but not necessary.

thogarty (approved):
Just the one requested change. Looks good aside from that.

equinix/resource_fabric_connection.go
srushti-patl: I have changed both the variable names and the method name.

thogarty (approved):
LGTM. Ship it.

PR #473, opened 2023-12-04
refactor: move data conversion code to internal/converters

t0mk:
This PR is part of refactoring effort. It moves some helper code to another internal subpackage.


internal/converters/hashcode.go
```diff
@@ -11,7 +11,7 @@ import (
 // and invert it if the result is negative.
 //
 // Originally from https://github.com/hashicorp/terraform-plugin-sdk/blob/main/internal/helper/hashcode/hashcode.go
-func hashcodeString(s string) int {
+func HashcodeString(s string) int {
```
ctreatma, line 14: We could move this to a `hash` or `hashcode` package and rename to `String`

internal/converters/converters.go
```diff
@@ -1,11 +1,11 @@
-package equinix
+package converters
 
 import (
 	"strconv"
 	"strings"
 )
 
-func contains(s []string, e string) bool {
+func Contains(s []string, e string) bool {
```
ctreatma, line 8: If we build on go 1.21 instead of 1.20 we could replace this with usage of the `slices` standard package (even if we don't upgrade, we could use `x/exp/slices`.

ctreatma: Replacement of this function could be saved for a later change, but if we opt to keep this implementation of `Contains`, I think it would fit better in a package named `slices`, since it isn't converting anything.

internal/converters/converters.go
```diff
@@ -14,15 +14,15 @@ func contains(s []string, e string) bool {
...
+func StringArrToIfArr(sli []string) []interface{} {
 	var arr []interface{}
 	for _, v := range sli {
 		arr = append(arr, v)
 	}
 	return arr
 }
 
-func convertStringArr(ifaceArr []interface{}) []string {
+func ConvertStringArr(ifaceArr []interface{}) []string {
```
ctreatma, line 25: Let's take this opportunity to improve these function names.  `StringArrToIfArr` is a nice, descriptive name, and if we follow that model here, this function would be called `IfArrtoStringArr`

internal/converters/converters.go
```diff
@@ -33,7 +33,7 @@ func convertStringArr(ifaceArr []interface{}) []string {
 	return arr
 }
 
-func convertIntArr(ifaceArr []interface{}) []string {
+func ConvertIntArr(ifaceArr []interface{}) []string {
```
ctreatma, line 36: This would be `IfArrToIntStringArr` (or can it be safely replaced with usage of other converters?)

internal/converters/converters.go
```diff
@@ -44,7 +44,7 @@ func convertIntArr(ifaceArr []interface{}) []string {
 	return arr
 }
 
-func convertIntArr2(ifaceArr []interface{}) []int {
+func ConvertIntArr2(ifaceArr []interface{}) []int {
```
ctreatma, line 47: This would be `IfArrToIntArr`

t0mk:
@ctreatma I fixed your review comments.

t0mk:
@ctreatma I remove hashcode.go from `equinix/` and fixed all the places where it was used.

ctreatma:
@t0mk could you push up that change?  GitHub still shows the old commit with both hashcode copies.

t0mk:
@ctreatma , sorry, I pushed it now. It should be ready for review again.

PR #479, opened 2023-12-07
refactor: First draft of changes to separate 2 basic resources away from equinix/

t0mk:
This PR is an attempt to move two basic resources away from the equinix namespace. Once this is in, it should be easier to start working on the framework plugin migration.


t0mk:
@ctreatma @displague What do you think?

internal/resources/metal/metal_ssh_key/resource.go
```diff
@@ -52,21 +52,21 @@ func metalSSHKeyCommonFields() map[string]*schema.Schema {
...
 			State: schema.ImportStatePassthrough,
 		},
 
-		Schema: metalSSHKeyCommonFields(),
+		Schema: MetalSSHKeyCommonFields(),
 	}
 }
 
-func resourceMetalSSHKeyCreate(d *schema.ResourceData, meta interface{}) error {
+func ResourceMetalSSHKeyCreate(d *schema.ResourceData, meta interface{}) error {
```
displague, line 69: nit: We shouldn't need to export the CRUD functions, just the Resource or DataSource itself.

displague: And since we are sharing CRUD and Schema between the project ssh key and user ssh keys, we could still use the exported Resource as the base and then give it a Project variation of Schema.  The shared schema (`MetalSSHKeyCommonFields`) could be exported.

internal/resources/metal/metal_ssh_key/resource.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/packethost/packngo"
 )
 
-func metalSSHKeyCommonFields() map[string]*schema.Schema {
+func MetalSSHKeyCommonFields() map[string]*schema.Schema {
```
displague, line 16: We should avoid repeating the package name in the function names. For example, this would be `metal_ssh_key.CommonFields()`.

internal/acceptance/ssh_key_helpers.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+
+	"github.com/equinix/terraform-provider-equinix/internal/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/packethost/packngo"
+)
+
+func TestAccCheckMetalSSHKeyExists(n string, key *packngo.SSHKey) resource.TestCheckFunc {
```
displague, line 12: I suspect this file should be named `..._test.go`

GNUmakefile
```diff
@@ -23,7 +23,14 @@ test:
 	echo $(TEST) | \
 		xargs -t ${GOCMD} test -v $(TESTARGS) -timeout=10m
 
-testacc:
+testaccmoved:
+	# go through all the directories in internal/resources and run tests
+	for d in $$(find ./internal/resources -maxdepth 2 -mindepth 2 -type d); do \
```
displague, line 28: These packages should already be covered by `$(TEST)`. The for loop over directories in internal/ doesn't seem necessary and I suspect we are retesting packages.

internal/config/envvars.go
```diff
@@ -0,0 +1,11 @@
+package config
+
+
+const (
```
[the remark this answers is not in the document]
t0mk: Why not have the envvars defined in a separate file? They are used by provider and sweeper code; and are obviously unreleated to code in config.go. I would consider a separate package (`internal/envvars`) but since it's configuration related, I dropped it here.

The envvars were moved to `internal` because sweeping in the acceptance tests for the ssh_key uses them, and can't use them from `equinix/` because of import cycyle. If I wouldn't actually try to make acceptance tests of moved metal_ssh_key work, I wouldn't know that they need to be moved. I just spend 15 minutes researching on how to justify this :(.

internal/acceptance/ssh_key_helpers.go
t0mk: This is not a test, just a helper function, and we've got one like this for most of the resource types. I'm not sure why.

I would name the file `resource_checkers.go` and put there all the other test helpers which check if resource exists remotely.

internal/config/envvars.go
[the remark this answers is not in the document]
ctreatma: If they're going to the `config` package (which is IMO reasonable since they are used as configuration), then there's not really a compelling reason to add them to a separate file that only defines the constants: it's still in the `config` package and it still gets imported along with all of the code in `config.go`.  Moving the constants to a different file in the same package shouldn't lead to an import cycle.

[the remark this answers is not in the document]
t0mk: OK, I moved the const block to config.go and removed envvars.go.

internal/acceptance/device_helpers.go
```diff
@@ -0,0 +1,119 @@
+package acceptance
```
displague, line 1: Not requesting any changes, just commenting:

These helpers may turn out to all be "metal" related. In the future we may want to consider moving such service-shared testing / utility code to a service specific home.

internal/services/metal/{resourcename}
internal/services/metal/testing

More likely, and for more gain, I think we would want to move towards a testing pattern like the one in the Linode provider:
https://github.com/linode/terraform-provider-linode/tree/dev/linode/instance/tmpl

internal/acceptance/ssh_key_helpers.go
t0mk: @displague What about this?

internal/resources/metal/metal_ssh_key/resource.go
t0mk: I changed the CRUD callbacks to private and renamed them. I also made the schema easier to share between metal_ssh_key and metal_project_ssh_key.

ctreatma (approved):
IMO this is good as-is; I see the relevant tests passing in CI, and the failed tests are all platform issues.  If there are things we want to adjust we can do that in follow-up PRs.

PR #480, opened 2023-12-07
refactor: clean up config package

ctreatma:
When the config struct was moved to its own package, we added a couple public variables to replace private ones.  However, the original private variable definitions were left in place, and all references to the private variables were left unchanged.

On closer inspection, one of the private variables (`ipAddressTypes`) is only used by one resource and can remain private. The other one (`uuidRE`) is used for validation that duplicates a built-in terraform validator, so I've replaced it with the built-in validator.


internal/validation/strings.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var (
+	uuidRE       = regexp.MustCompile("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$")
+	StringIsUuid = validation.StringMatch(uuidRE, "must be a valid UUID")
```
displague, line 11: The upstream [validation.IsUUID](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/validation/uuid.go) covers this need. The Parse function it relies on (from Hashicorp) works with upper, lower, and mixed case.

https://go.dev/play/p/U_z8v_5j2ip

ctreatma: Good call, I like removing things!

t0mk (approved):
It looks good. Is there some way we could detect these orphaned var? Does it show in VScode as unused?

PR #482, opened 2023-12-08
feat: Add equinix_fabric_network terraform provider

manu-equinix:
- This PR adds a new provider to manage an equinix network resource
- Usage is documented on the examples folder 
- provider name: equinix_fabric_network


equinix/fabric_mapping_helper.go
```diff
@@ -438,6 +437,25 @@ func operationToTerra(operation *v4.ConnectionOperation) *schema.Set {
 	return operationSet
 }
 
+func networkOperationToTerra(operation *v4.NetworkOperation) *schema.Set {
```
ctreatma, line 440: There is a large refactor in progress to enable us to gradually migrate the provider from the deprecated terraform-plugin-sdk to its replacement, terraform-plugin-framework.  One thing we've found during that refactor is that shared, private functions like this make it more difficult to refactor resources & data sources from the SDK to the framework.

Instead of adding new helpers to this file, I recommend creating a new package & code file under the `internal` directory.  For example, you could define this function as `NetworkOperationToTerra` in the file `internal/mapping/fabric.go`, specifying `package mapping` at the top of that file; then in files that use this helper you would need to import `github.com/equinix/terraform-provider-equinix/internal/mapping`, and you would call the function as `mapping.NetworkOperationToTerra(...)`.

examples/fabric/v4/network/README.md
```diff
@@ -0,0 +1,34 @@
+# ECX Fabric network CRUD operations
```
thogarty, line 1: We're no longer adding to the examples in the Provider repo. Add example usage with values from tfvars.example directly on the resource to the documentation.

Additionally, docs/ need to be added for the resource and data source with including example usage. You can use this doc as a reference for example usage: https://github.com/equinix/terraform-provider-equinix/blob/main/docs/resources/equinix_ecx_l2_connection.md

The addition to examples directory is not needed.

equinix/resource_fabric_network.go
```diff
@@ -0,0 +1,248 @@
...
+		Type_:         &netType,
+		Scope:         &netScope,
+		Location:      &location,
+		Notifications: notifications,
+		Project:       &project,
+	}
+
+	fabricNetwork, _, err := client.NetworksApi.CreateNetwork(ctx, createRequest)
+	if err != nil {
+		return diag.FromErr(err)
```
thogarty, line 64: All error responses from Fabric Go SDK need to be updated to follow our error output pattern. Use this for reference: https://github.com/equinix/terraform-provider-equinix/pull/449

equinix/resource_fabric_network.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		return diag.Errorf("either timed out or errored out while fetching Fabric Network for uuid %s and error %v", d.Id(), err)
+	}
+	// TO-DO
+	update, err := getNetworkUpdateRequest(dbConn, d)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	updates := []v4.NetworkChangeOperation{update}
+	_, res, err := client.NetworksApi.UpdateNetworkByUuid(ctx, updates, d.Id())
```
thogarty, line 126: `res` from Fabric_Go SDK should be left out of errors. It will only show the headers of the request and doesn't provide value to the end user. Just clutters the error output. Please follow the pattern set for error output on Fabric Resources defined here: https://github.com/equinix/terraform-provider-equinix/pull/449

thogarty (changes requested):
I haven't been able to get to the schema yet, but I've requested some changes as well as agreeing with Charles' comment.

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Fabric Network overall state",
+		},
+		"scope": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Fabric Network scope",
+		},
+		"equinix_asn": {
```
thogarty, line 85: I don't see this value `equinix_asn` in the schema for Network: https://app.swaggerhub.com/apis/equinix-api/fabric/4.11#/Networks/createNetwork

It should be removed if it's not there.

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+			Computed:    true,
+			Description: "Equinix ASN",
+		},
+		"type": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ValidateFunc: validation.StringInSlice([]string{"IPWAN", "EPLAN", "EVPLAN"}, true),
+			Description:  "Supported Network types - EVPLAN, EPLAN, IPWAN",
+		},
+		"location": {
```
thogarty, line 96: If the value can be input into the schema or returned from the API response, then it needs to have Optional and Computed set to true.

Reference for what behaviors to include is explained in detail here: https://equinixjira.atlassian.net/wiki/spaces/IC/pages/145918564397/Terraform+Schema+Standards

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "Fabric Network location",
+			MaxItems:    1,
+			Elem: &schema.Resource{
+				Schema: createLocationSch(),
+			},
+		},
+		"project": {
```
thogarty, line 105: Computed+Optional

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "Fabric Network project",
+			Elem: &schema.Resource{
+				Schema: createNetworkProjectSch(),
+			},
+		},
+		"operation": {
```
thogarty, line 113: There is no `operation` in the PostRequest body. This should just be Computed

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Description: "Network operation information that is associated with this Fabric Network",
+			MaxItems:    1,
+			Elem: &schema.Resource{
+				Schema: createNetworkOperationSch(),
+			},
+		},
+		"change": {
```
thogarty, line 122: Computed only

equinix/fabric_network_schema.go
```diff
@@ -0,0 +1,148 @@
...
+		"href": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Computed:    true,
+			Description: "Unique Resource URL",
+		},
+	}
+}
+
+func createNetworkResourceSchema() map[string]*schema.Schema {
```
thogarty, line 63: Please review the schema. Some values from the response are left out.

equinix/fabric_network_read_schema.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		"type": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "network change type",
+		},
+	}
+}
+
+func readNetworkResourceSchema() map[string]*schema.Schema {
```
thogarty, line 25: This schema is more correct for what the create should have been. Though the behaviors on create will be different. Double check that it matches and I'll review again.

thogarty (changes requested):
Schema updates required.

equinix/resource_fabric_network.go
manu-equinix: Ok, got it. I see that the main branch was updated with this code just a couple of hours ago after my PR was out. Let me check if I can rebase on latest changes.

thogarty: All good! There's no way you could have known. Just wanted to make sure that now we have it we keep up with it. :)

displague:
@thogarty Has this already been merged as #553? Anything remaining to incorporate from this PR?

thogarty:
@displague yes, this was already merged in the other PR. I'll close this one.

thogarty:
Closing because it's a redundant effort that was already cleaned up.

PR #486, opened 2023-12-12
refactor: remove diagnostics wrapper

ctreatma:
While fixing timeout support in some of our resources, we introduced a private `diagnosticsWrapper` function to act as a shim between the context-aware resource lifecycle hooks and our pre-existing, non-context-aware resource lifecycle functions.  This shim was useful for gradual migration, but now it adds an unnecessary dependency between certain resources/datasources and the code in `provider.go`.

This removes the `diagnosticsWrapper` in favor of updating the resource lifecycle functions directly so that they have the correct function signature.  This work uncovered some places where we were needlessly calling `equinix_errors.FriendlyError`, so those are cleaned up here as well.

### DIFF
--- a/equinix/data_source_metal_spot_market_request.go
+++ b/equinix/data_source_metal_spot_market_request.go
@@ -11,13 +11,14 @@ import (
 
 	"github.com/equinix/terraform-provider-equinix/internal/config"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/packethost/packngo"
 )
 
 func dataSourceMetalSpotMarketRequest() *schema.Resource {
 	return &schema.Resource{
-		ReadContext: diagnosticsWrapper(dataSourceMetalSpotMarketRequestRead),
+		ReadContext: dataSourceMetalSpotMarketRequestRead,
 
 		Schema: map[string]*schema.Schema{
 			"request_id": {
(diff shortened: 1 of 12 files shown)


ctreatma:
Note: I've structured the [commits](https://github.com/equinix/terraform-provider-equinix/pull/486/commits) of this PR so that they can be reviewed separately; if reviewers prefer, I can instead open a separate PR for each commit.

equinix/data_source_metal_spot_market_request.go
```diff
@@ -112,7 +113,7 @@ func dataSourceMetalSpotMarketRequestRead(ctx context.Context, d *schema.Resourc
 
 	d.SetId(id)
 
-	return equinix_schema.SetMap(d, map[string]interface{}{
+	err = equinix_schema.SetMap(d, map[string]interface{}{
```
displague, line 116: It's worth considering how SetMap (which depends on Hashicorp's go-multierror](https://github.com/hashicorp/go-multierror)) should be updated to work best with [diagnostics](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/diag#Diagnostics) 

I'm looking at the `Append` capabilities in https://developer.hashicorp.com/terraform/plugin/framework/diagnostics and wondering if we would adopt that in SetMap rather than using multierrors, or if this is something that we consider with Framework.

ctreatma: I'm not sure that deeper integration with the diag package is worth the tradeoff of needing 2 implementations (or needing an additional shim to make it work for both)--the SDK and framwork diag packages have different interfaces, and `Append` is only defined for the framework implementation.

IMO the goal of the framework migration is to make it as easy as possible to move an SDK resource over to the framework, so having a single multierror-based SetMap that works for both SDK and framework is preferable until we complete the framework migration, and then once we're back on a single provider implementation we can optimize freely.

displague: I should have prefaced that this was more future thinking and not a direct review / blocker.

displague (approved):

PR #492, opened 2023-12-14
feat: add VM resizing for device resource core upgrade and new device status for device data source

rling-equinix:
In Dec release, we added the "core" field to the [Update Virtual Device](https://developer.equinix.com/dev-docs/ne/api-reference/virtual-devices#update-virtual-device) API to allow users to resize their devices. After device is provisioned, the customer can send a PATCH request to upgrade its core number. The device will be marked as "RESOURCE_UPGRADE_IN_PROGRESS" and once the upgrade is done, the device status will be provisioned again.
More details could be found in our API documentation.
reference: https://developer.equinix.com/docs?page=/dev-docs/ne/overview
product reference: https://docs.equinix.com/en-us/Content/Interconnection/NE/user-guide/NE-VNF-resize.htm#UpgradeResourceSizeforClusterDeviceDeployment

### DIFF
--- a/docs/data-sources/equinix_network_device.md
+++ b/docs/data-sources/equinix_network_device.md
@@ -42,6 +42,8 @@ NOTE: Exactly one of either `uuid` or `name` must be specified.
   * FAILED
   * DEPROVISIONING
   * DEPROVISIONED
+  * RESOURCE_UPGRADE_IN_PROGRESS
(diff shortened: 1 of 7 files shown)


rling-equinix:
go.mod and go.sum needs to be updated after https://github.com/equinix/ne-go/pull/26 (go client) is merged.

ctreatma (approved):

PR #498, opened 2023-12-19
refactor: move mutexkv to its own package

ctreatma:
Moving mutexkv out of provider.go makes it easier to restructure our code.

### DIFF
--- a/internal/mutexkv/mutexkv.go
+++ b/internal/mutexkv/mutexkv.go
@@ -1,8 +1,12 @@
-package equinix
+package mutexkv
 
 import (
 	"log"
 	"sync"
+)
+
+var (
+	Metal = NewMutexKV()
 )
 
 // MutexKV is a simple key/value store for arbitrary mutexes. It can be used to
--- a/equinix/provider.go
+++ b/equinix/provider.go
@@ -22,7 +22,6 @@ import (
 )
 
 var (
-	metalMutexKV         = NewMutexKV()
 	DeviceNetworkTypes   = []string{"layer3", "hybrid", "layer2-individual", "layer2-bonded"}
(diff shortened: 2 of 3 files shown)


internal/mutexkv/mutexkv.go
```diff
@@ -1,10 +1,14 @@
-package equinix
+package mutexkv
 
 import (
 	"log"
 	"sync"
 )
 
+var (
+	Metal = NewMutexKV()
```
t0mk, line 9: since it's used only in port-vlan attachment, shall we call it PortVlanAttachmentMutex?
```suggestion
	Metal = PortVlanAttachmentMutex()
```
If I would do it, I would drop it to `config/`, it's just one variable.  The mutexkv package is not justified IMO.

ctreatma: If we keep `mutexkv` as-is, there should not be separate instances for each resource/data source that uses it, because the purpose of the `mutexkv` type is to be a shared storage space for separate types to use the same mutexes.

However: since we're only using this in the port VLAN attachment resource, that indicates that we have no need for this type, so IMO it makes sense to eliminate the `mutexkv` entirely and update port VLAN attachment to just set up and use its own mutex.  I'll do that.

ctreatma: Whoops, I forgot that the VLAN attachment resource uses multiple mutexes, one per port ID, so `mutexkv` can't be removed.  IMO it's still preferable to have a shared mutexkv instance rather than initializing a private instance in the VLAN attachment resource, though, since the type is intended to allow synchronization across arbitrary collaborators.

t0mk (approved):

PR #501, opened 2023-12-21
refactor: finish moving device resource off of packngo

ctreatma:
We previously migrated the Equinix Metal device data source and the read pathways for the resource from `packngo` to `metal-go` (and then to `equinix-sdk-go`).  This finishes the migration of the device resource by replacing `packngo` with `equinix-sdk-go` in the create, update, and delete pathways for that resource.

### DIFF
--- a/equinix/helpers_device_test.go
+++ b/equinix/helpers_device_test.go
@@ -2,40 +2,21 @@ package equinix
 
 import (
 	"context"
-	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
-	"golang.org/x/exp/slices"
-
-	"github.com/packethost/packngo"
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
+	"github.com/equinix/terraform-provider-equinix/internal/config"
 )
-
-type mockHWService struct {
-	GetFn  func(string, *packngo.GetOptions) (*packngo.HardwareReservation, *packngo.Response, error)
-	ListFn func(string, *packngo.ListOptions) ([]packngo.HardwareReservation, *packngo.Response, error)
-	MoveFn func(string, string) (*packngo.HardwareReservation, *packngo.Response, error)
-}
-
-func (m *mockHWService) Get(id string, opt *packngo.GetOptions) (*packngo.HardwareReservation, *packngo.Response, error) {
-	return m.GetFn(id, opt)
-}
-
-func (m *mockHWService) List(project string, opt *packngo.ListOptions) ([]packngo.HardwareReservation, *packngo.Response, error) {
-	return m.ListFn(project, opt)
-}
-
-func (m *mockHWService) Move(id string, dest string) (*packngo.HardwareReservation, *packngo.Response, error) {
-	return m.MoveFn(id, dest)
-}
-
-var _ packngo.HardwareReservationService = (*mockHWService)(nil)
 
 func Test_waitUntilReservationProvisionable(t *testing.T) {
 	type args struct {
 		reservationId string
 		instanceId    string
-		meta          *packngo.Client
+		handler       func(w http.ResponseWriter, r *http.Request)
 	}
 
 	tests := []struct {
@@ -48,12 +29,10 @@ func Test_waitUntilReservationProvisionable(t *testing.T) {
 			args: args{
 				reservationId: "reservationId",
 				instanceId:    "instanceId",
-				meta: &packngo.Client{
-					HardwareReservations: &mockHWService{
-						GetFn: func(_ string, _ *packngo.GetOptions) (*packngo.HardwareReservation, *packngo.Response, error) {
-							return nil, nil, fmt.Errorf("boom")
-						},
-					},
+				handler: func(w http.ResponseWriter, r *http.Request) {
+					w.Header().Add("Content-Type", "application/json")
+					w.Header().Add("X-Request-Id", "needed for equinix_errors.FriendlyError")
+					w.WriteHeader(http.StatusInternalServerError)
 				},
 			},
 			wantErr: true,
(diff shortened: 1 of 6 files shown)


equinix/resource_metal_device_acc_test.go
```diff
@@ -1134,21 +1134,21 @@ func testAccWaitForMetalDeviceActive(project, deviceHostName string) resource.Im
...
-		client := meta.(*config.Config).Metal
-		devices, _, err := client.Devices.List(rs.Primary.ID, &packngo.ListOptions{Search: deviceHostName})
+		meta.(*config.Config).AddModuleToMetalGoUserAgent(rd)
+		client := meta.(*config.Config).Metalgo
+		devices, _, err := client.DevicesApi.FindProjectDevices(context.TODO(), rs.Primary.ID).Search(deviceHostName).Execute()
 		if err != nil {
 			return "", fmt.Errorf("error while fetching devices for project [%s], error: %w", rs.Primary.ID, err)
 		}
-		if len(devices) == 0 {
+		if len(devices.Devices) == 0 {
```
displague, line 1143: To save a few lines of change, could have done something like:
```
resp := ...
devices := resp.GetDevices()
```

displague (approved):
lgtm

equinix/helpers_device_test.go
```diff
@@ -134,27 +127,42 @@ func Test_waitUntilReservationProvisionable(t *testing.T) {
...
-						},
-					},
+				handler: func(w http.ResponseWriter, r *http.Request) {
+					reservation := metalv1.HardwareReservation{
+						Device: nil, Provisionable: metalv1.PtrBool(false),
+					}
+
+					body, err := reservation.MarshalJSON()
+					if err != nil {
+						return // TODO ... ?
```
displague, line 137: If this code could be reached, which it seems it will not, we could return a 500 error header and the error in a {errors} body.

ctreatma: Since this is a mock API, and a non-nil error would mean some kind of misconfiguration of the mock, I might lean towards panicking instead to reduce the risk that someone sees an error response from the mock and thinks it's coming from the actual API.

equinix/resource_metal_device.go
```diff
@@ -482,24 +507,18 @@ func reinstallDisabledAndNoChangesAllowed(attribute string) customdiff.ResourceC
...
-	_, ok := d.GetOk("ip_address")
-	if ok {
-		arr := d.Get("ip_address").([]interface{})
-		addressTypesSlice = getNewIPAddressSlice(arr)
-	}
+	createRequest := metalv1.CreateDeviceRequest{}
 
-	createRequest := &packngo.DeviceCreateRequest{
-		Hostname:     d.Get("hostname").(string),
+	/*	Hostname:     d.Get("hostname").(string),
```
ctreatma, line 515: These lines should be gone instead

equinix/resource_metal_device.go
```diff
@@ -726,33 +682,31 @@ func resourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta i
...
 
 func resourceMetalDeviceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	meta.(*config.Config).AddModuleToMetalUserAgent(d)
-	client := meta.(*config.Config).Metal
+	meta.(*config.Config).AddModuleToMetalGoUserAgent(d)
+	client := meta.(*config.Config).Metalgo
+
+	ur := metalv1.DeviceUpdateInput{}
 
+	// TODO does this need to be a separate device update request?
```
ctreatma, line 690: @displague forgot about this question:  the packngo version of this resource handled `locked` changes by running a separate update request that only changes the `locked` property of the device and then running a second update request to change the remaining properties.

I didn't find anything in the docs to indicate that `locked` has to be updated on its own; do we still need to update the `locked` property in a separate request from the other properties?

equinix/resource_metal_device.go
ctreatma: And now they're gone

equinix/helpers_device_test.go
ctreatma: Updated the tests to panic if there's an error here.  Resolving this thread but please feel free to reopen it if needed.

equinix/resource_metal_device_acc_test.go
ctreatma: Updated as described; resolving this thread but please reopen if needed.

equinix/resource_metal_device.go
ctreatma: We reached out to the API team for help understanding the behavior around `locked`; waiting for confirmation that we don't need a separate request to update that field before touching others.  Once we know the behavior we can also document it in the resource docs.

equinix/resource_metal_device.go
```diff
@@ -726,33 +674,30 @@ func resourceMetalDeviceRead(ctx context.Context, d *schema.ResourceData, meta i
...
 
 func resourceMetalDeviceUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	meta.(*config.Config).AddModuleToMetalUserAgent(d)
-	client := meta.(*config.Config).Metal
+	meta.(*config.Config).AddModuleToMetalGoUserAgent(d)
+	client := meta.(*config.Config).Metalgo
+
+	ur := metalv1.DeviceUpdateInput{}
 
 	if d.HasChange("locked") {
```
ctreatma, line 682: So...after digging into what it means for a device to be locked and confirming that we don't have to use a separate update request to change the lock status, it turns out that `locked` is a computed attribute for a `metal_device` resource, so it's not actually possible to hit this code path anyway.

So:  do we want to stick with the existing behavior that `locked` can't be changed or do we want to support locking/unlocking devices in terraform?

For reference, this is the information we got about what locking does:
> It seems that locking a device prevents you from:
> - Delete/deprovision the device
> - Reinstall the device
> - Prevents an instance with a termination time set to be reclaimed, even if the termination time was reached (for spot instances, we will reclaim it anyway)
> - Perform a Firmware update on the device (this feature is not GA yet)
Perform a Firmware update on the device (this feature is not GA yet)

cprivitere: I feel like it's a core state for our devices to support and so terraform should be able to do it.

displague: The list of lock affecting features is helpful context. Without that we may expect behaviors that are not present today: https://feedback.equinixdigital.com/instances/p/expand-lock-feature-for-metal-servers.

It's worth considering the ways Terraform and a Portal user would interact through [EM API device locking](https://metal.equinix.com/developers/api/#tag/Devices/operation/updateDevice) (not to be confused with [Terraform state locking](https://developer.hashicorp.com/terraform/language/state/locking)).

If Terraform takes an "optional" (default=false) behavior, rather than computed, then plan/apply from Terraform would inform users that an externally applied `lock` is going to be reset upon change.

When Terraform config requests `lock`, then any API changes that are incompatible with `lock` would fail until the Terraform config is modified to `lock=false` (or the `lock` line is removed).

I'm not seeing any problems in taking this approach.

ctreatma: OK, sounds like we're all agreed; I'm going to update the resource so that `locked` can be set and update the docs accordingly.

ctreatma: I enabled locking/unlocking a device resource; this PR has been updated but here's a direct link to the commit that has those changes: https://github.com/equinix/terraform-provider-equinix/pull/501/commits/836f1f045e284accc17f47a4801acc4b45f9726f

I opted to make `locked` optional and computed, rather than optional with a default.  IMO changing it from computed to default has a bigger impact on existing users, because if we switch to a default of false then any terraform user who currently has a locked device is at risk for accidentally unlocking the device as a result of a non-major-version provider upgrade.

equinix/resource_metal_device_acc_test.go
```diff
@@ -1210,3 +1168,60 @@ func TestAccMetalDeviceUpdate_timeout(t *testing.T) {
 		},
 	})
 }
+
+func TestAccMetalDevice_LockingAndUnlocking(t *testing.T) {
```
displague, line 1172: lock/unlock are properties of a device. It would be sufficient to bake these test steps (and similar non-destructive attribute modifications) into an existing TestAccMetalDevice test.  I realize that defeats some of the benefits of testing in isolation, but it also reduces the amount of physical servers we need to occupy for real time during tests. 

It's a balancing act. Perhaps we can address that in a separate E2E improvement PR.

displague: In this case, we are verifying that `destroy` is blocked by the lock. I overlooked that. This test is very reasonable.

displague (approved):

displague:
The only E2E failure that doesn't look like flake:
```
=== NAME  TestAccMetalSpotMarketRequestCreate_WithTimeout
    resource_metal_spot_market_request_acc_test.go:245: Step 1/1, expected an error with pattern, no match on: Error running apply: exit status 1
        
        Error: POST https://api.equinix.com/metal/v1/projects/0667bb71-9428-4919-8c1d-58fc0fb95613/spot-market-requests?include=devices%2Cproject%2Cplan: 422 2.21 exceeds the maximum bid price of 2.20 
        
          with equinix_metal_spot_market_request.request,
          on terraform_plugin_test.tf line 122, in resource "equinix_metal_spot_market_request" "request":
         122: resource "equinix_metal_spot_market_request" "request" {
```

PR #502, opened 2023-12-21
refactor: move error functions out of package.go

ctreatma:
This moves some private error-related functions out of `provider.go` into a separate package so that resources and data sources aren't unnecessarily tied to the provider code file.


displague (approved):

thogarty (approved):
LGTM! Ship it.

PR #509, opened 2024-01-08
refactor: move converters out of provider.go

ctreatma:
This moves some type conversion utility functions out of `provider.go` into the existing `converters` package so that they can be safely used by multiple resources and data sources without forcing those resources and data sources to be in the same package as the provider definition.


displague (approved):

displague:
@srushti-patl @yvikrant @thogarty fyi, related to #106

PR #510, opened 2024-01-08
chore: stabilize Metal organization test

### LINKED ISSUE #463: [Bug]: e2e TestAccMetalOrganization_create error 500
TestAccMetalOrganization_create test is been failing for a while, I haven't identified when it last worked fine

https://github.com/equinix/terraform-provider-equinix/actions/runs/6829925647/job/18577020590#step:5:423

ctreatma:
The terraform acceptance test for `metal_organization` creates an organization and then very quickly updates it.  The update step started failing at some point mid-2023 and was briefly fixed, but it has been failing consistently for quite a while now.

There is some background processing that happens when an organization is created, and the API will throw an error if the organization is updated before that background processing is finished.  This adds a delay between the create and update calls for an organization so that the affected test can pass.

Fixes #463.

### DIFF
--- a/equinix/resource_metal_organization_acc_test.go
+++ b/equinix/resource_metal_organization_acc_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"testing"
+	"time"
 
 	"github.com/equinix/terraform-provider-equinix/internal/config"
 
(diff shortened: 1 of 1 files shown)


cprivitere (approved):
lgtm

PR #513, opened 2024-01-10
feat: Adding capability to update cluster name of NE cluster device

schadalawada:
Added changes required to allow customers to update the cluster name of the device.

Reference: https://developer.equinix.com/dev-docs/ne/api-reference/virtual-devices#update-virtual-device

Closes #511


equinix/resource_network_device.go
```diff
@@ -1538,6 +1548,11 @@ func getNetworkDeviceStateChangeConfigs(c ne.Client, deviceID string, timeout ti
 			createNetworkDeviceStatusResourceUpgradeWaitConfiguration(c.GetDevice, deviceID, 5*time.Second, timeout),
 		)
 	}
+	if _, found := changes[neDeviceClusterSchemaNames["ClusterName"]]; found {
```
rling-equinix, line 1551: There won't be any device status changes when changing cluster name, and thus we can remove the status check here.

rling-equinix (changes requested):

rling-equinix (approved):

yvikrant (approved):

PR #516, opened 2024-01-11
fix: send an empty array for IP ranges instead of null

### LINKED ISSUE #515: [Bug]: TestAccMetalVRF_withIPRanges is failing consistently in CI
The test `TestAccMetalVRF_withIPRanges` is failing consistently in CI: https://github.com/equinix/terraform-provider-equinix/actions/runs/7491063515/job/20391401243#step:5:464

The error output is:
```
=== NAME  TestAccMetalVRF_withIPRanges
    resource_metal_vrf_acc_test.go:105: Step 4/4 error: Error running apply: exit status 1
        
        Error: API Error HTTP 422 Parameter ip_ranges cannot be null. Please either exclude it from the request or specify an empty array ([]).
        
          with equinix_metal_vrf.test,
          on terraform_plugin_test.tf line 18, in resource "equinix_metal_vrf" "test":
          18: resource "equinix_metal_vrf" "test" {
```

ctreatma:
Fix #515.

The `TestAccMetalVRF_withIPRanges` started failing recently with an error that `ip_ranges` cannot be null.  It appears this test failure was caused by #509, which consolidated the [`expandListToStringList`](https://github.com/equinix/terraform-provider-equinix/pull/509/files#diff-17a1d27f648a15532c90c0ed8e7b143adde23b59dec5bbe4222db41a0a94ea0aL225-L232) and `IfArrToStringArr` functions.

This updates the `IfArrToStringArr` converter to match the old `expandListToStringList` implementation, which fixes the failing test.

### DIFF
--- a/internal/converters/converters.go
+++ b/internal/converters/converters.go
@@ -17,12 +17,9 @@ func StringArrToIfArr(sli []string) []interface{} {
 }
 
 func IfArrToStringArr(ifaceArr []interface{}) []string {
-	var arr []string
-	for _, v := range ifaceArr {
-		if v == nil {
-			continue
-		}
-		arr = append(arr, v.(string))
+	arr := make([]string, len(ifaceArr))
+	for i, v := range ifaceArr {
+		arr[i] = fmt.Sprint(v)
 	}
 	return arr
 }


ctreatma:
FWIW, I don't really get why this implementation change is needed.

Near as I can tell, this was the line that was leading to test failure: https://github.com/equinix/terraform-provider-equinix/blob/main/equinix/resource_metal_vrf.go#L107

...but it seems like `&ipRanges` should have been non-nil in both cases based on this: https://go.dev/play/p/8vXQ6PWFi0Q

What am I missing?

ctreatma:
Test passing again in CI: https://github.com/equinix/terraform-provider-equinix/actions/runs/7493501381/job/20399350107#step:5:500

ctreatma:
The `IfArrToStringArr` function was introduced in #473; I looked through that PR and (1) the only things that were using that function prior to #509 were Metal resources and data sources and (2) where it _was_ used, it looks like it should be OK to return `[]` in the cases where that would have returned `[]`-but-actually-`nil`.

ocobles (approved):
lgtm. I didn't try it but I tried the playground example

Issue #402, opened 2023-10-04
packngo is deprecated. Use equinix-sdk-go


displague:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

## Resources
* [ ] [equinix_metal_bgp_session](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_bgp_session)
* [x]  [equinix_metal_connection](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_connection)
* [x] [equinix_metal_device](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_device)
* [ ] [equinix_metal_device_network_type](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_device_network_type)
* [ ] [equinix_metal_gateway](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_gateway)
* [ ] [equinix_metal_ip_attachment](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_ip_attachment)
* [ ] [equinix_metal_organization](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_organization)
* [ ] [equinix_metal_organization_member](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_organization_member)
* [ ] [equinix_metal_port](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_port)
* [ ] [equinix_metal_port_vlan_attachment](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_port_vlan_attachment)
* [x] [equinix_metal_project](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_project)
* [ ] [equinix_metal_project_api_key](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_project_api_key)
[...]

PR #517, opened 2024-01-11
refactor: migrate project resource & data source to equinix-sdk-go

ctreatma:
This updates the Equinix Metal project resource and data source from the deprecated `packngo` SDK to `equinix-sdk-go`.

In order to change which SDK is used by the project resource, I also needed to [rewrite some tests so that they use a mock API instead of a mock SDK](https://github.com/equinix/terraform-provider-equinix/pull/517/commits/dc88ae02038b244180167340dfdbc2c79d8a8728).

Relates to #402

### DIFF
--- a/equinix/resource_metal_project.go
+++ b/equinix/resource_metal_project.go
@@ -1,25 +1,28 @@
 package equinix
 
 import (
+	"context"
 	"fmt"
 	"path"
 	"strings"
+	"time"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
 
 	"github.com/equinix/terraform-provider-equinix/internal/config"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-	"github.com/packethost/packngo"
 )
 
 func resourceMetalProject() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceMetalProjectCreate,
-		Read:   resourceMetalProjectRead,
-		Update: resourceMetalProjectUpdate,
-		Delete: resourceMetalProjectDelete,
+		CreateContext: resourceMetalProjectCreate,
+		ReadContext:   resourceMetalProjectRead,
+		UpdateContext: resourceMetalProjectUpdate,
+		DeleteContext: resourceMetalProjectDelete,
 		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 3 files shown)


equinix/resource_metal_project.go
```diff
@@ -1,25 +1,28 @@
...
-	"github.com/packethost/packngo"
 )
 
 func resourceMetalProject() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceMetalProjectCreate,
-		Read:   resourceMetalProjectRead,
-		Update: resourceMetalProjectUpdate,
-		Delete: resourceMetalProjectDelete,
+		CreateContext: resourceMetalProjectCreate,
```
aayushrangwala, line 22: `CreateContext` will by default introduce a default timeout of 20 minutes
https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#default-timeouts-and-deadline-exceeded-errors

We will have to use `CreateWithoutTimeout` where no officially timeout is configured

and same for others

aayushrangwala (approved):

equinix/resource_metal_project.go
ctreatma: It's not clear to me when we should or shouldn't use the functions that can time out.  CRUD operations on a project should generally be very fast, so most of the time it probably won't matter if we use the `...Context` or `...WithoutTimeout` functions.  If, for some reason, it took longer than 20 minutes, I think it would be preferable for the resource to fail than to hang indefinitely.

aayushrangwala: Yes Ideally thats true. But I guess some resources may take more time and users would have to provide that explicitly via timeout block. AFAIK we have [officially documented timeouts](https://github.com/equinix/terraform-provider-equinix/blob/main/docs/resources/equinix_metal_device.md#timeouts) for certain resources and same is being configured in the code [as here](https://github.com/equinix/terraform-provider-equinix/blob/90516e3aa71d94696753723bc6a5a20e685fdc88/equinix/resource_metal_device.go#L43C3-L43C37)

For those places we can have `...Context` and wherever we did not have any timeout configured there we can have `...WithoutTimeout` because we have not declared timeouts so a resource may take more than 20 mins for the user and we are not interrupting it currently

ctreatma: I'm skeptical of the idea that we should let fast actions have the risk of hanging forever simply because we haven't established the exact definition of "fast," but at least 99% of the time the choice of function won't matter for this resource.   In light of that, I've updated the CRUD functions to use the `WithoutTimeout` versions.

aayushrangwala: Yes completely agree with you. Ideally any cloud resource (which is costing user) should have a timeout with the provision to autocleanup failed, unknown status resources.

PR #518, opened 2024-01-11
chore: move SDK provider into its own package

ctreatma:
This is a massive change that will be gradually broken down into other PRs; I'm opening this one as a draft that we can refer back to as we progress.


equinix/data_source_ecx_l2_sellerprofiles.go
```diff
@@ -111,10 +112,10 @@ func dataSourceECXL2SellerProfilesRead(ctx context.Context, d *schema.ResourceDa
 				continue
 			}
 		}
-		if len(metros) > 0 && !atLeastOneStringFound(metros, flattenECXL2SellerProfileMetroCodes(profile.Metros)) {
+		if len(metros) > 0 && !slices.ContainsOne(metros, flattenECXL2SellerProfileMetroCodes(profile.Metros)) {
```
ctreatma, line 115: Everything affected by the move of the private string slice functions is either a NE or an ECXv3 resource.  Might be preferable (lower risk, simpler diff) to move the custom implementations of these functions to a package in `internal` rather than replacing them with `slices` functionality.

ctreatma: The latest version of this PR attempts to make the code changes as minimal as possible; rather than moving things to separate packages or replacing internal functions with "better" alternatives, the updated PR just does whatever is needed to get us to the point of isolating the SDK provider code without changing any existing logic.

ctreatma:
Closing this for now.  Migration of Metal resources/datasources to terraform plugin framework is picking up steam, and as the rest of the existing resources/datasources get migrated over or are deprecated there will be less and less value in having the SDK provider definition sit alongside the framework provider definition.

PR #519, opened 2024-01-17
fix: fixing WaitForConnectionProviderStatus function for AWS connections

srushti-patl:
Added string(v4.PROVISIONING_ProviderStatus) to pending block for AWS connections


thogarty (approved):
LGTM. Shipping it.

PR #520, opened 2024-01-17
test: Add acceptance test for fabric_ports and GHA for fabric tests

thogarty:
* Acceptance Tests Added for:
   * `data "equinix_fabric_port"`
   * `data "equinix_fabric_ports"`
* GHA added for Fabric Acceptance Tests
   * Uses Two sets of Secrets for Different Tests
   * Fabric Tests must be tagged PNFV or PFCR to be picked up by GHA
   * Added placeholder for potential addition of sweepers

### DIFF
--- a/.github/workflows/fabric_acctest.yml
+++ b/.github/workflows/fabric_acctest.yml
@@ -1,4 +1,4 @@
-name: Acceptance Tests
+name: Fabric Acceptance Tests
 # This workflow determines whether a PR comes from an external fork
 # (which requires approval from us) or from a branch on this repository
 # (which means it was made by us and can run immediately).  Once a PR
@@ -67,10 +67,10 @@ jobs:
         run: |
           go build -v .
 
-  test:
+  test-PFNV:
     name: Matrix Test
     needs: build
-    concurrency: acctest
+    concurrency: fabricacctestpfnv
     runs-on: ubuntu-latest
     env:
       EQUINIX_API_ENDPOINT: "https://uatapi.equinix.com"
@@ -110,17 +110,6 @@ jobs:
         run: |
           go test ./... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m
 
-      - name: TF Fabric PFCR acceptance tests
-        timeout-minutes: 180
-        env:
-          TF_ACC: "1"
-          TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
-          EQUINIX_API_CLIENTID: ${{ secrets.EQUINIX_API_CLIENTID_PFCR }}
-          EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PFCR }}
-          METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
-        run: |
-          go test ./... -v -coverprofile coverage_pfcr.txt -covermode=atomic -count 1 -parallel 8 -run "(PFCR)" -timeout 180m
-
       - name: Sweeper PNFV
         if: ${{ always() }}
         env:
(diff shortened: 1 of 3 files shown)


tutkat (approved):

.github/workflows/fabric_acctest.yml
```diff
@@ -0,0 +1,155 @@
+name: Acceptance Tests
```
ctreatma, line 1: If this will be a separate workflow it should have a different name, since there's already a workflow named `Acceptance Tests`.

An alternative to consider is adding the steps for running Fabric acceptance tests into the existing acceptance test workflow; I kind of suspect that using a separate workflow as you did here was the right call, but I'm not sure.

.github/workflows/fabric_acctest.yml
```diff
@@ -0,0 +1,155 @@
...
+# request is not a draft, regardless of whether it is from an internal
+# branch or external fork.
+
+# Any changes to this job, even from internal contributors, require heavy scrutiny.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
+      - 'LICENSE'
```
ctreatma, line 18: If we stick with having separate acceptance test workflows for each service, then it would probably be better to use a `paths:` option and specify the Fabric-relevant path patterns, so you're not running acceptance tests for PRs that only touch NE resources.

.github/workflows/fabric_acctest.yml
```diff
@@ -0,0 +1,155 @@
...
+
+      - name: Get dependencies
+        run: |
+          go mod download
+
+      - name: TF Fabric PNFV acceptance tests
+        timeout-minutes: 180
+        env:
+          TF_ACC: "1"
+          TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
```
ctreatma, line 106: If these different types of tests can safely run in parallel (for example, they won't trample on each other's resources and/or compete for the same, limited capacity), you might want to instead have a separate acceptance test job for each so you get faster feedback when one fails.

equinix/data_source_fabric_port_acc_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func TestAccDataSourceFabricPort_PNFV(t *testing.T) {
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ExternalProviders: testExternalProviders,
+		Providers:         testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testDataSourceFabricPort("c4d85dbe-fa99-a999-f7e0-306a5c00af26"),
```
ctreatma, line 17: At minimum I'd suggest moving this ID to a variable or a constant.  What we've done in similar situations in Metal acceptance tests is to read the ID from an environment variable so we can quickly change the ID if we're testing against a different account or if the resource we're depending on gets deleted accidentally.

.github/workflows/fabric_acctest.yml
thogarty: I think we discussed having a separate workflow would be better awhile back so I stuck with that pattern. Adjusting name though. Missed that.

.github/workflows/fabric_acctest.yml
thogarty: I like it. Will do.

equinix/data_source_fabric_port_acc_test.go
thogarty: How do you handle the ENV variables? I didn't see any in the workflows for IDs. We would need to add a large amount in order to dynamically adjust the tests. I like the idea. What's the pattern you're doing for it right now?

.github/workflows/fabric_acctest.yml
thogarty: Agree. Implementing.

equinix/data_source_fabric_port_acc_test.go
ctreatma: Actually we only have one resource we're tracking, a dedicated connection in our test account.  The ID is stored as a secret and passed in to the test job here: https://github.com/equinix/terraform-provider-equinix/blob/main/.github/workflows/acctest.yml#L110

The code that looks that up is spaghetti but these search results show the strands: https://github.com/search?q=repo%3Aequinix%2Fterraform-provider-equinix%20metalDedicatedConnIDEnvVar&type=code

thogarty: I see. 

Let's say we had 50+ id's or other variables that would be nice to adjust on the fly. That would get very hard to maintain in the workflow file especially when we have to add additional tests that have similar needs. 

Could we postpone coming up with a solution for this for another follow up PR?

thogarty: I like the ENV variable method for Ports. I think that's the only resource that actually needs to exist outside of what we can create in Terraform.

I'll do a quick follow up to this one with that included as I'll be adding additional acceptance tests very soon.

srushti-patl (approved):
LGTM

equinix/data_source_fabric_port_acc_test.go
ctreatma: One option might be to put a JSON string in the env var instead of a single ID, but I'm not sure what kind of length limitations you might run into for a single env var.  Since these IDs are being committed in the clear anyway, you don't need to use a secret for it.

Regardless of the approach you take, I'd still at least recommend defining the IDs in variables and referencing those, rather than hard-coding the same ID in multiple places.

PR #521, opened 2024-01-17
refactor: move resource data change detectors to separate package

ctreatma:
This moves some functions and related tests for detecting changes to resource data out of `provider.go` into its own package so that it's easier to isolate the SDK provider into its own package.  This was broken out of #518


aayushrangwala (approved):

cprivitere (approved):
lgtm

PR #526, opened 2024-01-18
refactor: update tests to use new standalone terraform-plugin-testing module

ocobles:
a go module was released last year to keep testing functionality separated from schema logic https://github.com/hashicorp/terraform-plugin-testing We need this replacement to use some functionality required to test migration from sdkv2 to framework, like plancheck https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#example

Keeping both github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource and github.com/hashicorp/terraform-plugin-testing/helper/resource imports produces a panic error https://github.com/equinix/terraform-provider-equinix/pull/406#issuecomment-1897294576 so it is required to fully replace all

### DIFF
--- a/equinix/data_source_ecx_l2_sellerprofile_acc_test.go
+++ b/equinix/data_source_ecx_l2_sellerprofile_acc_test.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"testing"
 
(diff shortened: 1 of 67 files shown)


ctreatma (approved):
After a couple re-runs, the CI failures appear to be flakes and not related to these changes, so LGTM.

PR #528, opened 2024-01-18
refactor: move network type declarations out of provider.go

ctreatma:
This moves some variables for Equinix Metal network types into its own package.  Removing this from provider.go will make it easer to move the existing SDK provider definition into a separate package from the provider resources & data sources.  This is broken out of


cprivitere (dismissed):
lgtm

PR #530, opened 2024-01-19
refactor: move connection resource & data source to a separate package

ctreatma:
This is broken out of #518 

Moving the connection resource and data source to a separate package makes it easier to ensure that connection-related code remains isolated from the provider implementation and other resources.  As part of this move, unit tests of private functions related to connection speed are split out of the connection resource into a separate file; this enables moving the acceptance tests into a separate package while minimizing code changes.

### DIFF
--- a/internal/resources/metal/metal_connection/speed.go
+++ b/internal/resources/metal/metal_connection/speed.go
@@ -6,25 +6,25 @@ import (
 )
 
 var (
-	Mega          uint64 = 1000 * 1000
-	Giga          uint64 = 1000 * Mega
-	AllowedSpeeds        = []struct {
+	mega          uint64 = 1000 * 1000
+	giga          uint64 = 1000 * mega
+	allowedSpeeds        = []struct {
 		Int uint64
 		Str string
 	}{
-		{50 * Mega, "50Mbps"},
-		{200 * Mega, "200Mbps"},
-		{500 * Mega, "500Mbps"},
-		{1 * Giga, "1Gbps"},
-		{2 * Giga, "2Gbps"},
-		{5 * Giga, "5Gbps"},
-		{10 * Giga, "10Gbps"},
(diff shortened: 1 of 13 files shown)


ctreatma:
> necessary

This is debatable; technically we only need to make the variables public, which we could do without moving them.  Potentially this is instead an opportunity to move the connection resource and data source to a separate package as we did for the Metal SSH key resources and data source(s?).

t0mk:
The `TestSpeedConversion` was not run in the CI

internal/resources/metal/metal_connection/speed.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+var (
+	Mega          uint64 = 1000 * 1000
+	Giga          uint64 = 1000 * Mega
+	AllowedSpeeds        = []struct {
```
t0mk, line 11: Why is `AllowedSpeeds` public? It's not used in the test.

ctreatma:
> The `TestSpeedConversion` was not run in the CI

It ran and passed in the unit test workflow: https://github.com/equinix/terraform-provider-equinix/actions/runs/7617916480/job/20748005965?pr=530#step:6:966

It didn't run in the acceptance test workflow because it doesn't have `Metal` or `Provider` or `Migration` in the name (the test name hasn't changed in this PR, only the location).

internal/resources/metal/metal_connection/speed.go
ctreatma: IMO they should either be all private or all public; as you point out they were only made public because they are referenced in a test.  Since the test that references them is a unit test I've instead opted to keep the test in the same package as the code so that these can remain private; this PR has been updated with those changes.

t0mk (approved):

PR #532, opened 2024-01-22
refactor: move isEmpty out of provider.go

ctreatma:
The isEmpty function isn't a provider feature and is used by multiple resources, so it should go in its own package.

This also removes the `setSchemaValueIfNotEmpty` function because it included a reference to `isEmpty` and the only reference to the function was in a unit test covering the function.


ctreatma:
#518 has been updated to take a different approach, and on closer inspection I question the utility/necessity of `isEmpty`, so I'm closing this for now.

PR #535, opened 2024-01-23
feat: build on go 1.21

ctreatma:
Go 1.21 moved some packages from `golang.org/x` into the standard library; some of those, like [`maps`](https://pkg.go.dev/maps) and [`slices`](https://pkg.go.dev/slices), provide alternatives to some private helper functions in our codebase.  Building with 1.21 enables us to start encouraging the use of those standard library features as part of our various ongoing refactorings.


displague (approved):

displague:
If you are not taking advantage of those new standard library helpers (presumably at outgrowth of generics), can we create a new issue to track the conversions you've identified.

ctreatma:
#537 adds deprecation comments to the functions that would be replaced or reimplimented using stdlib functions

PR #536, opened 2024-01-23
refactor: Framework migration base resource timeouts + metal gateway resource and datasource

ocobles:
- Migration of `equinix_metal_gateway` and datasource `equinix_metal_gateway`
- Definition of `WithTimeouts` (credits to https://github.com/hashicorp/terraform-provider-aws/blob/3d49b511a6e6f6c9ea8cddd4b4ac24b469985469/internal/framework/base.go#L148), intended to be embedded in resources which use the special "timeouts" nested block. It allows set default timeouts, in this case required for the Delete function. This change also allows users to change the default timeout:

```hcl
resource "equinix_metal_gateway" "example" {
  ...
  
  timeouts {
     delete = "50m"
  }
}
```

Resources that require defining default timeouts will need to configure them in its `NewResource` function using:

**r.SetDefaultCreateTimeout(30 * time.Minute)
r.SetDefaultReadTimeout(30 * time.Minute)
r.SetDefaultUpdateTimeout(30 * time.Minute)
r.SetDefaultDeleteTimeout(30 * time.Minute)**

And extend Resource struct with `framework.WithTimeouts`
```go
func NewResource() resource.Resource {
	r := Resource{
		BaseResource: framework.NewBaseResource(
			framework.BaseResourceConfig{
				Name: "equinix_metal_gateway",
			},
		),
	}
	r.SetDefaultDeleteTimeout(20 * time.Minute)

	return &r
}

type Resource struct {
	framework.BaseResource
	framework.WithTimeouts
}
```

In order to keep the resourcheSchema function as standard as possible for all resources, I separated the timeouts block definition to the resource.go file where it will be required to override Schema func as below example:

```go
func (r *Resource) Schema(
	ctx context.Context,
	req resource.SchemaRequest,
	resp *resource.SchemaResponse,
) {
	s := resourceSchema(ctx)
	if s.Blocks == nil {
		s.Blocks = make(map[string]schema.Block)
	}
	s.Blocks["timeouts"] = timeouts.Block(ctx, timeouts.Opts{
		Delete: true,
	})
	resp.Schema = s
}
```

--
part of #365


ocobles:
Note: the sdkv2 version of this resource had a `DiffSuppressFunc` attribute for the ip_reservation_id field. `DiffSuppressFunc` does not exist in framework and the [alternatives](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/fields) are [Custom Types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom), [PlanModifiers](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/plan-modification#framework) field on attribute, or implementation of [ResourceWithModifyPlan](https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/plan-modification#framework) interface. In a first attempt I went with the ResourceWithModifyPlan implementation however after some local test I found it is not required at all, since the framework doesn't generates a diff with the `ip_reservation_id` computed value. So I removed the ModifyPlan implementation, but left the reference [here](https://github.com/equinix/terraform-provider-equinix/pull/536/commits/c846e5f3915e55195bc1f8c08295e2b081810b65#diff-24a31ed6951fe61f35f0e9ba5a436365d819d3fda0f3e0b218884e6bf59b97e6) for future resources with similar requirements.

https://github.com/equinix/terraform-provider-equinix/blob/2c09c5215f8656c0513b4e4df21d1dfdd85ca170/equinix/resource_metal_gateway.go#L65-L81

internal/resources/metal/gateway/resource.go
```diff
@@ -0,0 +1,205 @@
...
+		),
+	}
+	r.SetDefaultDeleteTimeout(20 * time.Minute)
+
+	return &r
+}
+
+type Resource struct {
+	framework.BaseResource
+	framework.WithTimeouts
```
ctreatma, line 33: Since framework forces us to be more explicit about which resources/datasources supports timeouts, we should take the opportunity to document timeouts for resources that need them.  (If there are resources/datasources where we don't need them, we should remove the timeouts instead of documenting them, but it looks like this one was actually using the timeout block.)

ocobles: Updated: https://github.com/equinix/terraform-provider-equinix/pull/536/files#diff-438cdf7ccefabaebbc6e26c352807e69f690d3ea98273db2153ba7a7ce607cd1

ctreatma (approved):

PR #537, opened 2024-01-24
refactor: move acceptance tests to the equinix_test package

ctreatma:
The biggest remaining challenge for moving the SDK provider into a separate package from the provider resources & data sources is the acceptance tests.  The aim of this PR is to touch customer-facing code as little as possible; the changes to resources and data sources is limited to:

- Making some functions public because they are referenced in acceptance tests and in the resource/data source itself
- Moving some private functions to new files because they were only used in acceptance tests but they were written in files that were not acceptance test files


ctreatma:
While this is a large PR, I think it's ultimately worthwhile to do this all at once.  If we migrate the tests a few resources at a time we might as well migrate the resources themselves a few at a time; by isolating the tests in one stroke, we put ourselves in a better position to refactor smaller things in the future.

equinix/helper_test.go
```diff
@@ -0,0 +1,179 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/equinix/ecx-go/v2"
+)
+
+//‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
+// Legacy test helper functions
```
ctreatma, line 12: I hate this file, but it's frankly intended to be ugly and short-lived; this exists because of how our legacy code is structured.  A lot of what's in here will go away as we remove support for deprecated APIs, and the rest has clear alternatives that we can use rather than moving those files to separate packages and making them public.

As much as I dislike this file, having it here is essential for keeping this diff small and getting us in good shape for future smaller refactors.

equinix/helper_test_test.go
```diff
@@ -0,0 +1,43 @@
+package equinix_test
+
```
ctreatma, line 2: This file bugs me too, but again, it's a conscious choice to minimize the code changes here.  Note that `helper_test.go` must have that name in order to be loaded for the tests, so this file is called `helper_test_test.go`; one alternative would be to just drop these tests into `helper_test.go`.

equinix/port_helpers.go
```diff
@@ -342,7 +342,7 @@ func portSanityChecks(cpr *ClientPortResource) error {
 	return nil
 }
 
-func portProperlyDestroyed(port *packngo.Port) error {
+func PortProperlyDestroyed(port *packngo.Port) error {
```
ctreatma, line 345: This file is used in both tests and customer-facing code, so it's made public here, but otherwise doesn't change.

equinix/provider.go
```diff
@@ -249,51 +248,3 @@ func isEmpty(v interface{}) bool {
 		return false
 	}
 }
-
-func slicesMatch(s1, s2 []string) bool {
```
ctreatma, line 253: This was only used in acceptance tests, so it has moved to the abomination that is the `helper_test.go` file

equinix/provider.go
```diff
@@ -249,51 +248,3 @@ func isEmpty(v interface{}) bool {
...
-			}
-		}
-		if !found {
-			return false
-		}
-	}
-	return true
-}
-
-func slicesMatchCaseInsensitive(s1, s2 []string) bool {
```
ctreatma, line 277: This was used in tests and in one resource, so it has been duplicated to facilitate splitting the tests apart.  While we could replace it with newer functions from an external package, the resource that uses it is ECX and will disappear in a few months, and the tests that use it would be better updated by the owning team.  Duplicating the function puts us in a better position for future refactorings than aiming for perfection here.

equinix/provider_test.go
```diff
@@ -1,120 +1,11 @@
...
 	"testing"
 
-	"github.com/equinix/ecx-go/v2"
-	"github.com/equinix/terraform-provider-equinix/internal/config"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/stretchr/testify/assert"
 )
 
-var (
```
ctreatma, line 17: Everything removed here went to `helper_test.go`; while we could put a different file structure in place, most if not all of this code should go away ASAP, so dumping it all in one file is meant as motivation to eliminate this stuff entirely in the future.

equinix/resource_fabric_cloud_router.go
```diff
@@ -231,14 +231,14 @@ func resourceCloudRouterDelete(ctx context.Context, d *schema.ResourceData, meta
...
-	err = waitUntilCloudRouterDeprovisioned(d.Id(), meta, ctx)
+	err = WaitUntilCloudRouterDeprovisioned(d.Id(), meta, ctx)
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("API call failed while waiting for resource deletion. Error %v", err))
 	}
 	return diags
 }
 
-func waitUntilCloudRouterDeprovisioned(uuid string, meta interface{}, ctx context.Context) error {
+func WaitUntilCloudRouterDeprovisioned(uuid string, meta interface{}, ctx context.Context) error {
```
ctreatma, line 241: This is another function that was referenced in acceptance tests and customer-facing code.  Again, the aim of this PR is to get acceptance tests isolated with minimal logic changes, so I've opted to make this function public.  When this resource is migrated to a separate package there will be an opportunity to revisit this choice.

equinix/resource_fabric_connection.go
```diff
@@ -394,14 +394,14 @@ func resourceFabricConnectionDelete(ctx context.Context, d *schema.ResourceData,
 		return diag.FromErr(equinix_errors.FormatFabricError(err))
 	}
 
-	err = waitUntilConnectionDeprovisioned(d.Id(), meta, ctx)
+	err = WaitUntilConnectionDeprovisioned(d.Id(), meta, ctx)
```
ctreatma, line 397: Same as above, this is made public in-place in order to separate the customer-facing code from the acceptance tests with minimal changes to the customer-facing code.

equinix/resource_fabric_routing_protocol.go
```diff
@@ -264,7 +264,7 @@ func resourceFabricRoutingProtocolDelete(ctx context.Context, d *schema.Resource
 		return diag.FromErr(equinix_errors.FormatFabricError(err))
 	}
 
-	err = waitUntilRoutingProtocolIsDeprovisioned(d.Id(), d.Get("connection_uuid").(string), meta, ctx)
+	err = WaitUntilRoutingProtocolIsDeprovisioned(d.Id(), d.Get("connection_uuid").(string), meta, ctx)
```
ctreatma, line 267: Again with the making a function public in-place

equinix/resource_metal_device.go
```diff
@@ -813,7 +813,7 @@ func resourceMetalDeviceDelete(ctx context.Context, d *schema.ResourceData, meta
 	return nil
 }
 
-func waitForActiveDevice(ctx context.Context, d *schema.ResourceData, meta interface{}, timeout time.Duration) error {
+func WaitForActiveDevice(ctx context.Context, d *schema.ResourceData, meta interface{}, timeout time.Duration) error {
```
ctreatma, line 816: Waiter, there's a public function in my soup

equinix/resource_metal_port.go
```diff
@@ -239,7 +239,7 @@ func resourceMetalPortDelete(ctx context.Context, d *schema.ResourceData, meta i
 			}
 		}
 		// TODO(displague) error or warn?
-		if warn := portProperlyDestroyed(cpr.Port); warn != nil {
+		if warn := PortProperlyDestroyed(cpr.Port); warn != nil {
```
ctreatma, line 242: This is updated to match the change in `port_helpers.go`

internal/acceptance/acceptance.go
```diff
@@ -45,6 +45,26 @@ func init() {
 	}
 }
 
+func TestAccPreCheck(t *testing.T) {
```
ctreatma, line 48: When the SSH key resources were moved out of the `equinix` package, the `TestAccPreCheckForMetal` function was added.  That function only checks for Metal-related env vars.  This one is a drop-in replacement for the private `testAccPreCheck` function in order to keep the diff smaller; we can revisit this implementation in a future refactor.

equinix/helper_test.go
thogarty: Thanks for the context 👍

internal/acceptance/env.go
```diff
@@ -0,0 +1,20 @@
+package acceptance
+
+import (
+	"fmt"
+	"os"
+)
+
+func GetFromEnvDefault(varName string, defaultValue string) string {
```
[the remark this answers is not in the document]
thogarty: From an earlier PR comment about desiring to use ENV variables for testing I think it makes sense to keep them. Having them separate and public is a good method considering the desire to remove the `helper_test.go` and `helper_test_test.go` files.

equinix/helper_test_test.go
thogarty: Wouldn't be a bad idea just to move it into `helper_test.go`. There's not a lot here.

equinix/provider_test.go
thogarty: Another piece of evidence for eliminating the extra `helper_test_test.go` file and just dumping its contents into `helper_test.go`.

thogarty (approved):
This LGTM. I would dump `helper_test_test.go` into `helper_test.go` though so we only have one file that ultimately needs to be removed once we discard the code that it is being used for.

Thanks for the refactor, @ctreatma !

tutkat (approved):

thogarty (approved):
LGTM.

PR #538, opened 2024-01-24
fix: correctly set organization_id in metal_project data source

ctreatma:
The metal_project data source was attempting to read the organization ID for the specified project by looking directly at the `organization.id` field.  That field is only present on a project when the API request has `organization` in the `include` parameter.  Rather than include organization details and make the data source slower, this replicates the logic from the metal_project resource, which uses the organization's `href` field to get the organization ID.

While I was in this area of the code I also removed some debugging print statements that were merged unintentionally.

### DIFF
--- a/internal/resources/metal/project/datasource.go
+++ b/internal/resources/metal/project/datasource.go
@@ -143,7 +143,7 @@ func dataSourceMetalProjectRead(ctx context.Context, d *schema.ResourceData, met
 	d.Set("payment_method_id", path.Base(project.PaymentMethod.GetHref()))
 	d.Set("name", project.GetName())
 	d.Set("project_id", project.GetId())
-	d.Set("organization_id", path.Base(project.Organization.GetId())) // Should be gethref?
+	d.Set("organization_id", path.Base(project.Organization.AdditionalProperties["href"].(string))) // spec: organization has no href
 	d.Set("created", project.GetCreatedAt().Format(time.RFC3339))
 	d.Set("updated", project.GetUpdatedAt().Format(time.RFC3339))
 	d.Set("backend_transfer", project.AdditionalProperties["backend_transfer_enabled"].(bool)) // No backend_transfer_enabled property in API spec
--- a/internal/resources/metal/project/datasource_test.go
+++ b/internal/resources/metal/project/datasource_test.go
@@ -32,6 +32,9 @@ func TestAccDataSourceMetalProject_byId(t *testing.T) {
 					resource.TestCheckResourceAttrPair(
 						"equinix_metal_project.foobar", "id",
 						"data.equinix_metal_project.test", "id"),
+					resource.TestCheckResourceAttrPair(
+						"equinix_metal_project.foobar", "organization_id",
+						"data.equinix_metal_project.test", "organization_id"),
 				),
 			},
 		},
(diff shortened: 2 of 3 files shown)


equinix/data_source_metal_project.go
```diff
@@ -143,7 +143,7 @@ func dataSourceMetalProjectRead(ctx context.Context, d *schema.ResourceData, met
 	d.Set("payment_method_id", path.Base(project.PaymentMethod.GetHref()))
 	d.Set("name", project.GetName())
 	d.Set("project_id", project.GetId())
-	d.Set("organization_id", path.Base(project.Organization.GetId())) // Should be gethref?
+	d.Set("organization_id", path.Base(project.Organization.AdditionalProperties["href"].(string))) // spec: organization has no href
```
displague, line 146: We could easily miss updating this line when/if the API spec is updated to have an HREF field. The string would then return empty with no error.

I could see a metal/ scoped helper being handy for this use-case which may be more common.

(ChatGPT for the quick assist) https://play.golang.com/p/YmGrpcJdvN-

ctreatma: The ChatGPT helper is solving a different problem from the one you mention.  In this case, the object doesn't have a `GetHref` but it does have an `href` so we have to look in `AdditionalProperties` instead.  While we could add a helper for this, I'm hesitant to add a helper for working around spec defects.  "The EM API spec failed to declare a meaningful, common property so I have to look in an arbitrary bag of extra stuff to find it" should not be a common use-case.   (I agree, though, that "the EM API sometimes provides an `id` but sometimes provides an `href` that has an ID" is a common use-case in Metal and a helper could be handy for that.)

As a stopgap for this particular issue I'm leaning towards having the test check that the `organization_id` is not `"."` (the return value of `path.Base` when its argument is empty).

displague: I've wanted an HrefOrId helper before and AdditionalProperties slipped out of mind while I was prompting. We could extend the example with an interface based on `GetAdditionalProperties` and since it returns a well defined map type, we could fallback to AdditionalProperties when GetHref or GetId are not present or empty. That said, I understand not wanting to introduce this for the reasons given.

One other scenario where that I've but couldn't say where (ports?), is that IDs can appear in an `href` URL anchor. `path.Base` is not sensitive to URL semantics and could consider the `#` to be part of the ID. `url.Parse` and `.Fragment` would be better fit in those cases.

ctreatma: I briefly added a helper but ended up pulling it out because we're not actually at risk of missing a spec change; if `href` becomes a named attribute it disappears from `AdditionalProperties` and we'll get a panic in the tests: https://github.com/equinix/terraform-provider-equinix/actions/runs/7660188778/job/20876969697#step:5:262

ocobles: I agree with @ctreatma's argument, and the test should serve to identify future changes to the API. As an extra check, we can validate in the test that the extracted ID is a valid UUID, that will ensure we don't miss that `#section` use case @displague mentioned

ocobles: btw, similar to this issue I found today that https://github.com/equinix/terraform-provider-equinix/blob/4f544011b23c24cce94d354131da06970644c779/internal/resources/metal/metal_connection/resource.go#L411 `conn.Ports[0].VirtualCircuits[0].Project.ID` is empty and we should take the ID from `path.Base(conn.Ports[0].VirtualCircuits[0].Project.URL)`

ocobles (approved):
lgtm, that test using the datasource makes it look secure against possible API changes

PR #540, opened 2024-01-25
test: Update Fabric GHA workflow

thogarty:
* Upload coverage artifacts in html format to see Go coverage details for each run
* Add empty test sweepers for Fabric to avoid failing jobs on sweepers that aren't related to Fabric tests

### DIFF
--- a/.github/workflows/fabric_acctest.yml
+++ b/.github/workflows/fabric_acctest.yml
@@ -125,11 +125,18 @@ jobs:
           go test ${SWEEP_DIR} -v -timeout 180m -sweep=${SWEEP} -sweep-allow-failures=${SWEEP_ALLOW_FAILURES} -sweep-run="nothing"
 
       - name: Upload coverage to Codecov
+        concurrency: codecov-upload
         if: ${{ always() }}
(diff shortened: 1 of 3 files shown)


.github/workflows/fabric_acctest.yml
```diff
@@ -193,3 +203,13 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage_pfcr.txt
+
+      - name: Create Go Coverage HTML File
```
ctreatma, line 207: It looks like the coverage is already being uploaded to CodeCov; what's the use case for also creating an actions artifact?

thogarty: When I was first looking at the page that it was being uploaded to I was getting some weird errors and wasn't able to navigate the page. Looks like it's cleared up so I agree the redundancy isn't necessary.

Is there a way to output just the `coverage.txt` file from codecov if we ever wanted to use the `go tool` for coverage reports in html?

srushti-patl (approved):
LGTM!

PR #541, opened 2024-01-25
refactor: move Metal VLAN resource and data source to equinix-sdk-go

ctreatma:
This refactors the VLAN resource and data source to use equinix-sdk-go instead of packngo.  As part of this refactoring I've moved the VLAN resource and data source to an isolated package.

PR is in draft for now because there are a number of spec issues that need to be addressed and I want to get a feel for how manageable it is to move resources & change the SDK in one PR.

Part of #402 
Part of #106

### DIFF
--- a/internal/resources/metal/vlan/datasource.go
+++ b/internal/resources/metal/vlan/datasource.go
@@ -129,7 +129,7 @@ func dataSourceMetalVlanRead(ctx context.Context, d *schema.ResourceData, meta i
 		"vlan_id":     vlan.GetId(),
 		"project_id":  vlan.AssignedTo.GetId(), // vlan assigned_to is an href; should be project?
 		"vxlan":       vlan.GetVxlan(),
-		"facility":    vlan.FacilityCode, // facility is deprecated, vlan is metro-scoped; remove this attr?
+		"facility":    nil, //vlan.FacilityCode, // facility is deprecated, vlan is metro-scoped; remove this attr?
 		"metro":       vlan.MetroCode,
 		"description": vlan.Description,
 	}))
--- a/internal/resources/metal/vlan/datasource.go
+++ b/internal/resources/metal/vlan/datasource.go
@@ -1,7 +1,7 @@
 package vlan
 
 import (
-	"fmt"
+	"context"
 
 	"github.com/equinix/terraform-provider-equinix/internal/converters"
 
@@ -10,13 +10,14 @@ import (
 
 	"github.com/equinix/terraform-provider-equinix/internal/config"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/packethost/packngo"
 )
 
 func DataSource() *schema.Resource {
 	return &schema.Resource{
-		Read: dataSourceMetalVlanRead,
+		ReadWithoutTimeout: dataSourceMetalVlanRead,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
@@ -73,8 +74,8 @@ func DataSource() *schema.Resource {
 	}
 }
 
-func dataSourceMetalVlanRead(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*config.Config).Metal
+func dataSourceMetalVlanRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*config.Config).Metalgo
 
 	projectRaw, projectOk := d.GetOk("project_id")
 	vxlanRaw, vxlanOk := d.GetOk("vxlan")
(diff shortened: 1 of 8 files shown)


internal/resources/metal/vlan/datasource.go
```diff
@@ -83,77 +84,53 @@ func dataSourceMetalVlanRead(d *schema.ResourceData, meta interface{}) error {
...
-	return equinix_schema.SetMap(d, map[string]interface{}{
-		"vlan_id":     vlan.ID,
-		"project_id":  vlan.Project.ID,
-		"vxlan":       vlan.VXLAN,
-		"facility":    vlan.FacilityCode,
+	return diag.FromErr(equinix_schema.SetMap(d, map[string]interface{}{
+		"vlan_id":     vlan.GetId(),
+		"project_id":  vlan.AssignedTo.GetId(), // vlan assigned_to is an href; should be project?
+		"vxlan":       vlan.GetVxlan(),
+		"facility":    vlan.FacilityCode, // facility is deprecated, vlan is metro-scoped; remove this attr?
```
ctreatma, line 132: In `metal-cli` we opted to [remove `facility` from VLAN subcommands because facility is deprecated and VLANs are at the metro level anyway;](https://github.com/equinix/metal-cli/pull/315/files#r1276358558) however I think removing it entirely could be a bigger impact in terraform so we might want to maintain support for it for now?

internal/resources/metal/vlan/resource.go
```diff
@@ -81,96 +82,103 @@ func resourceMetalVlan() *schema.Resource {
...
-	client := meta.(*config.Config).Metal
+func resourceMetalVlanDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	meta.(*config.Config).AddModuleToMetalGoUserAgent(d)
+	client := meta.(*config.Config).Metalgo
 
 	id := d.Id()
-	vlan, resp, err := client.ProjectVirtualNetworks.Get(id, &packngo.GetOptions{Includes: []string{"instances", "instances.network_ports.virtual_networks", "internet_gateway"}})
-	if equinix_errors.IgnoreResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(resp, err) != nil {
-		return equinix_errors.FriendlyError(err)
+	vlan, resp, err := client.VLANsApi.GetVirtualNetwork(ctx, id).Include([]string{"instances", "instances.network_ports.virtual_networks", "internet_gateway"}).Execute()
```
ctreatma, line 146: It's possible a different include would fix the issue here (at least partially; may need spec changes so the SDK can correctly identify types)
```suggestion
	vlan, resp, err := client.VLANsApi.GetVirtualNetwork(ctx, id).Include([]string{"instances", "virtual_networks", "internet_gateway"}).Execute()
```

internal/resources/metal/vlan/resource.go
```diff
@@ -81,96 +82,103 @@ func resourceMetalVlan() *schema.Resource {
 	}
 }
 
-func resourceMetalVlanCreate(d *schema.ResourceData, meta interface{}) error {
-	meta.(*config.Config).AddModuleToMetalUserAgent(d)
-	client := meta.(*config.Config).Metal
+func resourceMetalVlanCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	meta.(*config.Config).AddModuleToMetalGoUserAgent(d)
+	client := meta.(*config.Config).Metalgo
```
ctreatma, line 87: I always trip myself up on updating these two lines.  If `config.Config` had `GetMetalClient(d *schema.ResourceData)`, `GetFabricClient(d *schema.ResourceData)`, etc. functions we could handle the User-Agent modification there so it doesn't have to be replicated in every CRUD function in every resource & data source.

internal/acceptance/acceptance.go
```diff
@@ -62,6 +62,7 @@ func getFromEnvDefault(varName string, defaultValue string) string {
 	return defaultValue
 }
 
+// TODO: seems like this is sweeper config and broadly used; why is it non-standard?
 func GetConfigForNonStandardMetalTest() (*config.Config, error) {
```
ctreatma, line 66: Doesn't necessarily need to happen in this PR, but IMO this should be `GetConfigForSweeper` and should fully replicate the behavior of the analogous `sharedConfigForRegion` in the `equinix` package

internal/resources/metal/vlan/datasource.go
ctreatma: Based on API responses I saw in Chrome developer tools, both `facility` and `facility_code` are `null` anyway, so maybe this is a non-issue?  Perhaps for now we just explicitly set it to null everywhere in code and deprecate the attribute in the resource & data source schemas so we can remove it later?

internal/resources/metal/vlan/resource.go
ctreatma: > every resource & data source

Turns out we're only modifying the User-Agent in data sources.  Why is that?

displague:
Will need rebasing after #578

ctreatma:

PR #543, opened 2024-01-25
chore: fix NE path mappings and default owners in CODEOWNERS

ctreatma:
One of the path mappings specified for the Network Edge (NE) team was `*equinix_network*` which didn't match any files.  This replaces that line with 2 lines to match NE resources and data sources.  We could use `*network*`, but that would catch `resource_metal_device_network_type.go`.

This also updates the default owners for code in this repo to DevRel Engineering, since we should be responsible for any code in this repo that is not specific to a particular service.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,6 @@
 # See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file
 #
-*                       @equinix/governor-metal-client-interfaces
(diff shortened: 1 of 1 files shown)


displague:
This file was created before we started breaking every service out. Should we reflect the new service names while we are here? (would it be `**/services/ne/**` ?)

displague (approved):

ctreatma:
> This file was created before we started breaking every service out. Should we reflect the new service names while we are here? (would it be **/services/ne/** ?)

I think we should preemptively do that, but perhaps in a separate PR that also moves `internal/resources/` to `internal/services/`

PR #548, opened 2024-01-31
refactor: fabric connection and fabric port schema updates

thogarty:
* Consolidated+Migrated Fabric Connection Schema to Fabric Connection Resource
* Refactored Common Schema Elements to `internal` Package
* Consolidated+Migrated Fabric Port Schema to Fabric Port Resource
* Updated Service Profile, Routing Protocol, and Cloud Router Schema to use Common elements
* Updated Out of Date Documentation to match current changes
* Moved mapping helpers to their corresponding resource + common mapping helpers to `internal` package


srushti-patl (approved):
LGTM!

docs/data-sources/equinix_fabric_cloud_router.md
```diff
@@ -25,15 +25,16 @@ data "equinix_fabric_cloud_router" "cloud_router_data_name" {
 
 ### Optional
 
+- `project` (Block Set) Project information (see [below for nested schema](#nestedblock--project))
```
displague, line 28: How does a FCR lookup by project work? Is there a one FCR resource per project limitation? Does the datasource error if there are more than one FCR in the selected project? Is that a reasonable launch filter that will need to be augmented with additional filters in the future (name, location, feature based?)

Looking a little closer, it appears that `project` is not used when looking up FCR resources, only the `uuid` is used.

displague: I think what is intended is that the datasource would have `uuid` required.

equinix/resource_fabric_connection.go
```diff
@@ -18,6 +19,574 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+func FabricConnectionResourceSchema() map[string]*schema.Schema {
```
displague, line 22: This was previously a private in the package: `createFabricConnectionResourceSchema`. This is preferable since we don't want or expect outside users to depend on this function. Removing exported functions at a later date is a breaking change.

In the future migration to `internal/resources/fabric/connection/` (perhaps `internal/services/fabric/connection/`) the function could be exported and shared between other packages in the Terraform provider, without including the function in the public signature of the `github.com/equinix/terraform-provider-equinix/equinix` package.  That package will minimally expose the `Provider`.

equinix/resource_fabric_service_profile.go
```diff
@@ -22,6 +22,171 @@ import (
...
+		},
+		"bandwidth_from_api": {
+			Type:        schema.TypeBool,
+			Computed:    true,
+			Description: "Bandwidth from api",
+		},
+		"integration_id": {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: "Integration id",
```
displague, line 140: Curious what this field is used for. The description doesn't provide more information than the name.

displague: Same comment about bandwidth_from_api.

docs/data-sources/equinix_fabric_connection.md
```diff
@@ -62,15 +63,16 @@ Read-Only:
 
 - `account` (Set of Object) (see [below for nested schema](#nestedobjatt--a_side--access_point--account))
 - `authentication_key` (String)
-- **Deprecated** `gateway` Use `router` instead (Set of Object) (see [below for nested schema](#nestedobjatt--a_side--access_point--router))
-- `router` CloudRouter; Replaces `gateway` attribute (Set of Object) (see [below for nested schema](#nestedobjatt--a_side--access_point--router))
+- `gateway` (Set of Object) (see [below for nested schema](#nestedobjatt--a_side--access_point--gateway))
```
displague, line 66: Curious how we lost the deprecation on this field

equinix/resource_fabric_connection.go
displague: > Removing exported functions at a later date is a breaking change

because this project was released at 1.0.0 in the past.

https://0ver.org/ 🙂

internal/schema/set_map.go
```diff
@@ -20,9 +20,6 @@ func SetMap(d *schema.ResourceData, m map[string]interface{}) error {
 		if f, ok := v.(setFn); ok {
 			err = f(d, key)
 		} else {
-			if key == "router" {
```
ctreatma, line 23: Yay!

internal/schema/resourcedata.go
```diff
@@ -2,7 +2,7 @@ package schema
 
 import "reflect"
 
-// resourceDataProvider provies interface to schema.ResourceData
+// resourceDataProvider proxies interface to schema.ResourceData
```
ctreatma, line 5: 👍

internal/schema/fabric_common_mapping_helpers.go
```diff
@@ -0,0 +1,230 @@
...
+		purchaseOrderNumber := orderMap["purchase_order_number"]
+		billingTier := orderMap["billing_tier"]
+		orderId := orderMap["order_id"]
+		orderNumber := orderMap["order_number"]
+		order = v4.Order{PurchaseOrderNumber: purchaseOrderNumber.(string), BillingTier: billingTier.(string), OrderId: orderId.(string), OrderNumber: orderNumber.(string)}
+	}
+	return order
+}
+
+func OrderToTerra(order *v4.Order) *schema.Set {
```
ctreatma, line 25: Since you're moving these `fabric_common_` files anyway, I recommend creating a separate directory/package for them, perhaps:
- `internal/fabric/schema/mapping.go`
- `internal/fabric/schema/schemas.go`

Having it in a separate package helps clarify that this `OrderToTerra`, for example, is only useful in the context of Fabric integrations and should probably not be modified by other teams.

docs/data-sources/equinix_fabric_cloud_router.md
thogarty: Great catch. Agree completely. Will update.

equinix/resource_fabric_connection.go
thogarty: Copy that. Heard and will fix here and take that review forward into further Provider changes.

internal/schema/fabric_common_mapping_helpers.go
thogarty: Yea, good point. Will make that change.

docs/data-sources/equinix_fabric_connection.md
thogarty: Ah yes, I had fixed that in the resource, but forgot the data_source. Will update. Good comment.

equinix/resource_fabric_service_profile.go
thogarty: For this one, and some of the others, the unfortunate answer is that I don't think they actually are useful to the terraform users. They are apart of the Fabric API schema, but are not commonly (if ever) used by a customer to do anything. Difficult to get more information on them but I will double check.

For things like that I would actually like to remove them, but there would be a breaking change as you referenced because we've published v1.0.0 already.

thogarty: Nevermind, I found the details needed to provide a better description. Link here for reference: https://developer.equinix.com/docs?page=/dev-docs/fabric/overview

ctreatma (approved):
👍 on the changes to the common files

srushti-patl (approved):
LGTM

Issue #542, opened 2024-01-25
E2E tests fail with `text file busy`


ctreatma:
### Description

Every so often our tests fail with a `text file busy` error: https://github.com/equinix/terraform-provider-equinix/actions/runs/7660191910/job/20876975492#step:5:278

```
cannot run Terraform provider tests: error calling terraform version command: fork/exec /tmp/plugintest-terraform130545137/terraform: text file busy
```

This error is sporadic enough at the moment that we can reliably re-run the failed jobs, but we should keep an eye on this and see if there's something we can/should do in our test workflow to prevent these failures.

ctreatma:
This seems to have been closed by #551

PR #551, opened 2024-01-31
chore: update GHA workflows to use hashicorp/setup-terraform action

ctreatma:
Relates to #542 

We've seen an increase in `text file busy` failures during test runs on some PRs.  After searching for that error message in GitHub issues for other providers, I landed on https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests#github-actions-workflow, which recommends using [the hashicorp/setup-terraform action](https://github.com/hashicorp/setup-terraform) in GitHub Actions workflows.

This updates our acceptance test workflows to use the setup action to install terraform, in the hopes that this reduces or eliminates the `text file busy` errors.

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -97,14 +97,15 @@ jobs:
       run: |
         go mod download
 
+    - uses: hashicorp/setup-terraform@v3
+      with:
+        terraform_version: ${{ matrix.terraform }}
+        terraform_wrapper: false
+
     - name: TF acceptance tests
       timeout-minutes: 180
       env:
         TF_ACC: "1"
-        TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
-        # TF_SCHEMA_PANIC_ON_ERROR: "1"
-        # TF_LOG: "DEBUG"
-        #
         EQUINIX_API_TOKEN: ${{ secrets.EQUINIX_API_TOKEN }}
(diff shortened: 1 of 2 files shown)


ctreatma:
*NOTE:* the acceptance test workflows kicked off automatically for this PR are going to use the old workflow configuration because we use the `pull_request_target` trigger.  I will drop links to manual runs of this branch's configuration when they are available.

ctreatma:
Manual run of fabric acctest workflow that uses hashicorp/setup-terraform action: https://github.com/equinix/terraform-provider-equinix/actions/runs/7733004450/job/21084207834

ctreatma:
Manual run of Metal acctest workflow is happening here: https://github.com/equinix/terraform-provider-equinix/actions/runs/7733452047/job/21085565698

displague:
The latest version of the action is v3 - https://github.com/hashicorp/setup-terraform/releases/tag/v3.0.0

It is surprising that a tool that installs a terraform binary is necessary when our tests are using the go binary and imported packages. From the original test failure, it is clear a `terraform` binary is being run. I would imagine that a test built binary would be the one run rather than the one installed by this action.

Since the upstream testing instructions specify that this action should be run, I can't argue with adding it. I just don't understand how it should make a difference.

Having hit this error in Jenkins before, I looked for GitHub Action + Docker issue reports. https://github.com/moby/moby/issues/13594#issuecomment-263335330 for example.

displague (dismissed):

ocobles (approved):

PR #557, opened 2024-02-02
refactor: Migrate VRF to equinix-sdk-go and move into separate package

aayushrangwala:
This updates the Equinix Metal project resource and data source from the deprecated packngo SDK to equinix-sdk-go and moves them into separate package required for supporting plugin framework

Relates to https://github.com/equinix/terraform-provider-equinix/issues/402

### DIFF
--- a/equinix/resource_metal_virtual_circuit_acc_test.go
+++ b/equinix/resource_metal_virtual_circuit_acc_test.go
@@ -14,6 +14,10 @@ import (
 	"github.com/packethost/packngo"
 )
 
+const (
+	metalDedicatedConnIDEnvVar = "TF_ACC_METAL_DEDICATED_CONNECTION_ID"
+)
+
 func init() {
 	resource.AddTestSweepers("equinix_metal_virtual_circuit", &resource.Sweeper{
 		Name:         "equinix_metal_virtual_circuit",
--- a/internal/resources/metal/vrf/datasource.go
+++ b/internal/resources/metal/vrf/datasource.go
@@ -1,4 +1,4 @@
-package equinix
+package vrf
 
 import (
 	"context"
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func dataSourceMetalVRF() *schema.Resource {
+func DataSource() *schema.Resource {
 	return &schema.Resource{
 		ReadWithoutTimeout: dataSourceMetalVRFRead,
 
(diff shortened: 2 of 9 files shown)


internal/resources/metal/vrf/datasource.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func dataSourceMetalVRF() *schema.Resource {
+func DataSource() *schema.Resource {
```
ctreatma, line 10: While we're in here, let's add a test of the data source in `datasource_test.go` (I realize it shares almost all of its code with the resource, but that may not always be the case, so having a test helps protect against unintended consequences later).

equinix/resource_metal_virtual_circuit_acc_test.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/packethost/packngo"
 )
 
+const (
```
ctreatma, line 17: I think this might've leaked in from another change?

aayushrangwala: No, this constant was being reused in virtual circuit as well. Hence had to copy that separately

internal/resources/metal/vrf/datasource_test.go
```diff
@@ -0,0 +1,119 @@
...
+	var vrf metalv1.Vrf
+	rInt := acctest.RandInt()
+
+	datasourceName := "data.equinix_metal_vrf.foobar"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                  func() { acceptance.TestAccPreCheckMetal(t) },
+		PreventPostDestroyRefresh: true,
+		ExternalProviders:         acceptance.TestExternalProviders,
+		ProtoV5ProviderFactories:  acceptance.ProtoV5ProviderFactories,
```
ctreatma, line 89: It looks like these lines should be removed for the upgrade tests.  We can only declare providers at 1 level; I looked at the upgrade test for the SSH key resource and it doesn't have these provider declarations, only the ones you already have under `Steps`.

```suggestion
```

aayushrangwala: Yes right removed that and have on my local. Receiving different error now. Trying to fix that

ocobles: the two `_upgradeFromVersion` should be removed as this resource has not yet been migrated to the framework and then you cannot use the ProtoV5ProviderFactories

PR #559, opened 2024-02-05
test: fabric_service_profile resource/data_source acceptance tests Feature/cxf 83390

tutkat:
Added acceptance test for service profile


equinix/data_source_fabric_service_profile_acc_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"fmt"
+	"github.com/equinix/terraform-provider-equinix/internal/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"regexp"
+	"testing"
+)
+
+func testAccFabricReadServiceProfileConfig(uuid string) string {
```
thogarty, line 11: We don't want to rely on Service Profiles being created already for our Data_Source Reads. Go ahead and create one in the same config as the data_sources, and have them use it with `equinix_fabric_service_profile.resource_name.id` or `equinix_fabric_service_profile.resource_name.name`

Then our tests won't have the dependency on a Service Profile that could be deleted from our UAT environment and break our tests.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,134 @@ import (
...
-			operator = "="
-			values = ["%s"]
-		}
-	}
-`, name)
-}
-
-func TestAccFabricReadServiceProfileByUuid(t *testing.T) {
+func TestAccFabricCreateServiceProfile_SP_FCR_ba(t *testing.T) {
+	spName1 := "fabric_tf_acc_test_01"
```
thogarty, line 21: Our resource names should end in `_PFCR` so that if we need to sweep them up (i.e. delete them because the Unit Tests didn't for some reason) then we can.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,134 @@ import (
...
-			property = "/name"
-			operator = "="
-			values = ["%s"]
-		}
-	}
-`, name)
-}
-
-func TestAccFabricReadServiceProfileByUuid(t *testing.T) {
+func TestAccFabricCreateServiceProfile_SP_FCR_ba(t *testing.T) {
```
thogarty, line 20: All test names need to end in `_PFCR` or `_PNFV` depending on which user they have so that our test running will only run fabric tests with the correct suffix.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,134 @@ import (
...
+	spName1 := "fabric_tf_acc_test_01"
+	spName2 := "fabric_tf_acc_test_02"
+	typel2 := "L2_PROFILE"
+	portType := "XF_PORT"
+	ports := GetFabricEnvPorts(t)
+	portTcFirst := ports["pfcr"]["dot1q"][0]
```
thogarty, line 26: These port calls will sometimes cause an issue if the ENV variables aren't set. Like in the `Go Tests` GHA workflow for the repo. You'll need to wrap them like we did in connections acceptance tests.

```
ports := GetFabricEnvPorts(t)
connectionsTestData := GetFabricEnvConnectionTestData(t)
var publicSPName, portUuid string
if len(ports) > 0 && len(connectionsTestData) > 0 {
  publicSPName = connectionsTestData["pfcr"]["publicSPName"]
  portUuid = ports["pfcr"]["dot1q"][0].Uuid
}
```

This protects the tests and allows them to be skipped when the test environment doesn't have `TF_ACC` set.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,134 @@ import (
...
-		}
-	}
-`, name)
-}
-
-func TestAccFabricReadServiceProfileByUuid(t *testing.T) {
+func TestAccFabricCreateServiceProfile_SP_FCR_ba(t *testing.T) {
+	spName1 := "fabric_tf_acc_test_01"
+	spName2 := "fabric_tf_acc_test_02"
+	typel2 := "L2_PROFILE"
```
thogarty, line 23: These constants can be placed directly in the config string, they don't need to be setup here.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,134 @@ import (
...
+	portType := "XF_PORT"
```
thogarty, line 24: Same about constants.

equinix/resource_fabric_service_profile_acc_test.go
thogarty: The spName's don't need to have they're own variable string. You could put them directly in the method signature and test assertions.

thogarty (changes requested):
Good start, have requested changes to make. If you have questions add your comments to the PR for visibility.

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/data_source_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/data_source_fabric_service_profile_acc_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+data "equinix_fabric_service_profile" "test" {
+		uuid = equinix_fabric_service_profile.test.uuid
+}`, spName, portUUID, portType, portMetroCode)
+}
+
+func TestAccFabricReadServiceProfileByUuid_PFCR(t *testing.T) {
+	ports := GetFabricEnvPorts(t)
+
+	var port v4.Port
```
thogarty, line 71: We shouldn't use structs for the values that will be passed inside of `resource.ParallelTest`. They need to be primitives. If we use more than one layer deep of the `v4.Port` struct type, and the ENV variable is not set then it will fail the test run when TF_ACC is not set. 

There's a GHA workflow that runs non-terraform tests and it will cause this workflow to fail.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,135 @@ import (
...
+	ports := GetFabricEnvPorts(t)
 
-func testAccFabricReadServiceProfilesListConfig(name string) string {
-	return fmt.Sprintf(`data "equinix_fabric_service_profiles" "test" {
-		filter {
-			property = "/name"
-			operator = "="
-			values = ["%s"]
-		}
+	var portTcFirst, portTcSecond v4.Port
```
thogarty, line 23: Same as first comment:

We shouldn't use structs for the values that will be passed inside of resource.ParallelTest. They need to be primitives. If we use more than one layer deep of the v4.Port struct type, and the ENV variable is not set then it will fail the test run when TF_ACC is not set.

There's a GHA workflow that runs non-terraform tests and it will cause this workflow to fail.

Please use the primitive type matching whichever subcategory of `v4.Port` you'll be using.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,135 @@ import (
...
-		PreCheck:  func() { acceptance.TestAccPreCheck(t) },
-		Providers: acceptance.TestAccProviders,
-		Steps: []resource.TestStep{
-			{
-				Config: testAccFabricReadServiceProfilesListConfig("Azure ExpressRoute"),
-				Check: resource.ComposeTestCheckFunc(
+						"equinix_fabric_service_profile.test", "type", fmt.Sprint("L2_PROFILE")),
 					resource.TestCheckResourceAttr(
-						"data.equinix_fabric_service_profiles.test", "data.#", fmt.Sprint(1)), // Check  total number of ServiceProfile list returned in the Response payloads
+						"equinix_fabric_service_profile.test", "state", fmt.Sprint("ACTIVE")),
```
thogarty, line 42: There's no need to use `fmt.Sprint()` for your strings. You can just use "" for your string values.

thogarty: This will need to be updated for all `fmt.Sprint()`'s that just have a simple string primitive inside them.

thogarty (changes requested):
More changes requested. Let me know if you have questions.

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,138 @@ import (
...
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:  func() { acceptance.TestAccPreCheck(t) },
-		Providers: acceptance.TestAccProviders,
+		PreCheck:     func() { acceptance.TestAccPreCheck(t) },
+		Providers:    acceptance.TestAccProviders,
+		CheckDestroy: checkServiceProfileDelete,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccFabricReadServiceProfileConfig("bfb74121-7e2c-4f74-99b3-69cdafb03b41"),
+				Config: testAccFabricCreateServiceProfileConfig(portUuidDot1Q, "XF_PORT", portMetroCodeDot1Q),
```
thogarty, line 38: You should get your port type directly from the Port ENV variable. The point is so that we don't have to update the strings here if we have to update the Port ENV variable. Add another variable to the list above and use it here.

equinix/data_source_fabric_service_profile_acc_test.go
```diff
@@ -0,0 +1,113 @@
...
+		portMetroCode = ports["pfcr"]["dot1q"][0].Location.MetroCode
+	}
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.TestAccPreCheck(t) },
+		Providers:    acceptance.TestAccProviders,
+		CheckDestroy: checkServiceProfileDelete,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFabricReadServiceProfileConfig("ds_con_sp_PFCR", portUuid, "XF_PORT", portMetroCode),
```
thogarty, line 82: You should get your port type directly from the Port ENV variable. The point is so that we don't have to update the strings here if we have to update the Port ENV variable. Add another variable to the list above and use it here.

thogarty (changes requested):
Should be the last set of changes requested. We want the port details to all be pulled from the ENV variable. Using the "XF_PORT" string without pulling it from the env variable will cause issues if/when we have to change port ENV variable. Do the same thing we do with other data elements in the Port struct so that it's only passing primitives into the `resource.ParallelTest`.

equinix/resource_fabric_service_profile_acc_test.go
tutkat: done

equinix/data_source_fabric_service_profile_acc_test.go
tutkat: done

equinix/data_source_fabric_service_profile_acc_test.go
```diff
@@ -0,0 +1,114 @@
...
+}`, spName, portUUID, portType, portMetroCode)
+}
+
+func TestAccFabricReadServiceProfileByUuid_PFCR(t *testing.T) {
+	ports := GetFabricEnvPorts(t)
+
+	var portUuid, portMetroCode string
+	if len(ports) > 0 {
+		portUuid = ports["pfcr"]["dot1q"][0].Uuid
+		portType = ports["pfcr"]["dot1q"][0].type
```
thogarty, line 73: The variable was never defined before assignment, and this will fail because `.type` is not part of the v4.Port struct. Please update.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,140 @@ import (
...
-		filter {
-			property = "/name"
-			operator = "="
-			values = ["%s"]
-		}
+	var portUuidDot1Q, portMetroCodeDot1Q string
+	var portUuidQinq, portMetroCodeQinq string
+	if len(ports) > 0 {
+		portUuidDot1Q = ports["pfcr"]["dot1q"][0].Uuid
+		portTypeDot1Q = ports["pfcr"]["dot1q"][0].type
```
thogarty, line 27: The variable is not declared before assignment, and again, `.type` is not a part of the v4.Port struct. Please update.

thogarty (changes requested):
Changes requested. Obvious bugs in PR. Please test after making changes to ensure correctness.

Additionally:
1. Please add a description to the PR so that it's clear what the change is without reading the code
2. Please update Names given to SPs. `ds_con_sp_PFCR` is unclear. Not sure what `ds` or `con` are referring two in either test sample.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,143 @@ import (
...
+		Providers:    acceptance.TestAccProviders,
+		CheckDestroy: checkServiceProfileDelete,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccFabricReadServiceProfileConfig("bfb74121-7e2c-4f74-99b3-69cdafb03b41"),
+				Config: testAccFabricCreateServiceProfileConfig(portUuidDot1Q, portTypeDot1Q, portMetroCodeDot1Q),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(
-						"data.equinix_fabric_service_profile.test", "name", fmt.Sprint("Azure ExpressRoute")),
+						"equinix_fabric_service_profile.test", "name", "fc_sp_PFCR""),
```
thogarty, line 46: Added a bug here in last commit. Unterminated string.

equinix/resource_fabric_service_profile_acc_test.go
```diff
@@ -17,139 +17,143 @@ import (
...
 			},
 		},
 	})
 }
 
-func testAccFabricCreateServiceProfileConfig(name string) string {
+func testAccFabricCreateServiceProfileConfig(portUUID string, portType string, portMetroCode string) string {
 	return fmt.Sprintf(`resource "equinix_fabric_service_profile" "test" {
-  name = "%s"
+  name = "fc_sp_PFCR""
```
thogarty, line 111: Added a bug here in last commit. Unterminated string.

thogarty (changes requested):
Changes requested. Please ensure you build and run tests after making changes so that your code is ready for a merge.

thogarty (approved):

thogarty:
Can confirm that service profile tests are working locally. Fabric GHA tests need some debugging and user data updates as well as sweeper changes before they will start passing again.

Approving change and merging.

PR #562, opened 2024-02-06
refactor: project sshkeys using equinix-sdk-go

codinja1188:
https://github.com/equinix/terraform-provider-equinix/issues/402


internal/errors/errors.go
```diff
@@ -194,6 +194,27 @@ func IgnoreResponseErrors(ignore ...func(resp *http.Response, err error) bool) f
 	}
 }
 
+func IgnoreMetalResponseErrors(ignore ...func(resp *http.Response, err error) bool) func(resp *http.Response, err error) error {
+	return func(resp *http.Response, err error) error {
+		var r *http.Response
+		if resp != nil {
+			r = resp
+		}
+		mute := false
```
ctreatma, line 203: This duplicates the `IgnoreHttpResponseErrors` function further down in this file

internal/resources/metal/project_ssh_key/datasource.go
```diff
@@ -44,18 +44,14 @@ func (r *DataSource) Read(
 	id := data.ID.ValueString()
 	search := data.Search.ValueString()
 	projectID := data.ProjectID.ValueString()
+	inc := []string{}
```
ctreatma, line 47: We don't need to add this or the `Include` call if we're not including anything

internal/resources/metal/project_ssh_key/models.go
```diff
@@ -19,14 +19,14 @@ type ResourceModel struct {
...
-	m.Updated = types.StringValue(key.Updated)
-	m.OwnerID = types.StringValue(path.Base(key.Owner.Href))
+func (m *ResourceModel) parse(key *metalv1.SSHKey) diag.Diagnostics {
+	m.ID = types.StringValue(key.GetId())
+	m.Name = types.StringValue(key.GetLabel())
+	m.PublicKey = types.StringValue(key.GetKey())
+	m.Fingerprint = types.StringValue(key.GetFingerprint())
+	m.Created = types.StringValue(key.CreatedAt.GoString())
+	m.Updated = types.StringValue(key.UpdatedAt.GoString())
+	m.OwnerID = types.StringValue(path.Base(key.GetHref()))
```
ctreatma, line 29: This is a material change to the logic; the old code was getting the owner ID, but this new code is getting the ID of the SSH key itself.

internal/resources/metal/project_ssh_key/models.go
```diff
@@ -46,14 +46,14 @@ type DataSourceModel struct {
...
-	m.Updated = types.StringValue(key.Updated)
-	m.OwnerID = types.StringValue(path.Base(key.Owner.Href))
+func (m *DataSourceModel) parse(key *metalv1.SSHKey) diag.Diagnostics {
+	m.ID = types.StringValue(key.GetId())
+	m.Name = types.StringValue(key.GetLabel())
+	m.PublicKey = types.StringValue(key.GetKey())
+	m.Fingerprint = types.StringValue(key.GetFingerprint())
+	m.Created = types.StringValue(key.CreatedAt.GoString())
+	m.Updated = types.StringValue(key.UpdatedAt.GoString())
+	m.OwnerID = types.StringValue(path.Base(key.GetHref()))
```
ctreatma, line 56: Some comment here as above: the ID on this line should come from the key owner, not the key itself

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -42,14 +42,13 @@ func (r *Resource) Create(
 	}
 
 	// Generate API request body from plan
-	createRequest := &packngo.SSHKeyCreateRequest{
-		Label:     plan.Name.ValueString(),
-		Key:       plan.PublicKey.ValueString(),
-		ProjectID: plan.ProjectID.ValueString(),
+	createRequest := &metalv1.SSHKeyCreateInput{
+		Label: plan.Name.ValueStringPointer(),
+		Key:   plan.PublicKey.ValueStringPointer(),
```
ctreatma, line 47: I forget how the API & SDK are structured around this, but if the project ID is not an input parameter for creating a project SSH key, then we need to be passing in the project ID somewhere else (possibly using a different operation than `SSHKeysApi.CreateSSHKey`).

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -123,7 +122,7 @@ func (r *Resource) Update(
 	resp *resource.UpdateResponse,
 ) {
 	r.Meta.AddFwModuleToMetalUserAgent(ctx, req.ProviderMeta)
```
ctreatma, line 124: All these `AddFwModuleToMetalUserAgent` also need to be changed to add the module name to the equinix-sdk-go User-Agent instead

internal/resources/metal/project_ssh_key/models.go
ocobles: It seems `Owner` field is no longer available in metal-go / API specs but I can see it in the API response. I'm wondering if `Entity` is a new replacement for owner, but still, entity is not in the API response and owner is.

ctreatma: It looks like the spec and the API have been mismatched for years on that attribute; the spec should say owner instead of entity.  Options are to (1) open a PR on [equinix-sdk-go](https://github.com/equinix/equinix-sdk-go) that patches the spec to rename entity to owner in the SSHKey schema or (2) use `AdditionalProperties` to work around the spec defect:

```go
// Types are approximate, this might not work
owner = key.AdditionalProperties["owner"].(map[string]interface{})
owner_href = owner["href"].(string)
m.OwnerID = types.StringValue(path.Base(owner_href))
```

internal/acceptance/ssh_key_helpers.go
```diff
@@ -34,3 +36,29 @@ func TestAccCheckMetalSSHKeyExists(n string, key *packngo.SSHKey) resource.TestC
 		return nil
 	}
 }
+
+func TestAccCheckMetalGOSSHKeyExists(n string, key *metalv1.SSHKey) resource.TestCheckFunc {
```
ctreatma, line 40: Rather than add a new function here, lets update the existing one to use equinix-sdk-go.  These check functions are only used in tests, so we can change them a bit more freely.

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -184,8 +186,8 @@ func (r *Resource) Delete(
 	id := state.ID.ValueString()
 
 	// Use API client to delete the resource
-	deleteResp, err := client.SSHKeys.Delete(id)
-	if equinix_errors.IgnoreResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(deleteResp, err) != nil {
+	deleteResp, err := client.SSHKeysApi.DeleteSSHKey(context.Background(), id).Execute()
+	if equinix_errors.IgnoreMetalResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(deleteResp, err) != nil {
```
ctreatma, line 190: This (and other references to the new function) should use the pre-existing `equinix_errors.IgnoreHttpResponseErrors` function

internal/resources/metal/project_ssh_key/resource.go
ctreatma: Looked into the SDK, and for a project SSH key we need to use [`CreateProjectSSHKey`](https://github.com/equinix/equinix-sdk-go/blob/main/services/metalv1/api_ssh_keys.go#L59)

internal/resources/metal/project_ssh_key/models.go
displague: We're going to need to be careful about accepting minor updates from equinix-sdk-go where new attributes are introduced because of the silent breaking change side-effect of `AdditionalProperties` keys being removed.

To be defensive about that, pushing an error rather than silently failing, we could add more checking:

```go
owner, exists := key.AdditionalProperties["owner"]
if !exists {
    return fmt.Errorf("owner not found")
}

ownerMap, ok := owner.(map[string]interface{})
if !ok {
    return fmt.Errorf("owner is not a map[string]interface{}")
}

ownerHref, exists := ownerMap["href"].(string)
if !exists {
    return fmt.Errorf("href not found or invalid type")
}

m.OwnerID = types.StringValue(path.Base(ownerHref))
```

If getting the ID from an AdditionalProperties map href is common enough, we could make a helper for it.

ctreatma: I suspect this would surface in [the same way as it did on the other PR](https://github.com/equinix/terraform-provider-equinix/pull/538#discussion_r1466944138):  an SDK upgrade would cause a panic that would be caught during testing.

IMO patching the SDK is preferable to adding a helper here.  This code is tech debt to work around a short-term issue.  Once we're investing more than a one-off workaround we should do it in a way that benefits all SDK users.

internal/resources/metal/project_ssh_key/resource.go
```diff
@@ -87,7 +88,7 @@ func (r *Resource) Read(
 	id := state.ID.ValueString()
 
 	// Use API client to get the current state of the resource
-	key, _, err := client.SSHKeys.Get(id, nil)
+	key, _, err := client.SSHKeysApi.FindSSHKeyById(context.Background(), id).Include(nil).Execute()
```
ctreatma, line 91: We can leave out the `.Include` call entirely if there's nothing to include:

```suggestion
	key, _, err := client.SSHKeysApi.FindSSHKeyById(context.Background(), id).Execute()
```

internal/resources/metal/project_ssh_key/datasource.go
```diff
@@ -65,21 +60,18 @@ func (r *DataSource) Read(
...
 	}
 
+	keys := keysList.GetSshKeys()
 	for i := range keys {
-		// use the first match for searches
-		if search != "" {
-			key = keys[i]
-			break
-		}
+		key = keys[i]
```
ctreatma, line 65: This is a major change in logic:  the old code took the first match, but this one will take the last.  The old code should remain in place.

internal/resources/metal/helper/helper.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+func GetMetalSSHKeyOwnerId(key *metalv1.SSHKey) basetypes.StringValue {
+	owner, exists := key.AdditionalProperties["owner"]
+	if !exists {
+		fmt.Errorf("owner not found")
```
ctreatma, line 15: We're not returning any errors from this function, so as currently written this helper will end up introducing the problem @displague is referring to rather than fixing it: rather than failing when the SDK changes, we will quietly change the organization ID from a UUID to an empty string.

We _could_ improve the helper by returning these errors and then checking for them in the caller, but IMO that's overkill and we should just read the owner value from AdditionalProperties and cast it to `map[string]interface{}` without checking for nil.  As mentioned in the other thread, when an attribute disappears from AdditionalProperties, blindly casting it will cause a panic.  That will be caught in the acceptance tests without the need for a helper, and the ugly code will motivate us to fix the issue rather than maintaining a helper after it's no longer necessary.

Issue #403, opened 2023-10-05
 crash while destroying two equinix_ecx_l2_connection


ocobles:
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.
╵
╷
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-equinix_v1.17.0 plugin:

fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 41 [running]:
net/textproto.MIMEHeader.Set(...)
        net/textproto/header.go:22
net/http.Header.Set(...)
        net/http/header.go:40
github.com/go-resty/resty/v2.(*Client).SetHeader(...)
        github.com/go-resty/resty/v2@v2.3.0/client.go:165
github.com/equinix/terraform-provider-equinix/equinix.(*Config).addModuleToECXUserAgent(0xc000078900?, 0xc0007ced28, 0x0?)
        github.com/equinix/terraform-provider-equinix/equinix/config.go:287 +0xee
github.com/equinix/terraform-provider-equinix/equinix.resourceECXL2ConnectionDelete({0x1f00ae8, 0xc00074f0b0}, 0xc000174480, {0x1ce65e0?, 0xc000771200?})
        github.com/equinix/terraform-provider-equinix/equinix/resource_ecx_l2_connection.go:770 +0x99
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).delete(0xc00044f960, {0x1f00b20, 0xc000706e70}, 0xd?, {0x1ce65e0, 0xc000771200})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:758 +0x12e
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0xc00044f960, {0x1f00b20, 0xc000706e70}, 0xc000121040, 0xc000174400, {0x1ce65e0, 0xc000771200})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:806 +0x605
[...]

Issue #286, opened 2023-01-13
equinix_network_bgp resource is not working


milko36:
Hi guys,
I have a deployment code for deploy L2 connections, vif on Amazon and bgp in Equinix but I found a bug while bgp was creation on Equinix. All steps are working but the deploy is stopped here:
module.hbg_network_equinix_fabric_connection.equinix_network_bgp.bgp-virtual["ld-2-eu-central-1"]: Creating...
module.hbg_network_equinix_fabric_connection.equinix_network_bgp.bgp-virtual["ld-2-eu-west-1"]: Creating...

If I try again run a terraform apply, the resource equinix_network_bgp is created.

We've saw this log:

│ Error: Plugin did not respond
│ 
│   with module.hbg_network_equinix_fabric_connection.equinix_network_bgp.bgp-virtual["ld-2-eu-central-1"],
│   on ../../main.tf line 71, in resource "equinix_network_bgp" "bgp-virtual":
│   71: resource "equinix_network_bgp" "bgp-virtual" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.

This is all log in the plugin:

Stack trace from the terraform-provider-equinix_v1.11.1 plugin:

fatal error: concurrent map writes

goroutine 62 [running]:
runtime.throw({0xe198ff?, 0x10?})
        runtime/panic.go:992 +0x71 fp=0xc0002d8c50 sp=0xc0002d8c20 pc=0x434231
runtime.mapassign_faststr(0xd900a0, 0xc0009947b0, {0xe0f569, 0xa})
        runtime/map_faststr.go:295 +0x38b fp=0xc0002d8cb8 sp=0xc0002d8c50 pc=0x41334b
net/textproto.MIMEHeader.Set(...)
        net/textproto/header.go:22
net/http.Header.Set(...)
        net/http/header.go:40
github.com/go-resty/resty/v2.(*Client).SetHeader(...)
        github.com/go-resty/resty/v2@v2.3.0/client.go:165
github.com/equinix/terraform-provider-equinix/equinix.(*Config).addModuleToNEUserAgent(0x0?, 0xc0002d8e00, 0x4527ad?)
        github.com/equinix/terraform-provider-equinix/equinix/config.go:261 +0xe5 fp=0xc0002d8d20 sp=0xc0002d8cb8 pc=0xb81b25
[...]

PR #563, opened 2024-02-07
refactor: move equinix-sdk-go Metal client setup into the config package

ctreatma:
When we need to use a Metal API client inside a terraform resource or datasource, we need to ensure that the client is instrumented correctly.  Previously this was done by repeating the same function calls in every CRUD function for every resource and datasource; as a result we were often missing instrumentation, and there was a high risk of losing instrumentation when converting from the old packngo SDK to equinix-sdk-go.

This refactors the equinix-sdk-go client setup code to better ensure that we get a correctly-configured client in all situations and help ensure that we maintain visibility into usage after doing SDK conversions.


internal/config/config.go
```diff
@@ -218,13 +216,41 @@ func (c *Config) NewMetalClient() *packngo.Client {
...
-	client.UserAgent = c.fullUserAgent(client.UserAgent)
+	client.UserAgent = c.tfSdkUserAgent(client.UserAgent)
 	c.metalUserAgent = client.UserAgent
 	return client
 }
 
-// NewMetalGoClient returns a new metal-go client for accessing Equinix Metal's API.
-func (c *Config) NewMetalGoClient() *metalv1.APIClient {
+func (c *Config) NewMetalClientForSDK(d *schema.ResourceData) *metalv1.APIClient {
+	client := c.newMetalClient()
```
ctreatma, line 225: Note this means we're setting up a new client every time, which should avoid coordination issues for User-Agent modifications.  If we're OK with creating a bunch of separate clients, we could also use this pattern for other clients and avoid the occasional `concurrent map writes` panics we've seen (#286, #403).

internal/config/config.go
```diff
@@ -163,29 +161,28 @@ func (c *Config) Load(ctx context.Context) error {
...
 	return nil
 }
 
 // NewFabricClient returns a new client for accessing Equinix Fabric's v4 API.
 // uncomment the funct when migrating Fabric resources to use
 // functions from internal/
-func (c *Config) NewFabricClient() *v4.APIClient {
-	transport := logging.NewTransport("Equinix Fabric", http.DefaultTransport)
-	authClient := &http.Client{
+func (c *Config) NewFabricClient(authClient *http.Client) *v4.APIClient {
```
ctreatma, line 183: Whoops; this leaked in from a different side quest

internal/config/config.go
cprivitere: I don't. understand what the downside would be.

cprivitere (approved):
With the exception of my comment above, this all seems reasonable to me. I would still like to see @displague or @ocobles sign off on it too.

internal/config/config.go
ctreatma: The leak has been fixed; this change will come back in a later episode.

internal/config/config.go
ctreatma: I'm not totally clear on the overhead of creating these separate clients.  There's the potential for increased resource utilization, but we're still reusing the `http.DefaultTransport` across all clients so I _think_ any increase should be negligible.

displague (approved):

PR #567, opened 2024-02-14
refactor: Move vlans to separate package and adopt equinix-sdk-go

aayushrangwala:
relates: #402


ctreatma:
This PR should move the resource and datasource to terraform-plugin-framework, keeping packngo in place for now.  Switching the SDK duplicates #541 and is guaranteed to run into spec/SDK deficiencies; migrating to framework without changing the SDK ensures that the necessary changes are all contained in this repo.

PR #568, opened 2024-02-14
refactor: migrate resource metal_organization to framework

codinja1188:


internal/resources/metal/organizations/datasource.go
```diff
@@ -0,0 +1,86 @@
+package organizations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
```
ctreatma, line 7: We should keep packngo in place and only migrate the resource from SDK to framework.  Changing the SDK will probably be too much change for one PR.

ctreatma:
For the test files, rather than start from scratch you should move the existing test files for the organization resource and ata source into the new directory.  The expectation of these migrations is that the existing tests still pass after the migration with no changes (except for maybe changing which provider is passed in to the test cases).

ctreatma:
Most recent CI has a couple failures that are relevant to these changes:

ctreatma:
Most recent CI has a couple failures that are relevant to these changes:

```
2024/02/20 15:19:55 [DEBUG] GET https://api.equinix.com/metal/v1//devices/9c51d627-4761-4193-9350-f868612c2824?include=project%2Cmetro%2Cfacility%2Chardware_reservation
    resource_metal_organization_member_acc_test.go:55: Step 1/3 error: Error running pre-apply plan: exit status 1
        
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 16, in resource "equinix_metal_organization" "test":
          16: resource "equinix_metal_organization" "test" {
        
        The provider hashicorp/equinix does not support resource type
        "equinix_metal_organization".
--- FAIL: TestAccResourceMetalOrganizationMember_basic (0.47s)
...
=== NAME  TestAccDataSourceMetalGateway_privateIPv4
    datasource_test.go:13: Step 1/1 error: Error running pre-apply plan: exit status 1
        
        Error: failed to read provider configuration schema for registry.terraform.io/hashicorp/equinix: failed to retrieve schema from provider "registry.terraform.io/hashicorp/equinix": Error converting data source schema: The schema for the data source "equinix_metal_organizations" couldn't be converted into a usable type. This is always a problem with the provider. Please report the following to the provider developer:
        
        AttributeName("project_ids"): protocol version 5 cannot have Attributes set
        
--- FAIL: TestAccDataSourceMetalGateway_privateIPv4 (0.44s)
```

It looks like there are many examples of the second failure, and there may be other examples of the first failure.  There were (are?) similar failures for #568 so it looks like the way we're converting nested blocks isn't quite right.

internal/resources/metal/helper/regex.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+)
+
+func StringToRegex(pattern string) (regExp *regexp.Regexp) {
```
ctreatma, line 11: I don't think we should write wrapper functions for this.  The resources should continue to define validation directly, like `stringvalidator.RegexMatches(regexp.MustCompile(<pattern>), <message>)` rather than using a wrapper.

internal/resources/metal/organization_member/models.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+
+	"github.com/packethost/packngo"
+)
+
+type ResourceModel struct {
+	invitee         types.String `tfsdk:"invitee"`
```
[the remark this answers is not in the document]
ctreatma: OK.  If it isn't necessary to move them both in one PR, they should not be moved in the same PR.  Can you undo these changes?

ctreatma:
I updated all Metal acceptance tests in #586 so that we don't have to keep updating how the providers are wired up in tests.  Now that that is merged, you will need to rebase this PR on `main` and resolve conflicts.

internal/resources/metal/organization/datasource_schema.go
```diff
@@ -0,0 +1,84 @@
...
+		"project_ids": schema.ListNestedAttribute{
+			Computed: true,
+			NestedObject: schema.NestedAttributeObject{
+				Attributes: map[string]schema.Attribute{
+					"id": schema.StringAttribute{
+						Computed: true,
+					},
+				},
+			},
+		},
```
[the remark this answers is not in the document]
ocobles: This is a different case because the element type is an object. First you need to change the schema. You are using the resource `address` schema instead of the datasource schema. All the fields were computed https://github.com/equinix/terraform-provider-equinix/blob/450917b0e9018ebf302231a5fd1606b1889e3a78/equinix/data_source_metal_organization.go#L72-L96

You will usually need to define a `ListNestedBlock` but it is different for computed blocks (all fields in the nested object are computed). Use this as an example: https://github.com/equinix/terraform-provider-equinix/blob/450917b0e9018ebf302231a5fd1606b1889e3a78/internal/resources/metal/connection/datasource_schema.go#L94-L99

for `address` in the resource schema you will need to define a `ListNestedBlock`  in the blocks section, because there you have some editable fields (required/optional). It would be something like:

```
var dataSourceSchema = schema.Schema{
    Attributes: map[string]schema.Attribute{
        // YOUR ATTRIBUTES
    },
    Blocks: map[string]schema.Block{
        "address": schema.ListNestedBlock{
             Description: "Address information block",
             CustomType:  fwtypes.NewListNestedObjectTypeOf[PortModel](ctx),
             NestedObject: schema.NestedBlockObject{
                 Attributes: map[string]schema.Attribute{
                     "address": schema.StringAttribute{ ... },
                     "city": schema.StringAttribute{ ... },
                  },
	},
```

internal/resources/metal/organization/datasource.go
```diff
@@ -0,0 +1,79 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Extract the ID of the resource from the state
+	id := data.OrganizationID.ValueString()
+
+	var org *packngo.Organization
+
+	// orgs, _, err := client.Organizations.List(&packngo.GetOptions{Includes: []string{"address"}})
```
ctreatma, line 55: Continuing thread from https://github.com/equinix/terraform-provider-equinix/pull/568#discussion_r1499500723:

This is still not equivalent to the SDKv2 datasource.  The datasource should accept _either_ a name or an ID.  If it is given an ID, it should run a `Get`.  If it is given a name, it should run a `List` and then look through the list for an organization with a matching name.

internal/resources/metal/organization/resource_test.go
```diff
@@ -97,6 +63,49 @@ func TestAccMetalOrganization_create(t *testing.T) {
 	})
 }
 
+// Test to verify that switching from SDKv2 to the Framework has not affected provider's behavior
+func TestAccMetalGateway_upgradeFromVersion(t *testing.T) {
```
ctreatma, line 67: This test name is incorrect; it should not mention gateway.

internal/resources/metal/organization/resource_test.go
```diff
@@ -1,62 +1,28 @@
...
 
-func init() {
-	resource.AddTestSweepers("equinix_metal_organization", &resource.Sweeper{
-		Name:         "equinix_metal_organization",
-		Dependencies: []string{"equinix_metal_project"},
-		F:            testSweepOrganizations,
-	})
-}
-
-func testSweepOrganizations(region string) error {
```
ctreatma, line 25: The sweeper should not be deleted.  If #593 is merged before this one the sweeper should be updated to match the changes made there.  In the meantime the sweeper should stay as-is.

internal/resources/metal/organization/datasource.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Extract the ID of the resource from the state
+	orgId := data.OrganizationID
+	name := data.Name
+
+	if !orgId.IsNull() || !name.IsNull() {
+		resp.Diagnostics.AddError(
+			"You must provide either orgId OR Name", "",
+		)
+	}
```
ctreatma, line 58: I think we could replace these lines with an update to `datasource_schema.go` to specify this for the `organization_id` attribute:
```
Validators: []validator.String{
  stringvalidator.ExactlyOneOf(path.MatchRoot("organization_id"), path.MatchRoot("name")),
}
```

internal/resources/metal/organization/resource_test.go
```diff
@@ -98,8 +81,68 @@ func TestAccMetalOrganization_importBasic(t *testing.T) {
 	})
 }
 
+// Test to verify that switching from SDKv2 to the Framework has not affected provider's behavior
+func TestAccMetalOrganization_upgradeFromVersion(t *testing.T) {
```
ctreatma, line 85: This test is failing; it looks like there are some differences between the SDK and framework implementations:
```
=== NAME  TestAccMetalOrganization_upgradeFromVersion
    resource_test.go:88: Step 2/2 error: Pre-apply plan check(s) failed:
        expected empty plan, but equinix_metal_organization.test has planned action(s): [update]
```

internal/resources/metal/organization/resource_test.go
```diff
@@ -1,14 +1,16 @@
-package equinix
+package organization_test
```
ctreatma, line 1: There's a test failure in the project tests as well that looks related to these changes:
```
=== NAME  TestAccMetalProject_organization
    resource_test.go:361: Step 1/2 error: Error running apply: exit status 1
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to equinix_metal_organization.test, provider
        "provider[\"registry.terraform.io/hashicorp/equinix\"]" produced an
        unexpected new value: .description: was null, but now cty.StringVal("").
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```
(Note: this shouldn't be resolved with changes to the project test; this failure is an indication that migrating from SDK to framework changed the resource behavior in some way, and that behavior change needs to be fixed.)

internal/resources/metal/project/resource_test.go
```diff
@@ -340,6 +340,7 @@ func testAccMetalProjectConfig_organization(r string) string {
 	return fmt.Sprintf(`
 resource "equinix_metal_organization" "test" {
 	name = "tfacc-project-%s"
+	description = "quux"
```
ctreatma, line 343: It doesn't look like description is required for organizations; why is this being added in the test?

internal/validation/strings.go
```diff
@@ -12,5 +14,15 @@ var (
 	StringIsPortDefinition = validation.StringMatch(
 		regexp.MustCompile("^(([0-9]+(,[0-9]+){0,9})|([0-9]+-[0-9]+)|(any))$"),
 		"port definition has to be: up to 10 comma sepparated numbers (22,23), range (20-23) or word 'any'")
-	StringIsSpeedBand = validation.StringMatch(regexp.MustCompile("^[0-9]+(MB|GB)$"), "SpeedBand should consist of digit followed by MB or GB")
+	StringIsSpeedBand   = validation.StringMatch(regexp.MustCompile("^[0-9]+(MB|GB)$"), "SpeedBand should consist of digit followed by MB or GB")
+	StringIsCountryCode = stringvalidator.RegexMatches(StringToRegex("(?i)^[a-z]{2}$"), "Address country must be a two letter code (ISO 3166-1 alpha-2)")
```
ctreatma, line 18: Would `validation.Stringmatch(regexp.MustCompile(...` work here as it does on the line above?  I'm not clear what we gain by using `stringvalidator.RegexMatches`, especially if that requires adding a helper function that silently swallows a potential error.

internal/resources/metal/project/resource_test.go
ctreatma: Was this done to address this comment? https://github.com/equinix/terraform-provider-equinix/pull/568/files#r1507920057

ctreatma: If so, as mentioned in that comment, that test failure shouldn't be fixed by changing this test; it should be fixed by making the organization framework resource handle the `description` field the same way the SDKv2 resource handled it.

internal/validation/strings.go
ctreatma: Note this still needs to be addressed.  A direct call to `regexp.MustCompile` will panic if the regex is invalid.  The `StringToRegex` helper will log an error and return nil and it's not clear what would happen in that case.  We should remove the helper.

internal/resources/metal/organization/resource_test.go
ctreatma: This test is still failing in the latest CI, so it looks like there is still a difference between how the SDKv2 resource stored organizations in state and how the framework resource stores organizations in state.  I think this is a test that can be run locally against the playground org.  I'm not sure how to get more information out of the test, though.  @ocobles do you know if there's a way to get terraform acceptance tests to print the plan for each step?

internal/resources/metal/organization/resource_schema.go
```diff
@@ -0,0 +1,87 @@
...
+							Required:    true,
+							Validators: []validator.String{
+								stringvalidator.LengthBetween(0, 18),
+								equinix_validation.StringIsCountryCode,
+							},
+						},
+						"state": schema.StringAttribute{
+							Description: "State name",
+							Optional:    true,
+							Computed:    true,
```
ctreatma, line 80: It's possible the `_upgradeFromVersion` test failure is related to this line.  The SDKv2 resource only had `Optional: true` for `address.state`.  The addition of `Computed: true` means that terraform will mark the value as unknown during plan if it isn't explicitly configured.  I tried adding [a `UseStateForUnknown` plan modifier](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown) but that didn't seem to do anything.  Removing `Computed: true` was also not enough to resolve the issue (that may be because the ResourceModel handles null/empty `address.state` differently than the SDKv2 resource handled it.

ctreatma: Oh and also had to add `stringplanmodifier.UseStateForUnknown` on the `created` and `updated` properties in the resource schema

ctreatma: I pushed up a branch with my changes for comparison: 
Tests for that branch are running here: https://github.com/equinix/terraform-provider-equinix/actions/runs/8179319403

In summary:
- Removed `Computed: true` from `address.state` and updated `parseAddress` to skip setting `address.state` if it's nil or empty, as was done in SDKv2 implementation
- Marked description, website, twitter, and logo fields as `Computed: true` with a default value of `""`, which seems to match how the SDKv2 resource was represented in state

ctreatma: The only test failure in CI on my branch is `TestAccMetalIPAttachment_metro`, which failed due to lack of server capacity.  How did you initially incorporate my changes into your local repo and which tests were failing locally?

ctreatma: The latest commit on this PR has a bunch of failures like this:
```
    resource_test.go:67: Step 1/2 error: Error running apply: exit status 1
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to equinix_metal_organization.test, provider
        "provider[\"registry.terraform.io/hashicorp/equinix\"]" produced an
        unexpected new value: .address[0].city: was cty.StringVal("London"), but now
        null.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to equinix_metal_organization.test, provider
        "provider[\"registry.terraform.io/hashicorp/equinix\"]" produced an
        unexpected new value: .address[0].country: was cty.StringVal("GB"), but now
        null.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to equinix_metal_organization.test, provider
        "provider[\"registry.terraform.io/hashicorp/equinix\"]" produced an
        unexpected new value: .address[0].zip_code: was cty.StringVal("12345"), but
        now null.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to equinix_metal_organization.test, provider
        "provider[\"registry.terraform.io/hashicorp/equinix\"]" produced an
        unexpected new value: .address[0].address: was cty.StringVal("tfacc org
        street"), but now null.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```

I recommend reverting/removing the last commit on this PR and cherry-picking the last commit from the branch I linked in a previous comment on this thread.

ctreatma (approved):
LGTM.  All relevant tests passed; CI failures are all platform issues affecting tests that don't use metal_organization functionality.

PR #572, opened 2024-02-15
refactor: Framework migration metal connection resource and datasource

ocobles:
Migration of `equinix_metal_connection` and datasource `equinix_metal_connection‡`


ctreatma:
Can't tell where exactly it's coming from, but I see a large number of test failures in CI, and I see this error message as well:

```
    resource_metal_device_acc_test.go:225: Step 1/1 error: Error running pre-apply plan: exit status 1
        
        Error: failed to read schema for equinix_metal_project.test in registry.terraform.io/hashicorp/equinix: failed to retrieve schema from provider "registry.terraform.io/hashicorp/equinix": Error converting data source schema: The schema for the data source "equinix_metal_connection" couldn't be converted into a usable type. This is always a problem with the provider. Please report the following to the provider developer:
        
        AttributeName("service_tokens"): protocol version 5 cannot have Attributes set
```

In addition to the attribute mentioned in there, I also see references to `hashicorp/equinix`, but I don't know if that's an issue with our code or with the way terraform prints error messages.

internal/resources/metal/connection/datasource_schema.go
```diff
@@ -0,0 +1,158 @@
...
+		"status": schema.StringAttribute{
+			Description: "Status of the connection resource",
+			Computed:    true,
+		},
+		"token": schema.StringAttribute{
+			Description:        "Only used with shared connection. Fabric Token required to continue the setup process with [equinix_ecx_l2_connection](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_ecx_l2_connection) or from the [Equinix Fabric Portal](https://ecxfabric.equinix.com/dashboard)",
+			Computed:           true,
+			DeprecationMessage: "If your organization already has connection service tokens enabled, use `service_tokens` instead",
+		},
+		"service_tokens": schema.ListNestedAttribute{
```
ctreatma, line 90: It looks like `ListNestedAttributes` may be a protocol version 6 thing, and we're currently using protocol version 5: https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed#framework

ctreatma: Grasping at straws a bit here, because other than that migration page and the error message we're getting, I can't find anything that explicitly says that this function only works for protocol v6.  I pushed up a duplicate branch that includes a fix for the virtual circuit acceptance tests and converts the provider to protocol v6 to see if that resolves the issue: https://github.com/equinix/terraform-provider-equinix/actions/runs/7935657886/job/21669301881

If it does, though, we'll have to think about how to proceed, since switching to protocol v6 means dropping support for terraform < 1.0.  (From what I can tell, the protocol v5 version of this change would mean losing descriptions for the nested attributes.)

internal/resources/metal/connection/resource.go
```diff
@@ -0,0 +1,480 @@
...
+			resp.Diagnostics.AddError(
+				"Error creating Metal Connection",
+				"Could not create Metal Connection: "+err.Error(),
+			)
+			return
+		}
+	}
+
+	// Parse API response into the Terraform state
+	resp.Diagnostics.Append(plan.parse(ctx, conn)...)
```
ctreatma, line 111: The SDK version of this resource called the read function at the end of the create function.  Framework discourages that, but we should probably either (1) replicate the behavior by making a `client.Connections.Get` call to ensure we have all the necessary data or (2) add `include` parameters to the two create requests, assuming the underlying API operations support include parameters.

internal/resources/metal/connection/resource.go
```diff
@@ -0,0 +1,498 @@
...
+
+	// Extract the ID of the resource from the state
+	id := plan.ID.ValueString()
+
+	// Prepare update request based on the changes
+	updateRequest := &packngo.ConnectionUpdateRequest{}
+	// TODO (ocobles) The legacy SDK code includes below code snippet which
+	// looks incorrect as "locked" is a device field. Delete it when we are sure it is not necessary
+	//
+	// if d.HasChange("locked") {
```
ctreatma, line 195: Yeah this cannot possibly have worked before.  I say torch it.

internal/resources/metal/connection/resource.go
```diff
@@ -0,0 +1,480 @@
...
+			resp.Diagnostics.AddError(
+				"Error creating Metal Connection",
+				"Could not create a shared Metal Connection: "+err.Error(),
+			)
+			return
+		}
+	}
+
+	// Dedicated connection specific logic
+	if plan.Type.ValueString() == string(packngo.ConnectionDedicated) {
```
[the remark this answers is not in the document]
ocobles: I took the opportunity to not assume that there are only two types of connection, so it was something more proactive than forced, although I know that if that is the case in the future, many more changes will have to be made

ocobles:
last test execution failed in https://github.com/equinix/terraform-provider-equinix/actions/runs/7997097147/job/21840875078?pr=572#step:6:3250 but it works locally 

image

And same with  https://github.com/equinix/terraform-provider-equinix/actions/runs/7997097147/job/21840875078?pr=572#step:6:3121
image

internal/resources/metal/connection/datasource_schema.go
ctreatma: Closing the loop here:  once everything was cut over to protocol v6 there were errors about types not supporting optional values, so probably the same problem that was actually happening here, except the error message with v6 was better.  In that case the only reason to upgrade would be to maintain our sanity, so:  no need to upgrade.

ctreatma (approved):
I can't claim to have gone through all 2000 lines of this change with a fine-toothed comb, but everything looks good to me and I see that the only test failures are for things that would not be impacted by these changes.

internal/resources/metal/connection/resource_test.go
```diff
@@ -359,3 +360,44 @@ func TestAccMetalConnection_sharedVlans(t *testing.T) {
...
+	rs := acctest.RandString(10)
+	cfg := testAccMetalConnectionConfig_Shared_zside(rs)
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.TestAccPreCheckMetal(t) },
+		CheckDestroy: testAccMetalConnectionCheckDestroyed,
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"equinix": {
+						VersionConstraint: "1.29.0", // latest version with resource defined on SDKv2
```
displague, line 376: Does this version number need to be bumped going forward?

ocobles: not necessarily, since it is the latest version available with the resource using sdkv2. In fact, we can remove this test once migrated, it helped validate compatibility

ocobles:
Lessons Learned and Changes Introduced:

- Migrating to protocol v5, computed blocks are now implemented using a ListAttribute with an ElementType of types.ObjectType. This results in the loss of underlying field properties, meaning we no longer have descriptions of object fields. https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks-computed
- When nested schemas contain editable fields, schema.ListNestedBlock should be used instead of schema.ListNestedAttribute, as the latter is only available for protocolv6.
- Utilizing custom types enables us to obtain the ObjectType of the model using reflection.
- Fields like "facility," which are optional, computed, and have the plan modifier "RequiresReplace," should use https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#usestateforunknown. Otherwise, when there is an update to any other field, this field would be marked as (known after apply), thus forcing a replace.
- With the absence of StateFunc, logic must be added in Parse model. This will occur in all "metro" resources where we convert the value to lowercase. Refer to https://github.com/equinix/terraform-provider-equinix/blob/03ca8870fb216c0c2b85577d57a7f58bc064a6c4/internal/resources/metal/connection/models.go#L120-L126 for guidance.

PR #573, opened 2024-02-15
test: equinix_fabric_routing_protocol resource and data source

thogarty:
* Added acceptance test for routing protocol resource
* Data source tests for routing protocol resource included in resource file because of test setup
* Test setup creates an FCR, and a Connection on that FCR
* Config applies the Direct and BGP routing protocols to the created FCR connection
* Data_Source config retrieves the RPs created in the config
* Test assertions are performed on both the resource and data sources in the same test.
* Go test timeout must be >15m for it to work successfully


tutkat (approved):

srushti-patl (approved):
LGTM

PR #574, opened 2024-02-15
refactor: move device resource and datasource to separate package

ctreatma:
Part of #106

### DIFF
--- a/equinix/resource_metal_spot_market_request.go
+++ b/equinix/resource_metal_spot_market_request.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"time"
@@ -20,6 +21,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/packethost/packngo"
+)
+
+var (
+	matchIPXEScript = regexp.MustCompile(`(?i)^#![i]?pxe`)
 )
 
 func resourceMetalSpotMarketRequest() *schema.Resource {
--- a/internal/resources/metal/device/list_datasource.go
+++ b/internal/resources/metal/device/list_datasource.go
@@ -1,4 +1,4 @@
-package equinix
+package device
 
 import (
 	"context"
@@ -11,8 +11,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func dataSourceMetalDevices() *schema.Resource {
-	dsmd := dataSourceMetalDevice()
+func ListDataSource() *schema.Resource {
+	dsmd := DataSource()
 	sch := dsmd.Schema
 	for _, v := range sch {
 		if v.Optional {
--- a/internal/resources/metal/device/resource.go
+++ b/internal/resources/metal/device/resource.go
@@ -1,4 +1,4 @@
-package equinix
+package device
 
 import (
 	"context"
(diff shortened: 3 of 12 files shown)


internal/resources/metal/device/list_datasource.go
```diff
@@ -11,8 +11,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func dataSourceMetalDevices() *schema.Resource {
-	dsmd := dataSourceMetalDevice()
+func ListDataSource() *schema.Resource {
+	dsmd := DataSource()
```
ctreatma, line 15: I opted to move the `metal_devices` data source into the same directory as the `metal_device` resource and datasource, because I think that's better than putting it in a separate `internal/resources/metal/devices` directory.

_If_ that is agreed to be a good approach, then that raises the question: should `metal_device_bgp_neighbors` _also_ go in this directory, or does it still go in a separate `internal/resources/metal/device_bgp_neighbors` directory?

equinix/resource_metal_spot_market_request.go
```diff
@@ -22,6 +23,10 @@ import (
 	"github.com/packethost/packngo"
 )
 
+var (
+	matchIPXEScript = regexp.MustCompile(`(?i)^#![i]?pxe`)
```
ctreatma, line 27: Chose to duplicate this matcher instead of finding a place to put it.  IMO we can clean that up after framework and equinix-sdk-go migrations are done.

internal/resources/metal/device/resource.go
```diff
@@ -40,15 +40,15 @@ var (
...
+func Resource() *schema.Resource {
 	return &schema.Resource{
 		Timeouts: &schema.ResourceTimeout{
 			Create: schema.DefaultTimeout(30 * time.Minute),
 			Update: schema.DefaultTimeout(30 * time.Minute),
 			Delete: schema.DefaultTimeout(30 * time.Minute),
 		},
 		CreateContext:      resourceMetalDeviceCreate,
-		ReadWithoutTimeout: resourceMetalDeviceRead,
+		ReadWithoutTimeout: Read,
```
displague, line 51: Why not ReadContext?

displague:
Should be rebased after #586

equinix/resource_metal_spot_market_request.go
displague: I don't think the validation this enforces is necessary in the Terraform provider. Perhaps https://github.com/equinix/terraform-provider-equinix/blob/21c9dcc8dcd6eff9075b8c29cdb5fa57f170befa/equinix/resource_metal_spot_market_request.go#L274-L277 would be better as a warning or left to the API to enforce.  It's well enough to leave this validation alone since there are no known edge cases where the combination of an IPXE URL and a Custom IPXE Userdata are necessary.

displague: Perhaps the regexp could be moved closer to where it is used since it is not used elsewhere.

internal/resources/metal/device/resource.go
ctreatma: `ReadWithoutTimeout` and `ReadContext` are both context-aware lifecycle methods: the first argument to the function specified for either of those hooks must be a `context.Context`.  The difference between the two hooks is that `ReadContext` enforces a 20-minute default timeout that can be overridden in a `timeout` block and `ReadWithoutTimeout` doesn't enforce any timeout.  When this is migrated to framework (probably in this PR at some point), we might be forced to revisit timeouts anyway since SDKv2 and framework model/implement them differently.

displague:
I'm curious what this looks like after a rebase (and again after #593 and perhaps other open general PRs); whether we've reached a point in refactoring where we can move all remaining metal resources to `internal/resources/metal/*` with little more than `git mv`, import list changes, and function exporting?

internal/resources/metal/device/list_datasource.go
displague: The pluralization of the resource names, as in multi-response datasources, was a subtle point that I missed on first glance of this comment.

AWS followed this pattern. Seems fine to me. 
https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ec2/ec2_instances_data_source.go

ctreatma:

Issue #471, opened 2023-11-30
[Bug]: network account query returned more than one result


mwardbopp:
### Terraform Version

1.6.5

### Equinix Provider Version

1.20

### Effected Terraform Resources

equinix_network_account

### Terraform Config Files

data "equinix_network_account" "ny" {
  metro_code = "NY"
  name = "Aviatrix"
}

### Debug Output

https://gist.github.com/mwardbopp/2feba86732ed77cd4c1208880b8a4689

### Expected Behavior

network account should be identified

### Actual Behavior

│ Error: network account query returned more than one result, please try more specific search criteria
│ 
│   with data.equinix_network_account.ny,
│   on equinix.tf line 79, in data "equinix_network_account" "ny":
│   79: data "equinix_network_account" "ny" {

### Steps to Reproduce

terraform plan

worked ~30 days ago, tested on provider <1.18.

unsure what the api response was previously but now responds with this below. Not sure if terraform provider needs to provide the ability to filter on projectID?

{
    "accountCreateUrl": null,
    "accounts": [
        {
            "accountName": "Aviatrix",
            "accountNumber": "111111",
            "accountId": "E-111111",
            "accountUcmId": "111111",
            "accountStatus": "Active",
            "portalOrgId": "111111",
            "portalOrgName": "Aviatrix - 111111",
            "siblingCustOrgFlag": false,
            "creditHold": false,
            "projectId": "7defed40",
            "global": false
        },
        {
            "accountName": "Aviatrix",
            "accountNumber": "111111",
            "accountId": "E-111111",
            "accountUcmId": "111111",
            "accountStatus": "Active",
            "portalOrgId": "111111",
            "portalOrgName": "Aviatrix - 111111",
            "siblingCustOrgFlag": false,
            "creditHold": false,
            "projectId": "a981b91e",
            "global": false
        },
        {
            "accountName": "Aviatrix",
            "accountNumber": "111111",
            "accountId": "E-111111",
            "accountUcmId": "9F851BB8-xxxx",
            "accountStatus": "Active",
            "portalOrgId": "111111",
            "portalOrgName": "Aviatrix - 111111",
            "siblingCustOrgFlag": false,
            "creditHold": false,
            "projectId": "e5459aac-xxxxxx",
            "global": false
        },
        {
            "accountName": "Aviatrix",
            "accountNumber": "111111",
            "accountId": "E-111111",
            "accountUcmId": "111111",
            "accountStatus": "Active",
[...]

PR #576, opened 2024-02-16
feat: Include project id in client side filter for get accounts data source

kpdhulipala:
feat: Include project id in client side filter for get accounts data source
PR in ne-go: https://github.com/equinix/ne-go/pull/29
Closes the issue : https://github.com/equinix/terraform-provider-equinix/issues/471


rling-equinix (approved):
Lgtm

displague (approved):
This is optional client-side filtering of the equinix_network_account results based on a matching project_id. PR looks to do just that.

The field is added to the datasource documentation.

The test lines added verify that the datasource results now includes the project_id field.

What is not being tested is the filtering by project_id. The test isn't used in E2E testing, so it's hard to say how many results we would get back in the results. The code appears to be correct and I would expect the filter to work.

PR #578, opened 2024-02-19
chore: equinix_metal_vlan plugin sdk v2 to framework migration

aayushrangwala:
Part of https://github.com/equinix/terraform-provider-equinix/issues/612


internal/resources/metal/vlans/datasource_test.go
```diff
@@ -18,9 +20,9 @@ func TestAccDataSourceMetalVlan_byVxlanFacility(t *testing.T) {
 	fac := "sv15"
 
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:          func() { testAccPreCheck(t) },
-		ExternalProviders: testExternalProviders,
-		Providers:         testAccProviders,
+		PreCheck:          func() { acceptance.TestAccPreCheckMetal(t) },
+		ExternalProviders: acceptance.TestExternalProviders,
+		ProviderFactories: acceptance.TestAccProviderFactories,
```
ctreatma, line 25: I see a lot of failures in CI with error messages like this:
```

=== CONT  TestAccMetalPortCreate_hybridBonded_timeout
    resource_metal_port_acc_test.go:424: Step 1/5, expected an error with pattern, no match on: Error running pre-apply plan: exit status 1
        
        Error: Invalid resource type
        
          on terraform_plugin_test.tf line 145, in resource "equinix_metal_vlan" "test":
         145: resource "equinix_metal_vlan" "test" {
        
        The provider hashicorp/equinix does not support resource type
        "equinix_metal_vlan".
--- FAIL: TestAccMetalPortCreate_hybridBonded_timeout (0.89s)
```

Any tests that have a `metal_vlan` resource will need to be updated to use the provider factory instead of the `Provider` or `Providers` key.  I recommend picking one test to run locally and figuring out the changes needed to get just that one working, and then copy that change to other VLAN-related tests.

aayushrangwala: Yeah looking into it

equinix/resource_metal_virtual_circuit_acc_test.go
```diff
@@ -129,8 +129,9 @@ func TestAccMetalVirtualCircuit_dedicated(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{ // Error: Error waiting for virtual circuit 863d4df5-b3ea-46ee-8497-858cb0cbfcb9 to be created: GET https://api.equinix.com/metal/v1/virtual-circuits/863d4df5-b3ea-46ee-8497-858cb0cbfcb9?include=project%2Cport%2Cvirtual_network%2Cvrf: 500 Oh snap, something went wrong! We've logged the error and will take a look - please reach out to us if you continue having trouble.
 		PreCheck:          func() { testAccPreCheck(t) },
 		ExternalProviders: testExternalProviders,
-		ProviderFactories: testAccProviderFactories,
-		CheckDestroy:      testAccMetalVirtualCircuitCheckDestroyed,
+		//ProviderFactories:        testAccProviderFactories,
+		ProtoV5ProviderFactories: testAccProtoV5ProviderFactories,
```
ctreatma, line 133: This fix enabled the test to use the metal_vlan resource, but it now fails with a different error:
```
    resource_metal_virtual_circuit_acc_test.go:129: Step 1/3 error: Error running apply: exit status 1
        
        Error: Value Conversion Error
        
          with equinix_metal_vlan.test,
          on terraform_plugin_test.tf line 28, in resource "equinix_metal_vlan" "test":
          28:         resource "equinix_metal_vlan" "test" {
        
        An unexpected error was encountered trying to convert tftypes.Value into
        vlans.DataSourceModel. This is always an error in the provider. Please report
        the following to the provider developer:
        
        mismatch between struct and object: Struct defines fields not found in
        object: vlan_id.
```

internal/resources/metal/vlans/resource.go
```diff
@@ -0,0 +1,186 @@
...
+							equinix_errors.FriendlyError(err).Error())
+						return
+					}
+				}
+			}
+		}
+	}
+
+	resp, err = client.ProjectVirtualNetworks.Delete(vlan.ID)
+	if resp.StatusCode != http.StatusForbidden && resp.StatusCode != http.StatusNotFound {
```
ctreatma, line 181: There should be an `err != nil` check in here

ctreatma: On second glance it looks like this used to use `equinix_errors.IgnoreResponseErrors` and should continue to do that.

internal/resources/metal/vlans/resource.go
```diff
@@ -0,0 +1,186 @@
...
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	vlan, resp, err := client.ProjectVirtualNetworks.Get(
+		data.ID.ValueString(),
+		&packngo.GetOptions{Includes: []string{"instances", "meta_gateway"}},
+	)
+	if err != nil {
+		if resp.StatusCode != http.StatusForbidden && resp.StatusCode != http.StatusNotFound {
```
ctreatma, line 152: While this block structure change is, as far as I can tell, equivalent to the original, changing things like this introduces unnecessary risk.  Is there an incompatibility between framework code and the `IgnoreResponseErrors` function?

internal/resources/metal/vlans/resource.go
aayushrangwala: Yes fixed.

ctreatma:
There are still a bunch of tests failing due to how the provider is set up.  I updated all Metal acceptance tests in #586.  Now that that is merged, you can rebase this PR on `main` and that should reduce the number of test failures.

internal/resources/metal/vlans/resource.go
```diff
@@ -0,0 +1,183 @@
...
+						response.Diagnostics.AddError("Error unassign port with Vlan",
+							equinix_errors.FriendlyError(err).Error())
+						return
+					}
+				}
+			}
+		}
+	}
+
+	if equinix_errors.IgnoreResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(client.ProjectVirtualNetworks.Delete(vlan.ID)) != nil {
```
ctreatma, line 178: There's no `err` defined here, so line 180 throws a panic when it is reached.  This should look more like:
```suggestion
	err := equinix_errors.IgnoreResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(client.ProjectVirtualNetworks.Delete(vlan.ID))
	if  err != nil {
```
```

Comment (aayushrangwala) [reply internal/resources/metal/vlans/resource.go]: ohkay, updated

**Comment by aayushrangwala** [comment]
https://github.com/equinix/terraform-provider-equinix/actions/runs/8024567143/job/21924204857?pr=578
All the related tests are passing. Tests involving device creation are failing due to unavailability of resources

File: internal/resources/metal/vlans/resource.go (hunk 4 above)
Comment (aayushrangwala) [reply internal/resources/metal/vlans/resource.go]: I checked and functionally there are no differences. Still I reused `IgnoreResponseErrors`

File: internal/resources/metal/vlans/datasource_test.go [hunk 6]
```diff
@@ -322,7 +277,7 @@ func TestMetalVlan_matchingVlan(t *testing.T) {
 }
 
 func testAccMetalDatasourceVlanCheckDestroyed(s *terraform.State) error {
-	client := testAccProvider.Meta().(*config.Config).Metal
+	client := acceptance.TestAccProvider.Meta().(*config.Config).Metal
 
 	for _, rs := range s.RootModule().Resources {
 		if rs.Type != "equinix_metal_vlan" {
```
ctreatma, line 283: Looks good!  Guidance from @ocobles is to add an `_upgradeFromVersion` test as well so we can be more confident that we haven't introduced breaking changes by migrating to framework.  Could you add one test like that for the datasource and one for the resource.   Here's an example from Oscar's most recent framework migration PR: https://github.com/equinix/terraform-provider-equinix/pull/572/files#diff-dc80edad0c4ecf395554d4f118aa4dbb2f5963b52a7eb4ae8aa8e57184b1ee8eR128

aayushrangwala: Yes, added upgrade tests as well. Passing locally, please approve

aayushrangwala:
Tests are passing now

internal/resources/metal/vlans/resource.go
```diff
@@ -0,0 +1,183 @@
...
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	response.Diagnostics.Append(response.State.Set(ctx, &data)...)
+}
+
+func (r *Resource) Update(ctx context.Context, request resource.UpdateRequest, response *resource.UpdateResponse) {
+	//TODO implement me
+	panic("implement me")
```
ctreatma, line 132: This should not panic.  See @displague's relevant comment on a different PR: https://github.com/equinix/terraform-provider-equinix/pull/588#discussion_r1502936323

internal/resources/metal/vlans/resource_schema.go
```diff
@@ -0,0 +1,98 @@
...
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.Expressions{
+						path.MatchRoot("metro"),
+					}...),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+				// TODO: aayushrangwala to check if this is needed with the framework changes
```
ctreatma, line 52: Was this implemented somewhere or does it still need to be done?

internal/resources/metal/vlans/resource_schema.go
```diff
@@ -0,0 +1,98 @@
...
+				//
+				// suppress diff if metro is manually set for first time, and
+				// facility is already set
+				//if len(new) > 0 && old == "" && facOk {
+				//	return facOk
+				//}
+				//return old == new
+				//},
+				// TODO: add statefunc in framework
+				//StateFunc: converters.ToLowerIf,
```
ctreatma, line 86: Same question for these: were they implemented somewhere or do they still need to be done?

ctreatma: @ocobles can you show @aayushrangwala how you handled the StateFunc for gateway?

aayushrangwala: Handled in parse functions

internal/resources/metal/vlans/resource.go
aayushrangwala: Yes removed panic

internal/resources/metal/vlans/resource_schema.go
```diff
@@ -0,0 +1,74 @@
...
+			"metro": schema.StringAttribute{
+				Description: "Metro in which to create the VLAN",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.ConflictsWith(path.MatchRoot("facility")),
+					stringvalidator.AtLeastOneOf(path.MatchRoot("facility"), path.MatchRoot("metro")),
```
ctreatma, line 61: `stringvalidator.ExactlyOneOf(...)` might be a better fit here (I didn't notice that one when I added AtLeastOneOf).

internal/resources/metal/vlans/models.go
```diff
@@ -0,0 +1,81 @@
...
+		m.Facility = types.StringValue(vlan.FacilityCode)
+	}
+
+	m.Description = types.StringValue(vlan.Description)
+	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
+
+	// version of this resource. StateFunc doesn't exist in terraform and it requires implementation
+	// of bespoke logic before storing state. To ensure backward compatibility we ignore lower/upper
+	// case diff for now, but we may want to require input upper case
+	if !strings.EqualFold(m.Metro.ValueString(), vlan.MetroCode) {
```
ctreatma, line 42: The API seems to return `null` for `metro` and `metro_code` when a VLAN was created in a facility, even if you `include=metro`; for a VLAN created in a facility we have to read the metro code from `vlan.Facility.Metro.Code`

ctreatma: Also I think we need a test for the metro casing behavior; I suspect that an `EqualFold` check while parsing API responses isn't enough to suppress the diff, but I don't know.  The test would:
- create a vlan with `metro = "da"`
- change to `metro = "DA"`
- expect an empty plan

internal/resources/metal/vlans/sweeper.go
```diff
@@ -1,4 +1,4 @@
-package vlan
+package vlans
```
ctreatma, line 1: This (and the directory name) should be `vlan`; we've been using singular resource names for packages thus far.

aayushrangwala: Changed

internal/planmodifiers/caseinsensitive.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type caseInsensitivePlanModifier struct{}
+
+func (d *caseInsensitivePlanModifier) PlanModifyString(ctx context.Context, request planmodifier.StringRequest, response *planmodifier.StringResponse) {
+	oldValue := request.StateValue.ValueString()
+	newValue := request.PlanValue.ValueString()
+
+	result := oldValue
+	if !strings.EqualFold(strings.ToLower(newValue), strings.ToLower(oldValue)) {
```
ctreatma, line 22: Nit: EqualFold ignores case, so why convert the args to lower-case?

internal/resources/metal/vlan/resource_schema.go
```diff
@@ -0,0 +1,76 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"metro": schema.StringAttribute{
+				Description: "Metro in which to create the VLAN",
+				Optional:    true,
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					//stringplanmodifier.RequiresReplace(),
+					equinixplanmodifiers.CaseInsensitiveString(),
```
ctreatma, line 58: This might need a `UseStateForUnknown()` based on the CI run:
```
o/hashicorp/equinix" planned an invalid value
        for equinix_metal_vlan.test.metro: planned value cty.StringVal("") does not
        match config value cty.UnknownVal(cty.String).
```

internal/planmodifiers/caseinsensitive.go
aayushrangwala: Right fixed

internal/resources/metal/vlan/resource_schema.go
aayushrangwala: added

internal/resources/metal/vlan/resource.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// Parse API response into the Terraform state
+	response.Diagnostics.Append(data.parse(vlan)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	// Set state to fully populated data
+	response.Diagnostics.Append(response.State.Set(ctx, &data)...)
+	return
```
ocobles, line 104:
```suggestion
```

aayushrangwala: fixed

internal/resources/metal/vlan/datasource_test.go
```diff
@@ -335,3 +291,53 @@ func testAccMetalDatasourceVlanCheckDestroyed(s *terraform.State) error {
 
 	return nil
 }
+
+func TestAccDataSourceMetalVlan_byVxlanMetro_upgradeFromVersion(t *testing.T) {
+	rs := acctest.RandString(10)
+	metro := "sv"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acceptance.TestAccPreCheckMetal(t) },
```
ctreatma, line 300: This test is failing because of differences between the SDKv2 resource state and the Framework resource state:

```
=== NAME  TestAccDataSourceMetalVlan_byVxlanMetro_upgradeFromVersion
    datasource_test.go:299: Step 2/2 error: Pre-apply plan check(s) failed:
        expected empty plan, but data.equinix_metal_vlan.bardsvlan has planned action(s): [read]
        expected empty plan, but data.equinix_metal_vlan.dsvlan has planned action(s): [read]
        expected empty plan, but equinix_metal_vlan.barvlan has planned action(s): [delete create]
        expected empty plan, but equinix_metal_vlan.foovlan has planned action(s): [delete create]
```

We'll need to look through the old schema and the new schema and try to figure out which attributes could be different.  This might be a matter of missing/different default values; missing `UseStateForUnknown`; Framework resource storing an attribute as `""` whereas SDKv2 resource stored it as `nil` (or vice versa).

I haven't found a way to get the acceptance test to print the plan to stdout, but maybe there is a way to do that?  Another way to dig into problems like this is to temporarily print the state and plan to stdout in the resource's `Update` function and run this specific test to see what the differences are.

aayushrangwala: handled this by adding a default value in facility

internal/resources/metal/vlan/models.go
```diff
@@ -0,0 +1,98 @@
...
+	facilityCode := types.StringNull()
+	if vlan.Facility != nil {
+		facilityCode = types.StringValue(vlan.Facility.Code)
+		metroCode = types.StringValue(strings.ToLower(vlan.Facility.Metro.Code))
+	}
+	// version of this resource. StateFunc doesn't exist in terraform and it requires implementation
+	// of bespoke logic before storing state. To ensure backward compatibility we ignore lower/upper
+	// case diff for now, but we may want to require input upper case
+	if !strings.EqualFold(m.Facility.ValueString(), facilityCode.ValueString()) {
+		m.Facility = facilityCode
```
ctreatma, line 88: I see a bunch of failures in CI due to facility changing from `""` to nil after apply.  Not sure where it could be coming from, but maybe here?  This should probably just set `m.Facility = facilityCode` without the `EqualFold` check so that `m.Facility` is always set.

internal/resources/metal/vlan/models.go
```diff
@@ -0,0 +1,98 @@
...
+	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
+
+	if vlan.Project.ID != "" {
+		m.ProjectID = types.StringValue(vlan.Project.ID)
+	}
+
+	var metroCode types.String
+	facilityCode := types.StringNull()
+	if vlan.Facility != nil {
+		facilityCode = types.StringValue(vlan.Facility.Code)
```
ctreatma, line 81: Rather than the EqualFold check below, this should be converted to lower-case just like metro code
```suggestion
		facilityCode = types.StringValue(strings.ToLower(vlan.Facility.Code))
```

aayushrangwala: fixed

internal/resources/metal/vlan/models.go
ctreatma: Even with the recommended changes, there are still a large number of tests failing due to facility being nil after apply.  Are all of the VLAN tests using a dedicated connection or do some of them use shared connections so they can be run locally?

aayushrangwala: All the tests are passing now. Although a few of them failing due to device unavailablility and timeout

internal/resources/metal/vlan/resource_test.go
```diff
@@ -0,0 +1,243 @@
...
+						"equinix_metal_vlan.foovlan", "metro", metro),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "facility", facility),
+				),
+			},
+			{
+				Config:             testAccCheckMetalVlanConfig_metro(rs, metro, "tfacc-vlan"),
+				ExpectNonEmptyPlan: false,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
```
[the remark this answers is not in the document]
displague: @ctreatma do you mean `resource.TestCheckResourceAttr`?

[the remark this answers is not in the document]
ctreatma: yup, updated

internal/resources/metal/vlan/resource_test.go
```diff
@@ -0,0 +1,243 @@
...
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "metro", metro),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "facility", facility),
```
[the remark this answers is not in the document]
ctreatma: Actually I think your original test expectation was correct; the API will still return `facility` because the VLAN was created in a facility, so the attribute should stay the same after migrating.

internal/resources/metal/vlan/resource_test.go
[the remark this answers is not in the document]
ctreatma: I'm misreading the test; it should already be preventing recreation via `ExpectNonEmptyPlan: false` (i.e. the test expects an empty plan already).

ctreatma:
Needed to update the resource schema slightly so that:
- `facility` only triggers replacement if it's explicitly configured
- `facility` doesn't default to an empty string
- `metro` change always requires replacement
https://github.com/equinix/terraform-provider-equinix/pull/578/commits/5fc741715fedba4dbb5e82e990dea703f25e377b

ctreatma (approved):
After updating the resource schema to handle switching from facility to metro, the only test failure in CI was a post-test cleanup failure in a metal port test.

PR #580, opened 2024-02-21
docs: Adding Network Resource and data source examples to the doc

srushti-patl:
Added Fabric Network Resource and Data Source examples to the docs


docs/resources/equinix_fabric_network.md
```diff
@@ -10,6 +10,23 @@ description: |-
...
+resource "equinix_fabric_network" "new_network"{
+  name = "Network-SV"
+  type = "EVPLAN"
+  scope = "GLOBAL"
+  notifications{
+    type = "ALL"
+    emails = ["example@equinix.com","test1@equinix.com"]
+  }
+  project {
+  	project_id = "776847000642406"
```
thogarty, line 25: nit: indentation is different for notifications and for project. Terraform fmt resolves to 2 spaces per tab. Our examples should follow the same format.

You can modify your editor settings to have it default to that for future things.

docs/data-sources/equinix_fabric_network.md
```diff
@@ -10,6 +10,14 @@ description: |-
 
 Fabric V4 API compatible data resource that allow user to fetch Fabric Network for a given UUID
 
+## Example Usage
+
+```hcl
+data "equinix_fabric_network" "network_data_name" {
```
thogarty, line 16: This is perfect for how the data source is used.

Though something we're working towards is having the data source examples include outputs that provide the attribute mappings (data selections like how we do with connections getting the redundancy group) so the customers are educated on how to retrieve the different elements from the data source.

If you find it to be quick then include. If not just let me know in a comment and we'll setup a task to pickup in Jira for getting the data source docs updated all at once.

thogarty (approved):
@srushti-patl Approved, but please update the indentation on the project_id before merging.

docs/data-sources/equinix_fabric_network.md
srushti-patl: Could we follow up on this task later? I will create a JIRA task for updating all the data sources

thogarty: Sounds good!

docs/resources/equinix_fabric_network.md
thogarty: Indentation is still off.

thogarty (approved):
Approving; you should be able to fix indentation and merge after.

Issue #539, opened 2024-01-24
[Bug]: Metal Hardware reservations are hanging without timing out during delete


displague:
### Terraform Version

1.7

### Equinix Provider Version

1.26

### Effected Terraform Resources

equinix_metal_device

### Expected Behavior

A deprovision should succeed or timeout

### Actual Behavior

A failed deprovision hangs indefinitely

### Steps to Reproduce

* create an instance from a hardware reservation
* attempt to delete the instance
* experience a timeout

displague:
I didn't expect a search for `context.TODO` to be as obviously helpful as it seems to be (with only 2 matches):
https://github.com/equinix/terraform-provider-equinix/blob/11dbb2b4741991b6db133c52ed0250d3623c6d52/equinix/helpers_device.go#L115

`context.Background` matched 1 non-testing result which could also be related:
https://github.com/equinix/terraform-provider-equinix/blob/11dbb2b4741991b6db133c52ed0250d3623c6d52/equinix/data_source_metal_devices.go#L54-L91

`context.Background` and `context.TODO` should be replaced with calls that pass the Terraform context into the EM API requests. 

The user-defined resource timeouts should be tested and honored, for ondemand instance deprovision and reservation instance deprovisions.

displague:
hwReservationStateRefreshFunc could take ctx as a parameter since waitUntilReservationProvisionable has a context that it could send along. This seems familiar and we may have intentionally removed that context in the past.

https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#statechangeconf has some notes on how to treat CRUD operations within a StateChangeConf refresh.

ctreatma:
Closed by #581

PR #581, opened 2024-02-21
fix: maintain provider context in resources and datasources

ctreatma:
We recently introduced equinix-sdk-go as a context-aware replacement SDK for packngo and _also_ migrated some resources and datasources to use context-aware CRUD lifecycle methods.  In those parallel changes we missed some opportunities to pass the terraform provider context into the SDK.

This updates the provider to replace all non-testing usage of `context.TODO()` and  `context.Background()` with references to the pre-existing terraform provider context.

Relates to #539

### DIFF
--- a/equinix/data_source_metal_devices.go
+++ b/equinix/data_source_metal_devices.go
@@ -51,7 +51,7 @@ func dataSourceMetalDevices() *schema.Resource {
 	return datalist.NewResource(dataListConfig)
 }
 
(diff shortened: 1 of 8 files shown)


ocobles:
There are some resources that have already been moved to internal that are still using `context.Background()`, shouldn't they be updated as well?

image

ctreatma:
Uh...I don't know why those changes aren't here.  Yes, those should be updated.

PR #582, opened 2024-02-21
fix: Updating Fabric Service Profile resource mapping functions

srushti-patl:
fixed mapping functions in resource_fabric_service_profile.go and fabric_common_mapping_helpers.go


equinix/resource_fabric_service_profile.go
```diff
@@ -936,22 +929,20 @@ func accessPointColoFabricSpToTerra(accessPointColol []v4.ServiceProfileAccessPo
 	return mappedAccessPointColol
 }
 
-func serviceMetroFabricSpToTerra(serviceMetrol []v4.ServiceMetro) []map[string]interface{} {
+func serviceMetroFabricSpToTerra(serviceMetrol []v4.ServiceMetro) []interface{} {
```
thogarty, line 932: If the method input is a slice, then we need to keep the for loop. We will add map[string]interface{} elements to a []interface{} made with the right length by index. Then return the []interface{} directly. We won't need to use append().

equinix/resource_fabric_service_profile.go
```diff
@@ -880,49 +880,42 @@ func customFieldFabricSpToTerra(customFieldl []v4.CustomField) []map[string]inte
...
 	return marketingInfoSet
 }
 
-func accessPointColoFabricSpToTerra(accessPointColol []v4.ServiceProfileAccessPointColo) []map[string]interface{} {
+func accessPointColoFabricSpToTerra(accessPointColol []v4.ServiceProfileAccessPointColo) []interface{} {
 	if accessPointColol == nil {
 		return nil
 	}
-	mappedAccessPointColol := make([]map[string]interface{}, len(accessPointColol))
+	mappedAccessPointColol := make([]interface{}, len(accessPointColol))
```
thogarty, line 918: The `l` suffix to represent list is erroneous and confusing. We can just add an `s` to indicate the plurality of it.

equinix/resource_fabric_service_profile.go
```diff
@@ -880,49 +880,42 @@ func customFieldFabricSpToTerra(customFieldl []v4.CustomField) []map[string]inte
 	return mappedCustomFieldl
 }
 
-func processStepFabricSpToTerra(processStepl []v4.ProcessStep) []map[string]interface{} {
+func processStepFabricSpToTerra(processStepl []v4.ProcessStep) []interface{} {
```
thogarty, line 883: If the method input is a slice, then we need to keep the for loop. We will add `map[string]interface{}` elements to a `[]interface{}` made with the right length by index. Then return the `[]interface{}` directly. We won't need to use `append()`.

equinix/resource_fabric_service_profile.go
```diff
@@ -863,11 +863,11 @@ func resourceServiceProfilesSearchRequest(ctx context.Context, d *schema.Resourc
 	return setFabricServiceProfilesListMap(d, serviceProfiles)
 }
 
-func customFieldFabricSpToTerra(customFieldl []v4.CustomField) []map[string]interface{} {
+func customFieldFabricSpToTerra(customFieldl []v4.CustomField) []interface{} {
 	if customFieldl == nil {
 		return nil
 	}
-	mappedCustomFieldl := make([]map[string]interface{}, len(customFieldl))
+	mappedCustomFieldl := make([]interface{}, len(customFieldl))
```
thogarty, line 870: Use `s` as the suffix instead of `s` to indicate multiple items. Lower cognitive overhead when reading the names.

equinix/resource_fabric_service_profile.go
thogarty: I know during our session I may have started this one, but was just trying to represent the concept and missed the input being a list type.

internal/fabric/schema/fabric_common_mapping_helpers.go
```diff
@@ -140,19 +140,15 @@ func LocationToTerra(location *v4.SimplifiedLocation) *schema.Set {
...
-	mappedLocations := make([]interface{}, len(locations))
-	for i, location := range locations {
-		mappedLocations[i] = map[string]interface{}{
-			"region":     location.Region,
-			"metro_name": location.MetroName,
-			"metro_code": location.MetroCode,
-			"ibx":        location.Ibx,
-		}
-	}
+	mappedLocations := make(map[string]interface{})
```
thogarty, line 143: Nice 👍

thogarty (changes requested):
Changes requested. Let me know if you have questions.

equinix/resource_fabric_service_profile.go
```diff
@@ -1114,13 +1109,13 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
 		return v4.MarketingInfo{}
 	}
 	marketingInfos := v4.MarketingInfo{}
-	for _, marketingInfo := range schemaMarketingInfo {
-		miLogo := marketingInfo.(map[string]interface{})["logo"].(string)
-		miPromotion := marketingInfo.(map[string]interface{})["promotion"].(bool)
+	for _, mti := range schemaMarketingInfo {
```
thogarty, line 1112: Let's leave the range value as `marketingInfo`.

equinix/resource_fabric_service_profile.go
```diff
@@ -1114,13 +1109,13 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
 		return v4.MarketingInfo{}
```
thogarty, line 1109: The method should return a `*v4.MarketingInfo`, and if the input is `nil` then we should return `nil`.

equinix/resource_fabric_service_profile.go
```diff
@@ -1133,40 +1128,40 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
...
+	if processSteps == nil {
 		return []v4.ProcessStep{}
 	}
-	var processSteps []v4.ProcessStep
-	for _, processStep := range processStepl {
-		psTitle := processStep.(map[string]interface{})["title"].(interface{}).(string)
-		psSubTitle := processStep.(map[string]interface{})["sub_title"].(interface{}).(string)
-		psDescription := processStep.(map[string]interface{})["description"].(interface{}).(string)
-		processSteps = append(processSteps, v4.ProcessStep{
+	var processStepRes []v4.ProcessStep
```
thogarty, line 1135: We should initiate this with make so that we can make it the same length as the input. Then we should also name it `processStepsRes` to account for the plural.

equinix/resource_fabric_service_profile.go
```diff
@@ -1133,40 +1128,40 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
...
+	for _, pcs := range processSteps {
```
thogarty, line 1136: We should use `index, processStep` for the range variables. Then inside the loop we should assign to `processStepsRes` by index. This matches the return type of `[]v4.ProcessStep`

equinix/resource_fabric_service_profile.go
```diff
@@ -1133,40 +1128,40 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
...
 	}
-	return processSteps
+	return processStepRes
 }
 
 func portsToFabric(schemaPorts []interface{}) []v4.ServiceProfileAccessPointColo {
 	if schemaPorts == nil {
 		return []v4.ServiceProfileAccessPointColo{}
 	}
 	var ports []v4.ServiceProfileAccessPointColo
```
thogarty, line 1153: You can either revert this method to how it was before, or I'm gonna ask for a lot of changes.

First, every `[]` variable should be made with `make` so that we can assign by index and elminate the extra processing needing by `append`.

All range variables should be `index, full_singular_name`.

All assignments should be done by index and not with `append`.

thogarty (changes requested):
Got a lot of changes requested. They need to be fulfilled exactly otherwise I will ask for them again. Let me know if you have any questions or if you need a call to go over it.

equinix/resource_fabric_service_profile.go
```diff
@@ -1044,44 +1039,44 @@ func accessPointTypeConfigToTerra(spAccessPointTypes []v4.ServiceProfileAccessPo
 	return mappedSpAccessPointTypes
 }
 
-func apiConfigToFabric(apiConfigl []interface{}) v4.ApiConfig {
-	if apiConfigl == nil {
+func apiConfigToFabric(apiConfigs []interface{}) v4.ApiConfig {
```
thogarty, line 1042: Marking this method for live code review.

equinix/resource_fabric_service_profile.go
```diff
@@ -1130,69 +1125,69 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
...
 			SellerRegionDescription: pSellerRegionDescription,
 			CrossConnectId:          pCrossConnectId,
-		})
+		}
 	}
-	return ports
+	return schemaPortRes
 }
 
 func virtualDevicesToFabric(schemaVirtualDevices []interface{}) []v4.ServiceProfileAccessPointVd {
```
thogarty, line 1177: Marking method for live code review.

equinix/resource_fabric_service_profile.go
```diff
@@ -1208,14 +1203,14 @@ func metrosToFabric(schemaMetros []interface{}) []v4.ServiceMetro {
 		return []v4.ServiceMetro{}
 	}
 	var metros []v4.ServiceMetro
```
thogarty, line 1205: Marking method for live code review.

equinix/resource_fabric_service_profile.go
```diff
@@ -1044,44 +1039,44 @@ func accessPointTypeConfigToTerra(spAccessPointTypes []v4.ServiceProfileAccessPo
...
 			EquinixManagedPort: psEquinixManagedPort,
 		}
 	}
-	return apiConfigs
+	return apiConfigRes
 }
 
-func authenticationKeyToFabric(authenticationKeyl []interface{}) v4.AuthenticationKey {
-	if authenticationKeyl == nil {
+func authenticationKeyToFabric(authenticationKeys []interface{}) v4.AuthenticationKey {
```
thogarty, line 1064: Marking method for live code review.

equinix/resource_fabric_service_profile.go
```diff
@@ -1130,69 +1125,69 @@ func marketingInfoToFabric(schemaMarketingInfo []interface{}) v4.MarketingInfo {
...
 	}
-	var processSteps []v4.ProcessStep
-	for _, processStep := range processStepl {
+	processStepRes := make([]v4.ProcessStep, len(processSteps))
+	for index, processStep := range processSteps {
 		psTitle := processStep.(map[string]interface{})["title"].(interface{}).(string)
 		psSubTitle := processStep.(map[string]interface{})["sub_title"].(interface{}).(string)
 		psDescription := processStep.(map[string]interface{})["description"].(interface{}).(string)
-		processSteps = append(processSteps, v4.ProcessStep{
+		processStepRes[index] = v4.ProcessStep{
```
thogarty, line 1140: Marking variable naming for live code review.

thogarty (changes requested):
Pending live code review.

thogarty (approved):
We're good here. Shipping it.

PR #584, opened 2024-02-22
chore(ci): skip shared connections when sweeping virtual circuits

ctreatma:
The sweeper step of our end-to-end test workflow sometimes fails because it finds & attempts to delete virtual circuits that belong to shared interconnections.  The API blocks deletion of these circuits because they are API-managed, not user-managed, and can only be deleted by deleting the interconnection itself.  This adds a check to the virtual circuit sweeper so that the sweeper will ignore virtual circuits that belong to shared interconnections.


displague (approved):

displague:
Unrelated sweeper warnings:

> Error: /22 16:12:16 [ERROR] Sweeper (equinix_metal_vlan) has dependency (equinix_metal_vrf), but that sweeper was not found
> Error: /22 16:12:16 [ERROR] Sweeper (equinix_metal_organization) has dependency (equinix_metal_project), but that sweeper was not found

PR #589, opened 2024-02-23
chore(ci): skip sweeping devices that are not in a deletable state

ctreatma:


equinix/resource_metal_device_acc_test.go
```diff
@@ -38,38 +39,46 @@ func init() {
...
-			if isSweepableTestResource(d.Hostname) {
-				dids = append(dids, d.ID)
+		for _, d := range ds.Devices {
+			if isSweepableTestResource(d.GetHostname()) {
+				nonSweepableDeviceStates := []metalv1.DeviceState{
+					metalv1.DEVICESTATE_PROVISIONING,
+					metalv1.DEVICESTATE_DEPROVISIONING,
+				}
+				if slices.Contains(nonSweepableDeviceStates, d.GetState()) {
+					log.Printf("[WARNING] skipping sweep for device %s because it is still %s", d.GetId(), d.GetState())
```
displague, line 72: We could include a note checking/confirming termination is in use. This would keep new maintainers from worrying about runaway resources.

displague (approved):

displague:
The sweeper error messages are not triggering that step to reflect a failed status.

```
Error: /23 19:09:17 [ERROR] Sweeper (equinix_metal_organization) has dependency (equinix_metal_project), but that sweeper was not found
```

I noticed that we have 130+ `tfacc-` projects in the testing org, 6 days old.  It seems that this error may be more meaningful than initially thought.

PR #591, opened 2024-02-23
feat!: upgrade to tfprotov6

ctreatma:
During the migration from SDKv2 to framework we've run into cases where code changes that are suggested in the documentation do not work because we are using Terraform Plugin Protocol v5. 

Adopting Terraform Plugin Protocol v6 gives us access to nicer ways of defining nested attributes, but it means dropping support for terraform < 1.0, which is technically a breaking change regardless of whether anyone is actually using this provider with older versions of terraform.  As a result, this PR uses the `!` Conventional Commits modifier to mark this change as a breaking change.

**TODO**: actually adopt the nested attribute changes so we can see if it's a meaningful improvement

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This repository is [Maintained](https://github.com/equinix-labs/equinix-labs/blo
 
 ## Requirements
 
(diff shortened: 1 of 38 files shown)


ctreatma:

PR #593, opened 2024-02-26
chore(ci): update sweeper registration for Metal resources

ctreatma:
When we migrated some resources to separate packages under `internal/resources`, the sweepers for the resources that were moved stopped running.  This adds a new `internal/sweeper` package that sets up the sweepers for resources under `internal/resources`.

In most cases, that change would be enough by itself, but the Metal sweepers have some interdependencies, and those interdependencies only work correctly if every relevant sweeper is registered within the same TestMain function.  To resolve that, this also moves the test sweepers for Metal resources that are still in `equinix/` into `internal/resources/metal` so that `internal/sweep` can know about all Metal sweepers regardless of where the resource itself is implemented.  As we continue with the various in-progress refactorings, the resource implementations will eventually land alongside their sweepers.


.github/workflows/acctest.yml
```diff
@@ -119,10 +119,9 @@ jobs:
         METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
         SWEEP: "all" #Flag required to define the regions that the sweeper is to be ran in
         SWEEP_ALLOW_FAILURES: "true" #Enable to allow Sweeper Tests to continue after failures
-        SWEEP_DIR: "./equinix"
       run: |
         # Added sweep-run to filter Metal test
-        go test ${SWEEP_DIR} -v -timeout 180m -sweep=${SWEEP} -sweep-allow-failures=${SWEEP_ALLOW_FAILURES} -sweep-run=$(grep -o 'AddTestSweepers("[^"]*"' equinix/resource_metal_* |cut -d '"' -f2 | paste -s -d, -)
+        go test $(go list ./... | grep 'internal/sweep\|equinix/equinix') -v -timeout 180m -sweep=${SWEEP} -sweep-allow-failures=${SWEEP_ALLOW_FAILURES} -sweep-run=$(grep -or 'AddTestSweepers("[^"]*"' | grep "_metal_" |cut -d '"' -f2 | paste -s -d, -)
```
ctreatma, line 124: This change has to be run via manual trigger because the pull_request_target trigger uses the workflow from the base branch (in this case, `main`).  Test run link pending.

ctreatma: Taking the opportunity to update the sweeper to handle locked devices.  Manual run in progress here: https://github.com/equinix/terraform-provider-equinix/actions/runs/8071102002/job/22049993375

internal/resources/metal/project/sweeper.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/equinix/terraform-provider-equinix/internal/sweep"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func AddTestSweeper() {
```
displague, line 11: This pattern is consistent with https://github.com/hashicorp/terraform-provider-aws/blob/e8e0ab740dae669a06b82f0a5a27ca7d1f311df3/internal/service/elasticache/sweep.go#L27 

Aside from having a different function name:
- sweeper is exposed as a package exported function
- calls `resource.AddtestSweepers` directly
- sweeper resides in non `_test.go` file

internal/resources/metal/project/sweeper.go
```diff
@@ -0,0 +1,44 @@
...
+		return fmt.Errorf("[INFO][SWEEPER_LOG] Error getting project list for sweeping projects: %s", err)
+	}
+	pids := []string{}
+	for _, p := range ps {
+		if sweep.IsSweepableTestResource(p.Name) {
+			pids = append(pids, p.ID)
+		}
+	}
+	for _, pid := range pids {
+		log.Printf("Removing project %s", pid)
```
displague, line 37:
```suggestion
		log.Printf("[DEBUG] Removing project %s", pid)
```

internal/resources/metal/project/sweeper.go
```diff
@@ -0,0 +1,44 @@
...
+		Dependencies: []string{"equinix_metal_vlan"},
+		F:            testSweepProjects,
+	})
+}
+
+func testSweepProjects(region string) error {
+	log.Printf("[DEBUG] Sweeping projects")
+	config, err := sweep.GetConfigForMetal()
+	if err != nil {
+		return fmt.Errorf("[INFO][SWEEPER_LOG] Error getting configuration for sweeping projects: %s", err)
```
displague, line 23: In the AWS provider, errors from the sweeper don't have the `[INFO][SWEEPER_LOG]` prefixes, they have no prefix. Just calling this out, I don't know what is "right".

displague: A few lines later we return an error with no prefix.

internal/sweep/sweep_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/equinix/terraform-provider-equinix/equinix"
+	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/project"
+	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/ssh_key"
+	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/vrf"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestMain(m *testing.M) {
+	// Register legacy sweepers for resources balled up under equinix/
```
displague, line 14: Tempting to stub out directories for these resources in `internal/...` and setup the sweepers as the only residents ahead of further migration. It would make those migrations smaller in scope and keep this PR scoped to sweepers without the `equinix.` hack.

I'm surprised we don't have Sweepers for non-Metal resources.

displague (approved):

internal/sweep/sweep_test.go
ctreatma: Yeah, I like the idea of just moving these sweepers to internal now.  This went through a few iterations and it's no longer necessary to keep the Metal sweepers in `equinix/` so keeping them there is just adding churn.  We'll still need the Makefile change because non-Metal sweepers are still under Equinix (it's fine to have them defined across multiple packages in general; the reason the Metal ones all have to be added in the same TestMain is because they have interdependencies).

PR #594, opened 2024-02-26
chore(ci): update SWEEP_DIR to find all sweepers

ctreatma:
Lately, we've seen error messages for the sweeper in CI that look like this:

```
2024/02/26 12:17:49 [ERROR] Sweeper (equinix_metal_organization) has dependency (equinix_metal_project), but that sweeper was not found
```

I started working on [a much more involved refactor to attempt to address this](https://github.com/equinix/terraform-provider-equinix/pull/593), but that PR still has error messages like the one above in its CI output.  It turns out that in both the `GNUmakefile` and in the `acctest.yml` GitHub Actions workflow, we were specifying a `SWEEP_DIR` of `./equinix`, which will exclude anything that exists under `./internal`.

I updated both locations to specify `./` as the `SWEEP_DIR`.  I manually ran `make sweep` to validate locally; the automatic CI run for this PR will not use the updated `acctest.yml`, so I will kick off a manual run of the e2e tests for this PR's head branch and link to the results in the PR comments.


ctreatma:
Manual run of the updated workflow for this branch is happening here: https://github.com/equinix/terraform-provider-equinix/actions/runs/8054910005/job/22000576016

ctreatma:
Collapsed this into #593 because it turns out we need a bit of both, plus some other stuff.

PR #598, opened 2024-02-29
test: configure provider for upgradefromversion tests

ocobles:
when we define the ProviderFactories at TestStep level instead of TestCase and we run that single test it fails in CheckDestroy function because the client Meta() interface is nil. This PR adds a precheck to make sure the provider is always configured.

### DIFF
--- a/internal/acceptance/acceptance.go
+++ b/internal/acceptance/acceptance.go
@@ -2,6 +2,7 @@ package acceptance
 
 import (
 	"os"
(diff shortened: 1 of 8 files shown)


ctreatma (approved):
Looks good to me; fixes the problem and is nicely constrained.  Long term I still think we should have a separate pathway for getting API clients in tests rather than going through the test provider, but that's not likely to happen any time soon.

PR #600, opened 2024-02-29
run Fabric integration tests when dependencies change

ctreatma:
Fabric acceptance tests only run when `_fabric_` code files are changed.  That means the workflow could miss breaking changes due to dependency changes (such as SDK upgrades).  This updates the workflow to trigger on PRs that change `go.mod` or `go.sum`, in addition to the existing triggers for Fabric-specific files.


thogarty (approved):
LGTM, thanks @ctreatma !

PR #601, opened 2024-02-29
refactor: delay Fabric v4 token exchange until resources are used

ctreatma:
This updates Fabric v4 client creation to allow the client to get an OAuth token when it is needed rather than running a token exchange at provider startup.  This reduces the risk of tokens expiring for long-running configurations and brings Fabric client initialization in line with how other Equinix API clients are initialized.

### DIFF
--- a/equinix/resource_fabric_routing_protocol.go
+++ b/equinix/resource_fabric_routing_protocol.go
@@ -3,11 +3,12 @@ package equinix
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"log"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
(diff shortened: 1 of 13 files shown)


equinix/resource_fabric_routing_protocol.go
```diff
@@ -287,7 +288,6 @@ func resourceFabricRoutingProtocol() *schema.Resource {
 
 func resourceFabricRoutingProtocolRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*config.Config).FabricClient
-	ctx = context.WithValue(ctx, v4.ContextAccessToken, meta.(*config.Config).FabricAuthToken)
```
ctreatma, line 290: Relying on the `authClient` for token initialization means we don't have to wire up the token in every Fabric resource lifecycle method; we can just use the client and trust that it will auth if it needs to.

ctreatma:
This functionality was added as part of the fabric-go => equinix-sdk-go migration in #606

Issue #160, opened 2022-01-24
metal_project_bgp_config


displague:
This is a continuation of https://github.com/packethost/terraform-provider-packet/issues/299 (limitations of importing a `metal_project` and adjust settings are discussed there).

If a resource for managing BGP configuration independent of `metal_project` interests you, please 👍 this issue. 

First, consider the following. If such a resource were provided today, it would be limited in the following ways:
* BGP can not be disabled once enabled
* BGP passwords can not be changed once set
* BGP ASN can not be changed once set
* BGP deployment type can not be changed once set

Because BGP settings affect all resources in the project, and these settings can not be temporarily adjusted, it may be an anti-pattern to attempt to control BGP settings independent of the project.  If the API were to provide new functionality with respect to the limitations pointed out above, I for one would reconsider the utility of a `metal_project_bgp_config` resource.

I would like to close this issue (and the original issue) as 'will not implement'. Please comment and emote on this issue to the contrary.

ocobleseqx:
IMO I would avoid a new (anti-pattern) resource. If I understood well the use case is: people import a terraform project to manage bgp, but if they want to destroy the terraform template and keep the project they can't. So in that case I think that a documented use of `terraform state rm` can help. Otherwise, (no my fav choice) what about an extra field in projects  like `on_destroy = delete|keep` or something like that ?

PR #604, opened 2024-03-01
refactor: migrate resource and datasource equinix_metal_project to framework 

ocobles:
Part of https://github.com/equinix/terraform-provider-equinix/issues/612


internal/resources/metal/project/datasource_test.go
```diff
@@ -4,14 +4,15 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	"github.com/equinix/terraform-provider-equinix/internal/acceptance"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	"github.com/packethost/packngo"
```
ctreatma, line 10: Nice!

internal/resources/metal/project/resource.go
```diff
@@ -3,311 +3,304 @@ package project
...
+	ctx context.Context,
+	req resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = resourceSchema(ctx)
 }
 
-func resourceMetalProjectCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*config.Config).NewMetalClientForSDK(d)
+func (r *Resource) Create(
```
ctreatma, line 37: Looks like something is missing here; haven't dug through this function yet but it looks like all tests that use a project resource are failing with this error:

```
=== NAME  TestAccDataSourceMetalPort_byId
    data_source_metal_port_acc_test.go:59: Step 1/1 error: Error running apply: exit status 1
        
        Error: Error creating project
        
          with equinix_metal_project.test,
          on terraform_plugin_test.tf line 109, in resource "equinix_metal_project" "test":
         109: resource "equinix_metal_project" "test" {
        
        Could not create project: API Error HTTP 404 404 Not Found
```

ocobles: it seems Im sending an empty organizationId when it is not defined

internal/resources/metal/project/resource.go
```diff
@@ -3,311 +3,304 @@ package project
...
-				return diag.Errorf("BGP Config can not be removed from a project, please add back\n%s", bgpConfStr)
-			}
-		}
-	} else {
-		_, resp, err := client.ProjectsApi.UpdateProject(ctx, d.Id()).ProjectUpdateInput(updateRequest).Execute()
+
+	// NOTE (ocobles): adding UpdateProject inside this condition to replicate old behavior
+	// but it is not clear to me if it was a mistake. I think the project should be updated if
+	// has changes regardless of whether there are changes to the BGP configuration or not.
+	// Open discussion: https://github.com/equinix/terraform-provider-equinix/discussions/466
```
ctreatma, line 250: I agree that this looks like a long-standing bug.  I say remove the check here so that it is possible to change BGP settings and core project settings in one plan/apply cycle, but we'll need to add checks here to make sure that the project settings _other than BGP_ are being changed so we don't send unnecessary update requests.

internal/resources/metal/project/bgp_config.go
```diff
@@ -56,7 +56,7 @@ func handleBGPConfigChanges(ctx context.Context, client *metalv1.APIClient, plan
 	bgpRemoved := plan.BGPConfig.IsNull() && !state.BGPConfig.IsNull()
```
ctreatma, line 56: Arguably out of scope, since this is aiming to replicate prior behavior, but:  could we catch and reject BGP settings removal earlier, maybe via config validation or plan modification?

internal/resources/metal/project/resource_schema.go
```diff
@@ -93,6 +93,9 @@ var bgpConfigSchema = map[string]schema.Attribute{
 	"asn": schema.Int64Attribute{
 		Description: "Autonomous System Number for local BGP deployment",
 		Required:    true,
+		PlanModifiers: []planmodifier.Int64{
+			int64planmodifier.RequiresReplace(),
```
ctreatma, line 97: This would result in replacing the entire project, right?  I'm not sure if we should take that risk or if it's better to reject the plan via validation.

internal/resources/metal/project/resource_test.go
```diff
@@ -211,17 +216,17 @@ func TestAccMetalProject_update(t *testing.T) {
 	})
 }
 
-func testAccCheckMetalSameProject(t *testing.T, before, after *packngo.Project) resource.TestCheckFunc {
+func testAccCheckMetalSameProject(t *testing.T, before, after *metalv1.Project) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		if before.ID != after.ID {
-			t.Fatalf("Expected device to be the same, but it was recreated: %s -> %s", before.ID, after.ID)
+		if before.GetId() != after.GetId() {
+			t.Fatalf("Expected device to be the same, but it was recreated: %s -> %s", before.GetId(), after.GetId())
```
ctreatma, line 222:
```suggestion
			t.Fatalf("Expected project to be the same, but it was recreated: %s -> %s", before.GetId(), after.GetId())
```

internal/resources/metal/project/resource_test.go
```diff
@@ -211,17 +216,17 @@ func TestAccMetalProject_update(t *testing.T) {
...
 		}
 		return nil
 	}
 }
 
 func TestAccMetalProject_BGPUpdate(t *testing.T) {
-	var p1, p2, p3 packngo.Project
+	var p1, p2, p3 metalv1.Project
```
ctreatma, line 229: This test is failing, which adds weight to avoiding the introduction of new `RequiresReplace` behavior in this PR:
```
=== NAME  TestAccMetalProject_BGPUpdate
    resource_test.go:222: Expected device to be the same, but it was recreated: 5025a781-8ff1-4537-876f-6d12c5552cd7 -> f308175d-fef9-46f8-a461-34698404e38f
```

internal/resources/metal/project/bgp_config.go
displague: Relates to:
* https://github.com/equinix/terraform-provider-equinix/issues/160
* https://github.com/packethost/terraform-provider-packet/issues/299

internal/resources/metal/project/resource_schema.go
displague: Relates to:
* https://github.com/equinix/terraform-provider-equinix/issues/160
* https://github.com/packethost/terraform-provider-packet/issues/299

ocobles: agree that reject the change will be better in this case

internal/resources/metal/project/bgp_config.go
ocobles: I've introduced a custom planmodifier immutablelist to cover that. I have done it generic in case it could be reused in the future. That's will be the expected output:

image

ctreatma:
The BGP update test was still failing because the project was being deleted and recreated.  After some digging I found it was because `organization_id` was `Computed` with `planmodifier.RequiresReplace()` but was not configured with `planmodifier.UseStateForUnknown()`.  As a result, the framework always set `organization_id` to unknown on plan, which marked the resource for replacement.

I updated the schema to add the missing plan modifier and updated the error message in the immutable list to match the existing test expectation: https://github.com/equinix/terraform-provider-equinix/pull/604/commits/30e64cc1743c9cd60120b8fec7a53350d03f015a

ctreatma (approved):
Ran CI twice and observed no failures in project resource / datasource tests and a couple flakes in unrelated tests.

Issue #595, opened 2024-02-28
[Bug]: 100G equinix_metal_connection not supported


lennardk:
### Terraform Version

Terraform v1.5.7

### Equinix Provider Version

version = "1.11.1"

### Effected Terraform Resources

equinix_metal_connection

### Terraform Config Files

data "equinix_metal_connection" interconnections {
    for_each      = var.virtual_circuits
    connection_id = each.value.connection_id
}

### Expected Behavior

100G ports are discovered for use

### Actual Behavior

`Error: 100000000000 is not allowed speed value. Allowed values: [50000000 200000000 500000000 1000000000 2000000000 5000000000 10000000000]`

### Steps to Reproduce

Order a 100G port
Try to use it it

There's no 'additional description' field so I'll just drop it here: probably just needs to be added to https://github.com/equinix/terraform-provider-equinix/blob/main/internal/resources/metal/connection/speed.go?

displague:
100G is documented as a dedicated port speed at https://deploy.equinix.com/product/interconnection/

ctreatma:
Support for arbitrary connection speeds in the `equinix_metal_connection` resource was added in v1.32.0.

Issue #354, opened 2023-08-10
Update `metal_connection` resource examples to use Fabric resources


ctreatma:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Description

All of the examples in the documentation for the `metal_connection` resource use the `ecx` resources for Fabric interactions.  In order to encourage users to adopt the new `fabric` resources, our examples should use `fabric` resources wherever possible.

### New or Affected Resource(s)

N/A

### Potential Terraform Configuration

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
```

### References

* https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_connection#example-usage

PR #605, opened 2024-03-01
fix: vlans is now required in the Metal API for shared connections

ctreatma:
The Metal API was recently updated to change the behavior of shared interconnections as follows:

- VLANs must be specified at creation for shared connections
- VLANs cannot be updated for shared connections

This updates the connection resource to align with the new API behavior so that users can catch configuration errors earlier.

Note this rolls back #273, which was written to address #222.

This also includes a fix for #595

### DIFF
--- a/internal/resources/metal/connection/resource.go
+++ b/internal/resources/metal/connection/resource.go
@@ -9,6 +9,7 @@ import (
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
 	"github.com/equinix/terraform-provider-equinix/internal/framework"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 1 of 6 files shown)


displague:
@ctreatma can we pick this up in the PR: https://github.com/equinix/terraform-provider-equinix/issues/595 (or another)

displague:
Putting blinders on to https://github.com/equinix/terraform-provider-equinix/issues/354

internal/resources/metal/connection/resource.go
```diff
@@ -38,6 +43,19 @@ func (r *Resource) Schema(
...
+func (r *Resource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var data ResourceModel
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.Type.ValueString() == "shared" {
```
displague, line 55: intentional dead code?

internal/resources/metal/connection/speed.go
```diff
@@ -19,6 +19,7 @@ var (
 		{2 * giga, "2Gbps"},
 		{5 * giga, "5Gbps"},
 		{10 * giga, "10Gbps"},
+		{100 * giga, "100Gbps"},
```
ctreatma, line 22: I don't love hiding this fix in an unrelated PR but it's so tiny.

(Side note, though, I have a bigger change to remove this list from the provider and just convert arbitrary speed values from int <=> string and let the API decide what's valid or not.)

0xch4z (approved):
LGTM

internal/resources/metal/connection/speed.go
displague:
```
=== RUN   TestSpeedConversion
    speed_test.go:26: Expected error converting invalid speed string to uint, got: 100000000000
    speed_test.go:31: Expected error converting invalid speed uint to string, got: 100Gbps
--- FAIL: TestSpeedConversion (0.00s)
```

displague: I've opened an issue to follow-up on this: https://github.com/equinix/terraform-provider-equinix/issues/609

PR #606, opened 2024-03-02
refactor: fabric connection equinix sdk go migration

thogarty:
* Removed all references to `fabric-go` and replaced with feature parity of `equinix-sdk-go/fabricv4`
  * Using reference to branch `equinix-sdk-go-migration` in code review on equinix-sdk-go/fabricv4 repo
  * Patches in this branch are necessary to resolve issues where the Fabric APIs are not adhering to API specs
  * Tickets have been cut for Fabric Teams to resolve; just using patching as short term fix
* Use Getters+Setters for all setting and retrieval of `equinix-sdk-go/fabricv4` structs
  * Improve request creation by only setting attributes on Go structs when they are given in the Terraform. Removes zero valued entries in API requests causing unintended side effects
* Updated acceptance tests to account for SDK update. 
* Updated config to create `equinix-sdk-go/fabricv4` using correct auth method
* Instantiated `equinix-sdk-go/fabricv4` with the correct user agent for Terraform to be used in metrics
* Handling error type assertions properly to avoid chaining errors during runtime


ctreatma:
Two questions:

> OpenAPIGenerator uses map[string]interface for the object API spec type and Fabric API has this as a string or object in its handling

Could you share links to the relevant parts of `fabric-go` and `equinix-sdk-go` for comparison to better understand the difference?

> Cannot update a single resource without changing everything 

Is there a heirarchy of these resources such that you could convert a "leaf" resource first, adding a shim to translate between old and new for that resource, and gradually push that up until all resources are converted?  That would increase the amount of code needed per resource for conversion but might allow more gradual adoption.  (For example, could you marshal the object for an equinix-sdk-go-based resource to JSON, and then unmarshal that JSON into a fabric-go object?)

equinix/resource_fabric_cloud_router.go
```diff
@@ -207,187 +207,187 @@ func resourceFabricCloudRouter() *schema.Resource {
...
+		createCloudRouterRequest.Order = order
 	}
 
-	fcr, _, err := client.CloudRoutersApi.CreateCloudRouter(ctx, createRequest)
+	fcr, _, err := client.CloudRoutersApi.CreateCloudRouter(ctx).CloudRouterPostRequest(createCloudRouterRequest).Execute()
 	if err != nil {
 		return diag.FromErr(equinix_errors.FormatFabricError(err))
 	}
-	d.SetId(fcr.Uuid)
+	d.SetId(*fcr.Uuid)
```
ctreatma, line 291: I recommend using the `Get<Attr>()` functions instead; their behavior more closely matches the old code and will save you from some (unfortunately not all) pointer headaches
```suggestion
	d.SetId(fcr.GetUuid())
```

internal/config/config.go
```diff
@@ -200,6 +200,48 @@ func (c *Config) NewFabricClient() *v4.APIClient {
 	return client
 }
 
+// NewFabricClientForSDK returns a terraform sdkv2 plugin compatible
+// equinix-sdk-go/fabricv4 client to be used to access Fabric's V4 APIs
+func (c *Config) NewFabricClientForSDK(d *schema.ResourceData) *fabricv4.APIClient {
```
ctreatma, line 205: Worth considering #601 here as well so you don't have to explicitly wire up the auth token in every CRUD method

thogarty:
> Two(-ish) questions:
> 
> > OpenAPIGenerator uses map[string]interface for the object API spec type and Fabric API has this as a string or object in its handling
> 
> Could you share links to the relevant parts of `fabric-go` and `equinix-sdk-go` for comparison to better understand the difference?
> 
> > Cannot update a single resource without changing everything
> 
> Is there a heirarchy of these resources such that you could convert a "leaf" resource first, adding a shim to translate between old and new for that resource, and gradually push that up until all resources are converted? That would increase the amount of code needed per resource for conversion but might allow more gradual adoption. (For example, could you marshal the object for an equinix-sdk-go-based resource to JSON, and then unmarshal that JSON into a fabric-go object?)

@ctreatma, Following up on first question; using CloudRouterChangeOperation as the example. You can see in the examples for the API calls that the `value` attribute is set to `object` in the spec. This allows it to handle either a simple string or a list of objects:

Handling list of objects:
image

Handling a simple string:
image

With the swagger-codegen fabric-go, the code generator creates the following model to handle this wide variety of possibles for `object` in the OAS. It uses `*interface{}`:
image

With the open-api-generator fabricv4, the code generator instead uses `map[string]interface{}` for the `object` type in OAS:
image

This example was just for the CloudRouterChangeOperation model, but all the update models for Fabric have the same issue in Go types as a result. So the update methods for Fabric resources are the ones that are having trouble with the migration.

It's difficult to handle the flexibility of different types for `object` in OAS with `map[string]interface{}`. Probably an incorrect definition in the spec to handle this, but the open-api-generator code generation isn't providing a good way to deal with this.

My first thought is to create patches for the update models to have them remain as `*interface{}` instead of `map[string]interface{}`, but I'm open to suggestions for a better way.

equinix/resource_fabric_cloud_router.go
thogarty: Will do 👍

internal/fabric/schema/fabric_common_mapping_helpers.go
```diff
@@ -85,58 +81,57 @@ func AccountToTerra[Account *v4.SimplifiedAccount | *v4.AllOfServiceProfileAccou
...
-func NotificationsToTerra(notifications []v4.SimplifiedNotification) []map[string]interface{} {
+func NotificationsGoToTerraform(notifications []fabricv4.SimplifiedNotification) []map[string]interface{} {
 	if notifications == nil {
 		return nil
 	}
 	mappedNotifications := make([]map[string]interface{}, len(notifications))
 	for index, notification := range notifications {
 		mappedNotifications[index] = map[string]interface{}{
-			"type":          notification.Type_,
+			"type":          notification.Type,
```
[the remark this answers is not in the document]
thogarty: Good call 👍

ctreatma:
@thogarty the `object` type in OpenAPI is a collection of property/value pairs; returning a plain string for a property that is documented as `type: object` is a violation of the spec. https://swagger.io/docs/specification/data-models/data-types/

The old generator used a slightly different approach for modeling the generic object type, and that approach happened to support responses that don't match the specification; you could patch the new SDK to match the old one, but it would be better to make the spec accurate. 

If this is truly a property that can be either an object or a string, then the spec needs to be updated to document it as
```yaml
oneOf:
- type: string
- type object
```
I'm not sure what the generator will spit out for that; more adjustments might be needed to get a good code interface.

It's possible the intent here was to declare `value` as [the `AnyType](https://swagger.io/docs/specification/data-models/data-types/#any), but again I'm not sure what the generator would spit out for that.

The example in the spec shows a simple string for "value"; can you share some valid requests that have other values in them?  How does the API actually process the value parameter?  Does it attempt to parse the value as JSON and then fall back to treating it as a plain string (or does it try other types before falling back to string)?

thogarty:
@ctreatma , this is definitely a spec violation issue. I've linked two examples for `value` here: 

> Handling list of objects: image
> 
> Handling a simple string: image

Based on those two examples it is being used in the API handling as two very different types:

Either one of the following:
```
type: array
items:
  type: object
  <undefined_object_properties>
```
or 
```
type: string
```

I think AnyType is the way to go here, but it still doesn't handle the issue where one type is `array` and the other is just `string`. I'll go back to Fabric on this one and figure out how to get this resolved.

Then I'll implement a patch in the spec for `equinix-sdk-go/fabricv4` until we can get the API Spec published.

In the meantime, I'll probably use the swagger-codegen implementation with a shim for the update API calls until this can be handled properly. Involves many different teams because all Fabric update endpoints are the culprits, and they are handled by different Fabric teams.

internal/config/config.go
displague: Also work taking a look at #617 while you are here

thogarty: @displague , the user agent is taken care of when switching over to equinix-sdk-go/fabricv4. That's included in this PR.

@ctreatma , have you tested #601? I tried testing it out with the change that I've made here and getting timeout exceeded while waiting for OAUTH token. I could've made an error in my attempt to make the same change with the sdk client creation method instead of just setting it during config.

thogarty:
@displague @ctreatma, this PR is ready for review. I'm on PTO till April 12th but will respond to comments when I'm back.

It's been tested with all Fabric Tests locally and doesn't include breaking changes. Our GHA is having strange oauth issues that we're still trying to iron out.

equinix/data_source_fabric_network_acc_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func TestAccDataSourceFabricNetwork_PFCR(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:          func() { acceptance.TestAccPreCheck(t) },
+		PreCheck:          func() { acceptance.TestAccPreCheck(t); acceptance.TestAccPreCheckProviderConfigured(t) },
```
ctreatma, line 13: Have you tried running these tests without the added call to `acceptance.TestAccPreCheckProviderConfigured(t)`?  That method was created to fix some tests that cover upgrading from an older version of the provider, because in those tests the first step runs with a pre-built existing release of the provider and that bypasses some necessary test setup.  I wouldn't expect it to be necessary here; is there something about how the Fabric parts of the provider are configured that requires adding this to all Fabric tests?

```suggestion
		PreCheck:          func() { acceptance.TestAccPreCheck(t) },
```

internal/errors/errors.go
```diff
@@ -69,16 +69,18 @@ func FormatFabricAdditionalInfo(additionalInfo []fabric.PriceErrorAdditionalInfo
...
 	errors = append(errors, err.Error())
-	if fabricErrs, ok := err.(fabric.GenericSwaggerError).Model().([]fabric.ModelError); ok {
-		for _, e := range fabricErrs {
-			errors = append(errors, fmt.Sprintf("Code: %s", e.ErrorCode))
-			errors = append(errors, fmt.Sprintf("Message: %s", e.ErrorMessage))
-			errors = append(errors, fmt.Sprintf("Details: %s", e.Details))
-			if additionalInfo := FormatFabricAdditionalInfo(e.AdditionalInfo); additionalInfo != "" {
-				errors = append(errors, fmt.Sprintf("AdditionalInfo: [%s]", additionalInfo))
+	if genericError, ok := err.(*fabricv4.GenericOpenAPIError); ok {
+		if fabricErrs, ok := genericError.Model().([]fabricv4.Error); ok {
```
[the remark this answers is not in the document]
thogarty: Sounds good. I'll make the adjustment 👍

equinix/data_source_fabric_network_acc_test.go
thogarty: With this one, we started getting very random oauth errors when running multiple fabric tests in parallel. The current main branch with `fabric-go` is failing frequently with those errors. When I tried to analyze it this was the only acceptance test change that occurred so I had made the assumption that it was causing it.

I merged in another change to main with this line included in the fabric tests hoping it would resolve the issue but it didn't take. I think with the new auth flow in this implementation the error doesn't persist and this was just to be sure it stayed in and wasn't being accidentally removed in this update. I'll double check and update you on what I find.

internal/errors/errors.go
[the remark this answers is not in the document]
thogarty: Shifted into the SDK: https://github.com/equinix/equinix-sdk-go/pull/51

srushti-patl (dismissed):
LGTM

ctreatma:
@thogarty the SDK v0.36.0 upgrade was merged in separately.  Could you rebase your PR on the latest from `main`?

srushti-patl (approved):

thogarty:
Hi @ctreatma, I've updated this for go.mod and go.sum. Using what is present in main so there aren't any changes. Srushti has approved this PR from the Fabric Governor perspective. We need a devrel approval before we can merge.

ctreatma (approved):

PR #607, opened 2024-03-03
Connection vrf support

0xch4z:
Depends on https://github.com/equinix/terraform-provider-equinix/pull/605 and some API spec updates

Changes:
- equinix_metal_connection now uses equinix-sdk-go instead of packngo
- "vrfs" field added to equinix_metal_connection resource + datasource
- "contact_email" can be updated without recreation
- "redundancy" updates cause recreation

Breaking changes:
- Updating "redundancy" forces recreate (this goes along with the
  previous change to force new on "vlans" changes, if a new vlan/vrf
  cannot be added then switching between primary and redundant should
  not be allowed)


go.mod
```diff
@@ -118,3 +118,5 @@ require (
 	google.golang.org/protobuf v1.32.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/equinix/equinix-go-sdk => ../../equinix-sdk-go/
```
displague, line 122: Is this 0.34.0?

0xch4z: This points to some manually patched open API specs while I wait for my spec PR to land.

I pushed this up to my fork here: https://github.com/0xch4z/equinix-sdk-go/commit/82dbfb149c9f3940e06eaff6c258cfaf407deeaa

internal/resources/metal/connection/models.go
```diff
@@ -122,51 +125,67 @@ func (m *ResourceModel) parse(ctx context.Context, conn *packngo.Connection) dia
 	return diags
 }
 
+func getAbstractVirtualCircuit(vc *metalv1.VirtualCircuit) abstractVirtualCircuit {
```
displague, line 128: A similar method exists in the generated SDK: `GetActualInstance`
https://github.com/equinix/equinix-sdk-go/blob/main/services/metalv1/model_virtual_circuit.go#L96

go.mod
displague: AFAIK, the `Facility` fields in that spec should be marked as deprecated.

0xch4z: fixed now

internal/resources/metal/connection/models.go
displague: I took a similar approach here: https://github.com/equinix/docker-machine-driver-metal/pull/81/files#diff-9cd07bb5297426eece5e3a7661c840fa020615eb84ef7f2db03ccca07bf62f40R343-R363

ctreatma:
@0xch4z is this PR ready for review?  Is anything blocking it?

internal/resources/metal/connection/models.go
0xch4z: using GetAcutualInstance here now but still using that interface so I don't need to switch on a concrete type to get some fields shared between both

0xch4z:
> @0xch4z is this PR ready for review? Is anything blocking it?

yep. ready for review now

internal/resources/metal/connection/speed.go
```diff
@@ -2,32 +2,16 @@ package connection
...
 
 var (
-	mega        uint64 = 1000 * 1000
-	giga        uint64 = 1000 * mega
-	SpeedFormat        = regexp.MustCompile(`^(\d+)((M|G)bps)$`)
+	mega int64 = 1000 * 1000
+	giga int64 = 1000 * mega
 )
 
-func speedStrToUint(speed string) (uint64, error) {
```
ctreatma, line 15: It looks like we don't need this string to int conversion for speed any more because equinix-sdk-go takes a speed as a string on input.  Just to be safe we should probably keep `SpeedFormat` around for now and update the `speed` attribute in the resource schema to validate that the value matches that regexp.

0xch4z: done

internal/resources/metal/connection/resource.go
```diff
@@ -4,17 +4,15 @@ import (
...
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
 	"github.com/equinix/terraform-provider-equinix/internal/framework"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/packethost/packngo"
```
ctreatma, line 15: It looks like there are still some things in this file and (possibly other files in this package) that are still using `packngo`.  It appears to be working fine, so this may be more of a nit.

0xch4z: oof, good catch. all fixed

ctreatma:
Missed this in my review:

> Updating "redundancy" forces recreate (this goes along with the
previous change to force new on "vlans" changes, if a new vlan/vrf
cannot be added then switching between primary and redundant should
not be allowed)

@0xch4z is this a reflection of API behavior?  That is, if I'm using an old version of the provider, I will get an API error that I can't change redundancy and I'll have to taint the resource to make terraform recreated it?

0xch4z:
> Missed this in my review:
> 
> > Updating "redundancy" forces recreate (this goes along with the
> > previous change to force new on "vlans" changes, if a new vlan/vrf
> > cannot be added then switching between primary and redundant should
> > not be allowed)
> 
> @0xch4z is this a reflection of API behavior? That is, if I'm using an old version of the provider, I will get an API error that I can't change redundancy and I'll have to taint the resource to make terraform recreated it?

so the API does allow "redundancy" to be updated, but I thought it went along with our change because if redundancy changes on a shared connection, the "vrfs"/"vlans" will need to be updated to be valid, which would cause a recreate, so I thought these go hand and hand. I can revert that if needed

ctreatma:
It looks like the forcenew for redundancy was removed from this PR at some point.  I'll update the PR description to remove the reference to that change and then merge; since the API allows changing the redundancy I think we should continue to allow it in the provider for now.  If if turns out that there really is no way to change it, the forcenew change can be introduced at that time and could arguably be seen as a bugfix rather than a breaking change.

PR #610, opened 2024-03-04
feat: push connection speed validation down to the API

### LINKED ISSUE #609: TestSpeedConversion test is failing on 100gbps int/str mismatch
The `TestSpeedConversion` test is failing on 100gbps int/str mismatch. This was introduced in [#605](https://github.com/equinix/terraform-provider-equinix/pull/605#discussion_r1509591905) while addressing #595 :

```
=== RUN   TestSpeedConversion
    speed_test.go:26: Expected error converting invalid speed string to uint, got: 100000000000
    speed_test.go:31: Expected error converting invalid speed uint to string, got: 100Gbps
--- FAIL: TestSpeedConversion (0.00s)
```

ctreatma:
Previously, we maintained a list of valid connection speed values in the terraform provider in order to steer users towards selecting a valid connection speed.  As shown by #595, validating values in the provider has the potential to break when the list of connection speed values supported by the API changes.

This updates the code that converts speed values between integer and string representations (i.e., the code that knows that `"50Mbps"` is the same as `500000000`) to convert (roughly) arbitrary values.  This means that new values are immediately "supported" by the provider, but it also means that invalid values will not be flagged until apply.

The test change in this PR is sufficient to resolve the issue introduced by merging #605 without any other change to provider behavior.

Closes #609

### DIFF
--- a/internal/resources/metal/connection/datasource_schema.go
+++ b/internal/resources/metal/connection/datasource_schema.go
@@ -50,7 +50,7 @@ func dataSourceSchema(ctx context.Context) schema.Schema {
 				Computed:    true,
 			},
 			"speed": schema.StringAttribute{
-				Description: "Port speed. Required for a_side connections. Values will be in the format '<number>Mbps' or '<number>Gpbs', for example '100Mbps`, '50Gbps'",
+				Description: "Connection speed - Values will be in the format '<number>Mbps' or '<number>Gpbs', for example '100Mbps`, '50Gbps', etc.",
 				Computed:    true,
 			},
 			"description": schema.StringAttribute{
--- a/internal/resources/metal/connection/datasource_schema.go
+++ b/internal/resources/metal/connection/datasource_schema.go
@@ -50,7 +50,7 @@ func dataSourceSchema(ctx context.Context) schema.Schema {
 				Computed:    true,
 			},
 			"speed": schema.StringAttribute{
-				Description: fmt.Sprintf("Port speed. Possible values are %s", allowedSpeedsString()),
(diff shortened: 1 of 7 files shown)


internal/resources/metal/connection/datasource_schema.go
```diff
@@ -50,7 +50,7 @@ func dataSourceSchema(ctx context.Context) schema.Schema {
 				Computed:    true,
 			},
 			"speed": schema.StringAttribute{
-				Description: fmt.Sprintf("Port speed. Possible values are %s", allowedSpeedsString()),
+				Description: fmt.Sprintf("Port speed. Required for a_side connections. Values will be in the format '<number>Mbps' or '<number>Gpbs'"),
```
displague, line 53: The .md should reflect the change. Perhaps we could give some examples speeds in the description.

0xch4z (approved):

internal/resources/metal/connection/datasource_schema.go
ctreatma: I updated the docs, but there was already some drift between the description in docs and the description in code; we probably should pick one and use it in both places, but which one to pick is an exercise left to the reviewer.

ctreatma: For comparison:
- [description in resource docs](https://github.com/equinix/terraform-provider-equinix/pull/610/files#diff-6ed6746523e1ea4d3f4fb0b43c3e840c082c07d57be197e2a4bbc269b311ad84R148)
- [description in resource code](https://github.com/equinix/terraform-provider-equinix/pull/610/files#diff-cf57b498cae31d4a345adc8449c2308c16e2e5570703a97e49122c71f43e112cR97)

PR #622, opened 2024-03-14
chore: refactor networkedge to Internal/

displague:
Moves the Network Edge resources and data sources out of the catch-all `equinix/`
package into one package per resource under `internal/resources/networkedge/`,
continuing the package split tracked in #106.

This is a pure refactor: no schema, behavior, or documentation changes. The
provider still registers exactly the same resource and data source names.

## What moves

| Terraform name | New package |
| --- | --- |
| `equinix_network_account` (data) | `internal/resources/networkedge/account` |
| `equinix_network_acl_template` | `internal/resources/networkedge/acl_template` |
| `equinix_network_bgp` | `internal/resources/networkedge/bgp` |
| `equinix_network_device` (+ data) | `internal/resources/networkedge/device` |
| `equinix_network_device_link` | `internal/resources/networkedge/device_link` |
| `equinix_network_device_software` (data) | `internal/resources/networkedge/device_software` |
| `equinix_network_device_type` (data) | `internal/resources/networkedge/device_type` |
| `equinix_network_file` | `internal/resources/networkedge/file` |
| `equinix_network_platform` (data) | `internal/resources/networkedge/platform` |
| `equinix_network_ssh_key` | `internal/resources/networkedge/ssh_key` |
| `equinix_network_ssh_user` | `internal/resources/networkedge/ssh_user` |

Each package exports `Resource()` and/or `DataSource()` in place of the old
package-private `resourceNetworkX()` / `dataSourceNetworkX()` constructors, and
`equinix/network_edge_provider_maps.go` wires them up.

## Test packages

`internal/acceptance` imports the `equinix` package in order to build the
provider under test. That makes an acceptance test an import cycle if it lives
in the same package as the resource it exercises:

```
equinix -> internal/resources/networkedge/device
        -> [test] internal/acceptance -> equinix
```

The Network Edge acceptance tests therefore move into external `<pkg>_test`
packages, which Go excludes from the cycle. This is the same arrangement the
Fabric resource packages already use. Unit tests that need package-internal
symbols stay in the internal test package alongside the resource.

The shared device / ACL / SSH-key / SSH-user config templates that the
acceptance tests build on now live in a `templates_acc_test.go` file in each
package that needs them. The four hand-written `copyMap` helpers those tests
relied on are replaced by the standard library's generic `maps.Clone`.

## Sweepers

Sweepers move out of test-file `init()` functions into a `sweeper.go` per
package, registered from `internal/sweep/services/networkedge.go` — matching the
existing Fabric sweeper layout.

They use the shared helpers in `internal/sweep` rather than anything Network
Edge specific:

- `sweep.IsSweepableTestResource` — the standard `tfacc` name-prefix check.
- `sweep.GetConfig` — added here. There is nothing product-specific about
  sweeper credentials, so this takes any of an API token, a client ID and secret
  pair, or an STS token exchange scope and subject token, and defaults
  `StsBaseURL` to `config.DefaultStsBaseURL`. `GetConfigForFabric` is left alone
  for #1047 to consolidate.

The now-unused `sharedConfigForRegion` and `isSweepableTestResource` helpers are
removed from `equinix/equinix_sweeper_test.go`.

## Not in this PR

- The `templates_acc_test.go` helpers are duplicated across the `device`,
  `device_link`, and `bgp` packages, because Go test files cannot be shared
  between packages. Hoisting them into a shared (non-test) helper package is
  worth doing, but is easier to review separately from the move.
- `testAccNetworkDevice` guards `order_reference` on the presence of
  `device-purchase_order_number`. That is carried over verbatim from
  `equinix/resource_network_device_acc_test.go` and is not changed here.
- `internal/comparisons` still has its own unexported `copyMap`, which
  `maps.Clone` would replace. The Network Edge test copies of that helper are
  gone (see below), but touching `internal/comparisons` is out of scope here.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...`, and `golangci-lint run` are
clean. Acceptance tests are unchanged in content and have not been run.


displague:
The best way to advance this PR will be to split out the helper refactorings into separate (easy to review) PRs. This will prevent the creeping in of new code with the patterns we want to move away from.

displague:
@ctreatma @equinix/governor-ne-network-edge-engineering 

I updated this refactor branch and local `go build ./...` and `go test -v ./...` are not complaining (no more import cycles in the branch!)

The biggest trick in moving the `init` function in `internal/acceptance/acceptance.go` to `_acc_test.go`.


There are a few more rounds of cleanup that could happen to make this PR easier to review:
- removing/replacing the deprecated helpers in a separate PR
- unifying the `template_acc_test.go` helpers which were copy/pasted around to avoid needing a separate package of helpers (or further polluting `acceptance` as we have done with one or two shared Metal config templates)
- removing the unused code (this is now required by the linter, along with other linting changes). In particular, the `template_acc_test.go` copy/paste has unused functions copied around for the previously mentioned reasons.
- revisiting the commit log (I expected after a rebase that some of the commits would have been removed entirely, this was not the case)

internal/nprintf/nprintf.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // NPrintf is a helper function to replace placeholders in a string with values from a map
-func NPrintf(format string, params map[string]interface{}) string {
+func Nprintf(format string, params map[string]interface{}) string {
```
displague, line 10: This rename was unintentional but consistent other than the function comment.

internal/comparisons/comparisons_test.go
```diff
@@ -27,7 +27,7 @@ func TestSubsets_negative(t *testing.T) {
 	assert.False(t, result, "Given strings were found")
 }
 
-func TestIsEmpty(t *testing.T) {
+func IsEmpty(t *testing.T) {
```
displague, line 30: This should be renamed back. introduced through overzealous search/replace

equinix/provider_test.go
```diff
@@ -124,26 +123,6 @@ func (t *testAccConfig) build() string {
 	return t.config
 }
 
-// nprintf returns a string with all the placeholders replaced by the values from the params map
-//
-// Deprecated: nprintf is shared between NE resource tests and has been
-// centralized ahead of those NE resources moving to separate packages.
-// Use github.com/equinix/terraform-provider-equinix/internal/nprintf.NPrintf instead
-func nprintf(format string, params map[string]interface{}) string {
```
displague, line 217: swapping out the deprecated functions for their `comparisons` copies could be done in a separate PR ahead of this work

displague: In anycase, this commit message should be changed from "refactor: move nprintf to separate package" to "refactor: replace deprecated functions with comparisons pkg equivalents"

internal/converters/converters.go
```diff
@@ -50,13 +50,13 @@ func ToLowerIf(v interface{}) string {
 	return strings.ToLower(v.(string))
 }
 
-// from https://stackoverflow.com/a/45428032
-func Difference(a, b []string) []string {
-	mb := make(map[string]struct{}, len(b))
+// Difference returns the elements in `a` that aren't in `b`.
```
displague, line 53: making `Difference` use generics doesn't have to be done in this PR.

equinix/provider.go
```diff
@@ -205,19 +204,6 @@ func configureProvider(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 	return &config, nil
 }
 
-// stringsFound returns true if all strings in source are found in target
-// Deprecated: stringsFound is shared between provider, tests, and resources
-// and is being relocated for package refactoring.
-// Use github.com/equinix/terraform-provider-equinix/internal/comparisons.Subsets
-func stringsFound(source []string, target []string) bool {
```
displague, line 200: This commit could be renamed from "refactor: move stringsFound and slicesMatch to comparisons package" to "refactor: remove stringsFound and slicesMatch (use comparisons pkg)"

equinix/resource_metal_device.go
```diff
@@ -9,10 +9,10 @@ import (
 	"path"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	"time"
-
```
displague, line 15: This commit, "refactor: replace isStringInSlice with slices.Contains" could be merged with the previous commit. There are only minor fmt'ing changes introduced.

displague:
#720 should be merged ahead of this so this PR can rebase and apply the sweeper related changes from that PR

displague:
After the latest rebase to pickup sweeper changes, and some changes to use SetMap (and a diag equivalent) for which more will be needed, there are new errors:

```
data_source_acc_test.go:25: Test validation error: TestStep 1/1 validation error: Providers must be specified at the TestCase level, or in all TestStep, or in TestStep.ConfigDirectory or TestStep.ConfigFile
```

I'm wondering if the init() calls in acceptance_test.go are not initializing the provider list correctly for testing.

internal/acceptance/acceptance_acc_test.go
```diff
@@ -0,0 +1,22 @@
...
+	TestAccProvider = equinix.Provider()
+	TestAccProviders = map[string]*schema.Provider{
+		"equinix": TestAccProvider,
+	}
+	TestExternalProviders = map[string]resource.ExternalProvider{
+		"random": {
+			Source: "hashicorp/random",
+		},
+	}
+	TestAccFrameworkProvider = provider.CreateFrameworkProvider(version.ProviderVersion).(*provider.FrameworkProvider)
```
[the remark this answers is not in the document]
displague: Something like this is needed to prevent the problems I initially encountered with this PR, which left it in draft. I believe a better approach was found in other resource sets. ( @ctreatma )

internal/converters/converters.go
displague: Agreed — reverted. `internal/converters/converters.go` is now untouched by this PR, so `Difference` keeps its `[]string` signature. (It currently has no callers in the repo, so switching it to generics is best done alongside a use for it.)

internal/acceptance/acceptance_acc_test.go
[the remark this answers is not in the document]
displague: Dropped this approach. `init()` stays where it was in `internal/acceptance/acceptance.go`, and `internal/acceptance/acceptance_acc_test.go` is gone — that file only ran during `go test` of the `acceptance` package itself, which left `TestAccProviders` nil for every other package and produced `Providers must be specified at the TestCase level`.

The import cycle is avoided the way the Fabric resource packages already do it: the Network Edge acceptance tests moved into external `<pkg>_test` packages, which Go excludes from the cycle. Unit tests that need package-internal symbols stay in the internal test package.

internal/resources/networkedge/device/templates_acc_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+func (t *testAccConfig) withSSHKey() *testAccConfig {
+	t.config += testAccNetworkDeviceSSHKey(t.ctx)
+	return t
+}
+
+func (t *testAccConfig) withSSHUser() *testAccConfig {
+	t.config += testAccNetworkDeviceUser(t.ctx)
+	return t
+}
```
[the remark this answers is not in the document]
displague: Acknowledged, and noted in the PR description as follow-up work. Go test files can't be shared across packages, so removing the duplication means hoisting these builders into a non-test helper package. That's a behavioral change to how the templates are built, so it's easier to review on its own than mixed into the file move.

internal/resources/networkedge/device/templates_acc_test.go
```diff
@@ -0,0 +1,461 @@
...
+  account_number        = data.equinix_network_account.test.number
+  version               = "%{device-version}"
+  core_count            = %{device-core_count}`, ctx)
+	if _, ok := ctx["device-purchase_order_number"]; ok {
+		config += nprintf.Nprintf(`
+  purchase_order_number = "%{device-purchase_order_number}"`, ctx)
+	}
+	if _, ok := ctx["device-purchase_order_number"]; ok {
+		config += nprintf.Nprintf(`
+  order_reference       = "%{device-order_reference}"`, ctx)
```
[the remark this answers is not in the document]
displague: This is pre-existing behavior, not introduced here — the same guard is in `equinix/resource_network_device_acc_test.go` on `main`. The templates were carried over verbatim so the move stays reviewable as a move; worth fixing, but separately.

internal/resources/networkedge/device_link/templates_acc_test.go
```diff
@@ -0,0 +1,461 @@
...
+  account_number        = data.equinix_network_account.test.number
+  version               = "%{device-version}"
+  core_count            = %{device-core_count}`, ctx)
+	if _, ok := ctx["device-purchase_order_number"]; ok {
+		config += nprintf.Nprintf(`
+  purchase_order_number = "%{device-purchase_order_number}"`, ctx)
+	}
+	if _, ok := ctx["device-purchase_order_number"]; ok {
+		config += nprintf.Nprintf(`
+  order_reference       = "%{device-order_reference}"`, ctx)
```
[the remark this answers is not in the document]
displague: This is pre-existing behavior, not introduced here — the same guard is in `equinix/resource_network_device_acc_test.go` on `main`. The templates were carried over verbatim so the move stays reviewable as a move; worth fixing, but separately.

internal/resources/networkedge/bgp/templates_acc_test.go
```diff
@@ -0,0 +1,461 @@
...
+  notifications         = %{device-notifications}
+  term_length           = %{device-term_length}
+  account_number        = data.equinix_network_account.test.number
+  version               = "%{device-version}"
+  core_count            = %{device-core_count}`, ctx)
+	if _, ok := ctx["device-purchase_order_number"]; ok {
+		config += nprintf.Nprintf(`
+  purchase_order_number = "%{device-purchase_order_number}"`, ctx)
+	}
+	if _, ok := ctx["device-purchase_order_number"]; ok {
```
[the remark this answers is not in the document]
displague: This is pre-existing behavior, not introduced here — the same guard is in `equinix/resource_network_device_acc_test.go` on `main`. The templates were carried over verbatim so the move stays reviewable as a move; worth fixing, but separately.

displague:
Rebased onto `main` and reworked. The PR description has been rewritten to match what the branch actually does now; the short version of what changed:

- **The helper-refactor commits are gone.** `main` has since landed the deprecated-helper removals (`stringsFound`/`isEmpty`/`slicesMatch` -> `internal/comparisons`) directly, so three commits from this branch collapsed to nothing on rebase. `internal/comparisons`, `internal/nprintf`, `internal/converters` and `internal/acceptance` are now untouched by this PR. In particular the accidental `NPrintf` -> `Nprintf` rename is dropped.
- **Metal is out of scope.** With Metal removed from the provider, the Metal-file edits and the Metal branches of the sweeper config are gone.
- **The moved files pick up `main`'s current content**, including the removal of the deprecated Network Edge attributes (#1008) and the repo-wide `interface{}` -> `any` change.
- **The import cycle is handled differently.** `internal/acceptance/acceptance_acc_test.go` is dropped and `init()` stays in `acceptance.go`; the Network Edge acceptance tests move to external `<pkg>_test` packages instead, matching the Fabric packages.
- **Sweeper config found a better home.** `internal/networkedge/sweep` and `internal/networkedge/testing_helpers` mirror the existing `internal/fabric/*` helpers, so nothing here depends on `internal/sweep/sweep.go` (which #1047 removes).

Series is down to 6 commits. `go build ./...`, `go vet ./...`, `go test ./...` and `golangci-lint run` are clean. Acceptance tests are unchanged in content and have not been run.

displague:
Correcting the sweeper part of my previous comment.

I had added `internal/networkedge/sweep` and `internal/networkedge/testing_helpers` mirroring the `internal/fabric/*` helpers. That's the wrong direction — per the discussion in #1047, `internal/sweep` is meant to stay as the shared cross-product baseline, and product-specific copies of it are what should be avoided. Those two packages are gone.

The Network Edge sweepers now use `internal/sweep` directly:

- `sweep.IsSweepableTestResource` — the standard `tfacc` name-prefix check, rather than a suffix allowlist.
- `sweep.GetConfig` — added in this PR. Nothing about sweeper credentials is product-specific, so it accepts any of an API token, a client ID and secret pair, or an STS token exchange scope and subject token, and defaults `StsBaseURL` to `config.DefaultStsBaseURL`. `GetConfigForFabric` is deliberately left untouched so #1047 can consolidate it without conflicting here.

The whole non-Network-Edge footprint of this PR is now:

```
 internal/sweep/services/networkedge.go | 18 ++++++++++++++++
 internal/sweep/sweep.go                | 39 ++++++++++++++++++++++++++++++++++
 internal/sweep/sweep_test.go           |  1 +
```

Also removed the four hand-copied `copyMap` helpers the acceptance tests carried around, in favour of the standard library's generic `maps.Clone`. One unexported copy remains in `internal/comparisons`; that file is otherwise untouched by this PR, so it's left for a separate change.

`go build ./...`, `go vet ./...`, `go test ./...` and `golangci-lint run` are clean.

displague:
To make the diff a little smaller, we could (in a sperate PR-set): adopt NPrintf first, map.Clone, and run goimports.

These could alternatively be handled as follow-ups since they are trivial in review.

One additional follow-up noted in the resolved comments is a refactor of the shared NE test functions (https://github.com/equinix/terraform-provider-equinix/pull/622#discussion_r2602992372).
Reducing their duplication should also help reduce the codecoverage drop we see from this refactor.

A possibly wrong conditional check for an attribute setting was also identified, which should be followed-up on: https://github.com/equinix/terraform-provider-equinix/pull/622#discussion_r2602992432

ctreatma:
> To make the diff a little smaller, we could (in a sperate PR-set): adopt NPrintf first, map.Clone, and run goimports.

I think the current diff scope is reasonable.  My main concern is confidence in the impact of the change.  A separate PR that introduces a passing acceptance test workflow would help with that.

PR #623, opened 2024-03-14
chore(ci): sweep shared connections

ctreatma:
The sweep task fails occasionally in CI because it attempts to delete VLANs that are attached to virtual circuits that belong to shared connections.  Virtual circuits on shared connections are API-managed and can only be deleted by deleting the shared connection, but we did not have a sweeper to clean up dangling shared connections after tests.

This adds a sweeper to clean up shared connections.  New shared connection types were recently added in the API, so the sweeper checks that a connection is not "dedicated" rather than checking that it is "shared."  A similar update was made to the virtual circuit sweeper.  The VLAN sweeper has been updated to declare the new connection sweeper as a dependency to ensure that shared connections have been cleaned up before we try to clean up VLANs.


ocobles:
linking the change for VC sweeper https://github.com/equinix/terraform-provider-equinix/pull/584/files

PR #642, opened 2024-04-04
fix: avoid inconsistent state errors for vlans

### LINKED ISSUE #633: [Bug]: equinix_metal_vlan breaks when metro is upper case or description is not specified
### Terraform Version

Terraform 1.7.4 on darwin_arm64

### Equinix Provider Version

1.34.0

### Effected Terraform Resources

equinix_metal_vlan

### Terraform Config Files

```terraform
resource "equinix_metal_vlan" "vlans" {
  count      = 2
  project_id = local.project_id
  metro      = "TY"
}
```

### Debug Output

```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to equinix_metal_vlan.vlans[1], provider "provider[\"registry.terraform.io/equinix/equinix\"]" produced an unexpected new value: .metro: was cty.StringVal("TY"), but now cty.StringVal("ty").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Panic Output

_No response_

### Expected Behavior

No error

### Actual Behavior

Received error

### Steps to Reproduce

Specify metro in uppercase as it's defined in the documentation at 
List of metros: https://deploy.equinix.com/developers/docs/metal/locations/metros/#metros-quick-reference

Works fine in provider 1.24.0. If a put metro in lower case, then I get the error:
```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to equinix_metal_vlan.vlans[0], provider "provider[\"registry.terraform.io/equinix/equinix\"]" produced an unexpected new value: .description: was null, but now cty.StringVal("").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

Could workaround above bug by adding `description = ""` to the equinix_metal_vlan resource, but according to the docs at https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_vlan#description , it's supposed to be an optional attribute.

ctreatma:
When the equinix_metal_vlan resource was migrated (#578) from terraform-plugin-sdk/v2 (SDKv2) to terraform-plugin-framework (framework), some unintended behavior was introduced around the description and metro attributes.

For equinix_metal_vlan resources under SDKv2, description was an optional attribute and metro was case insensitive, but after migration from SDKv2 to framework, the description was set to `""` when the attribute was omitted, and metro was required to be lower-case to avoid terraform errors.

This updates the VLAN model parsing function so that:
- if there is already metro value configured in state and it is equal to the value received from the API or from the config, ignoring case, then the value that is already in state is used instead of the value from the API
- if there is no description in the API response, the description property is omitted entirely instead of being set to an empty string

`TestAccMetalVlan_NoDescription` has been added to validate the resource behavior when the description is omitted and `TestAccMetalVlan_metro` has been updated to validate behavior when metro is specified in uppercase.

Fixes #633

### DIFF
--- a/internal/resources/metal/vlan/models.go
+++ b/internal/resources/metal/vlan/models.go
@@ -2,10 +2,12 @@ package vlan
 
 import (
 	"context"
+	"fmt"
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/packethost/packngo"
-	"strings"
 )
 
 type DataSourceModel struct {
@@ -19,12 +21,15 @@ type DataSourceModel struct {
 	AssignedDevicesIds types.List   `tfsdk:"assigned_devices_ids"`
 }
 
-func (m *DataSourceModel) parse(vlan *packngo.VirtualNetwork) diag.Diagnostics {
+func (m *DataSourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostics) {
 	m.ID = types.StringValue(vlan.ID)
 	m.VlanID = types.StringValue(vlan.ID)
-	m.Description = types.StringValue(vlan.Description)
 	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
 	m.Facility = types.StringValue("")
+
+	if vlan.Description != "" {
+		m.Description = types.StringValue(vlan.Description)
+	}
 
 	if vlan.Project.ID != "" {
 		m.ProjectID = types.StringValue(vlan.Project.ID)
@@ -36,7 +41,14 @@ func (m *DataSourceModel) parse(vlan *packngo.VirtualNetwork) diag.Diagnostics {
 	}
 
 	if vlan.Metro != nil {
-		m.Metro = types.StringValue(strings.ToLower(vlan.Metro.Code))
+		if m.Metro.IsNull() {
+			m.Metro = types.StringValue(vlan.Metro.Code)
+		} else if !strings.EqualFold(m.Metro.ValueString(), vlan.Metro.Code) {
+			d.AddWarning(
+				"unexpected value for metro",
+				fmt.Sprintf("expected vlan %v to have metro %v, but metro was %v",
+					m.ID, m.Metro, vlan.Metro.Code))
+		}
 	}
 
 	deviceIds := make([]types.String, 0, len(vlan.Instances))
@@ -56,11 +68,14 @@ type ResourceModel struct {
 	Description types.String `tfsdk:"description"`
 }
 
-func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) diag.Diagnostics {
+func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostics) {
 	m.ID = types.StringValue(vlan.ID)
-	m.Description = types.StringValue(vlan.Description)
 	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
 	m.Facility = types.StringValue("")
+
+	if vlan.Description != "" {
+		m.Description = types.StringValue(vlan.Description)
+	}
 
 	if vlan.Project.ID != "" {
 		m.ProjectID = types.StringValue(vlan.Project.ID)
@@ -72,8 +87,14 @@ func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) diag.Diagnostics {
 	}
 
 	if vlan.Metro != nil {
-		m.Metro = types.StringValue(strings.ToLower(vlan.Metro.Code))
+		if m.Metro.IsNull() {
+			m.Metro = types.StringValue(vlan.Metro.Code)
+		} else if !strings.EqualFold(m.Metro.ValueString(), vlan.Metro.Code) {
+			d.AddError(
+				"unexpected value for metro",
+				fmt.Sprintf("expected vlan %v to have metro %v, but metro was %v",
+					m.ID, m.Metro, vlan.Metro.Code))
+		}
 	}
-
 	return nil
 }
--- a/internal/resources/metal/vlan/resource_test.go
+++ b/internal/resources/metal/vlan/resource_test.go
@@ -80,7 +80,7 @@ func TestAccMetalVlan_metro(t *testing.T) {
 			{
 				Config: testAccCheckMetalVlanConfig_metro(rs, strings.ToUpper(metro), "tfacc-vlan"),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					testAccCheckMetalSameVlan("equinix_metal_vlan.foovlan", &vlan),
 					resource.TestCheckResourceAttr(
 						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
 					resource.TestCheckResourceAttr(
@@ -100,7 +100,7 @@ func TestAccMetalVlan_metro(t *testing.T) {
 			{
 				Config: testAccCheckMetalVlanConfig_metro(rs, strings.ToLower(upperMetro), "tfacc-vlan"),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					testAccCheckMetalSameVlan("equinix_metal_vlan.foovlan", &vlan),
 					resource.TestCheckResourceAttr(
 						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
 					resource.TestCheckResourceAttr(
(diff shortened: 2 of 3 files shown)


internal/resources/metal/vlan/models.go
```diff
@@ -56,12 +68,15 @@ type ResourceModel struct {
...
 
-func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) diag.Diagnostics {
+func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostics) {
 	m.ID = types.StringValue(vlan.ID)
-	m.Description = types.StringValue(vlan.Description)
 	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
 	m.Facility = types.StringValue("")
 
+	if vlan.Description != "" {
+		m.Description = types.StringValue(vlan.Description)
```
displague, line 77: Does this effectively prevent TF from recognizing an API cleared description?

Terraform initially set and continues to want: "foo"
API was updated to: ""
Terraform ignores the diff?

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,10 +77,64 @@ func TestAccMetalVlan_metro(t *testing.T) {
...
+					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "metro", metro),
+				),
+			},
+			{
+				Config: testAccCheckMetalVlanConfig_metro(rs, upperMetro, "tfacc-vlan"),
+				Check: resource.ComposeTestCheckFunc(
```
displague, line 92: Is there a check we can add to ensure that the resource was effectively recreated when the metro changed?

internal/resources/metal/vlan/models.go
ctreatma: VLAN doesn't have an update endpoint according to the spec & both SDKs (all attributes of VLAN resources are marked as requiring replacement, and have been for quite a while).  I added a test, though, to confirm that the description is removed; if we ever add support for updating VLAN attributes, that will help confirm that description continues to work as expected.

internal/resources/metal/vlan/resource_test.go
ctreatma: The model is written to throw an error if the metro received from the API is different from the one in state (other than casing): https://github.com/equinix/terraform-provider-equinix/pull/642/files#diff-8f6d46b17b852aeb7d85b49d7e72ea5359e1345953070c31a27b9264aa683f42R93-R96

If we _weren't_ recreating the VLAN, we should hit that error and fail the test.

ctreatma: Thinking about this more:  we should probably update the test to validate that the VLAN _isn't_ recreated when we change the metro casing

ctreatma: Test has been updated to check that the VLAN ID stays the same when the metro casing changes; I validated by temporarily removing the CaseInsensitiveString plan modifier from the resource (so that casing changes are seen as requiring an apply) and confirming that the test fails.

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,6 +77,89 @@ func TestAccMetalVlan_metro(t *testing.T) {
...
+		PreCheck:                 func() { acceptance.TestAccPreCheckMetal(t) },
+		ExternalProviders:        acceptance.TestExternalProviders,
+		ProtoV5ProviderFactories: acceptance.ProtoV5ProviderFactories,
+		CheckDestroy:             testAccMetalVlanCheckDestroyed,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckMetalVlanConfig_NoDescription(rs, metro),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					resource.TestCheckNoResourceAttr(
```
ctreatma, line 122: This test checks that we don't get "inconsistent state" errors when description is omitted in the initial config

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,6 +77,89 @@ func TestAccMetalVlan_metro(t *testing.T) {
...
+					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "metro", metro),
+				),
+			},
+			{
+				Config: testAccCheckMetalVlanConfig_NoDescription(rs, metro),
+				Check: resource.ComposeTestCheckFunc(
```
ctreatma, line 155: This test checks that we don't get "inconsistent state" errors when description is included in the initial config and then deleted later.

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,6 +77,89 @@ func TestAccMetalVlan_metro(t *testing.T) {
 						"equinix_metal_vlan.foovlan", "metro", metro),
 				),
 			},
+			{
+				Config:   testAccCheckMetalVlanConfig_metro(rs, strings.ToUpper(metro), "tfacc-vlan"),
+				PlanOnly: true,
```
ctreatma, line 82: This test step confirms that there are no changes to apply after changing the metro casing

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,6 +77,89 @@ func TestAccMetalVlan_metro(t *testing.T) {
...
+			},
+			{
+				Config: testAccCheckMetalVlanConfig_metro(rs, upperMetro, "tfacc-vlan"),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("equinix_metal_vlan.foovlan", plancheck.ResourceActionDestroyBeforeCreate),
```
ctreatma, line 88: This plan check confirms that this test step will destroy and recreate the VLAN because there was a material change to the metro from Silicon Valley to Dallas

internal/resources/metal/vlan/resource_test.go
```diff
@@ -63,6 +77,89 @@ func TestAccMetalVlan_metro(t *testing.T) {
...
+					testAccCheckMetalVlanExists("equinix_metal_vlan.foovlan", &vlan),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "description", "tfacc-vlan"),
+					resource.TestCheckResourceAttr(
+						"equinix_metal_vlan.foovlan", "metro", upperMetro),
+				),
+			},
+			{
+				Config:   testAccCheckMetalVlanConfig_metro(rs, strings.ToLower(upperMetro), "tfacc-vlan"),
+				PlanOnly: true,
```
ctreatma, line 101: This test step confirms that there are no changes to apply after changing the metro from `DA` to `da`

internal/resources/metal/vlan/resource_test.go
cprivitere: Any reason you don't just use upperMetro here?

internal/resources/metal/vlan/resource_test.go
cprivitere: And here you could use metro, right?

cprivitere (dismissed):
lgtm, but I'm not 100% up to speed on how these tests work. So maybe wait for @displague or @ocobles

cprivitere:
```
=== NAME  TestAccMetalConnection_sharedRedundantVrf
    testing_new.go:91: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Failed to delete Metal Connection d18ec3fc-e049-42a6-8f2a-7a248d0af2d1
        
        Timeout waiting for connection to be deleted
--- FAIL: TestAccMetalConnection_sharedRedundantVrf (79.21s)
FAIL
coverage: 67.9% of statements
FAIL	github.com/equinix/terraform-provider-equinix/internal/resources/metal/connection	79.228s
```

displague:
This one is relevant but seems to have failed on availability

> --- FAIL: TestAccMetalPort_hybridBondedVxlan (2.70s)

internal/resources/metal/vlan/resource_test.go
displague: `metro` and `upperMetro` are being used in different ways, one is to use a different case as a starting point before converting case and the other is to have different metros to ensure that changing the metro field has the intended effect.

internal/resources/metal/vlan/resource_test.go
ctreatma: I can see how this test is hard to reason about; in the latest commit I tried changing the names of the `upperMetro` and `metro` variables in an attempt to clarify intent/usage.  Let me know if that helped or no.

ctreatma:
@displague @displague I pushed up a minor change to the tests to try to make them more legible.  The e2e tests are flaky; I re-ran them 3 more times in CI, but there were some failures each time.  You can look at the output of each run and see the failures to decide if they're related to this change or not: https://github.com/equinix/terraform-provider-equinix/actions/runs/8604589925/job/23586041155?pr=642

PR #646, opened 2024-04-15
fix: Upgrading provider_connection_id schema to computed

srushti-patl:
- Marking provider_connection_id schema to computed in fabric_connection_resource


thogarty (approved):
LGTM. Solves customer issue for `provider_connection_id` not populating properly.

PR #650, opened 2024-04-16
docs: Update Fabric data_source attribute selection documentation

thogarty:
* For each Fabric data_source add attribute selection to documentation with output block examples


srushti-patl (approved):
LGTM

PR #654, opened 2024-04-22
chore: refactor nprintf

displague:
Extracted from #622


internal/sweep/sweep_test.go
```diff
@@ -1,9 +1,10 @@
 package sweep_test
 
 import (
-	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/vlan"
 	"testing"
 
+	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/vlan"
```
ctreatma, line 6: We should get something in place to enforce that imports are declared in the "correct" order, for some agreed-upon definition of correct, to avoid stylistic churn like this.

internal/nprintf/nprintf.go
```diff
@@ -0,0 +1,23 @@
+package nprintf
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// NPrintf is a helper function to replace placeholders in a string with values from a map
+func NPrintf(format string, params map[string]interface{}) string {
```
ctreatma, line 10: IMO it would be preferable to leave this in `equinix/` and deprecate it.  While there is some reduction of repetition here compared to using `fmt.Sprintf`, the repetition has the benefit of being explicit and flexible.

internal/sweep/sweep_test.go
displague: Agreed. I stalled on getting this PR consistent because there were some import changes affected but the tools I used (golangci-lint, goimports) didn't have a strong enough opinion to update the existing import blocks. Updating beyond the files affected would be preferrable in another PR or at least a separate commit.

goimports wouldn't regroup all imports ignoring existing arbitrary white-space. Since it honors the existing gaps the desired 3 group import stanzas (built-in, external, internal) was not achieved.

internal/nprintf/nprintf.go
displague: The nprintf extraction was made to simplify the changes and review needed for https://github.com/equinix/terraform-provider-equinix/pull/622 where additional packages and functions were being moved out of `equinix/` as part of #106 .

ctreatma: I think in this case I'd usually lean towards duplicating (not moving) nprintf into a separate package and marking the NPrintf function as deprecated in the same PR that moves the NE resources.  Duplicating the function enables us to move only the NE resources without touching anything else, and doing that in one PR makes it clear that this is tech debt we're taking on to accomplish a specific goal.

Given our CODEOWNERS configuration, though, I could see having a separate PR to introduce the nprintf package; IMO a PR like that should only touch resources for which we are the owners, so we would introduce the nprintf package as a deprecated approach that exists only to facilitate code migration and update the ECX resources--without moving them, since they're going away soon--to demonstrate that nprintf package is a drop-in replacement for the private nprintf function.  A follow-up PR could move the NE resources and it would be up to the NE team to review and adopt that PR.

displague: Duplicated and deprecated. (removed: 6490d46afdac4feedd0cb0975771145154f76ee7 and e7708e9175aa5ca3bae2abbb7586d287bb0168ae, the first may be cherry-picked later)

internal/nprintf/nprintf.go
```diff
@@ -0,0 +1,23 @@
+package nprintf
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// NPrintf is a helper function to replace placeholders in a string with values from a map
```
ctreatma, line 9:
```suggestion
// NPrintf is a helper function to replace placeholders in a string with values from a map
//
// Deprecated: nprintf.NPrintf exists to support migrating resources and data source out
// of the equinix package and into separate packages under `internal/resources`.  For new
// code, prefer to use fmt.Sprintf which enables more explicit and flexible formatting at the
// potential cost of higher repetition.
```
Nit: I lean towards deprecating this function entirely, in which case we should discouraged new usage, but that could be a separate follow-up discussion.

ctreatma (approved):

PR #661, opened 2024-04-30
fix: Updating CreateConnection method for passing additional_info in Updating Create Connection method in Fabric connection Resource

srushti-patl:
- Updated resourceFabricConnectionCreate method by passing additional_info parameters only when it's required


thogarty (approved):
LGTM! Ship it.

PR #672, opened 2024-05-09
fix: Updating Virtual device to service profile connections type in fabric connection resource doc

srushti-patl:
- Updated connection type in following examples in Fabric Connection Resources Doc:
- vd2azure connection
- vd2azure redundant connection


thogarty (approved):
LGTM.

PR #676, opened 2024-05-10
chore: tell renovate to run go mod tidy after updating

ctreatma:
We periodically have to manually run `go mod tidy` to clean up our dependencies because Renovate doesn't run that by default for its update PRs.  This runs one last manual tidy and updates the Renovate configuration to tell Renovate to clean up after itself from now on.

Reference: https://docs.renovatebot.com/configuration-options/#postupdateoptions

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,6 @@ require (
 
 require (
 	cloud.google.com/go v0.112.0 // indirect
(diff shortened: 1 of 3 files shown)


displague (approved):

displague:
I've noticed this problem and fix-up pattern before. Glad to see it addressed.

PR #677, opened 2024-05-13
fix: Updating connectionRedundancyTerraformToGo fucntion

srushti-patl:
- Setting Redundancy group only when it's not empty


thogarty (approved):
LGTM.

PR #678, opened 2024-05-16
feat: add `data "equinix_fabric_cloud_routers"` data source to product suite

thogarty:
* Cloud router data source only supports retrieving cloud router by UUID which isn't as useful when UUID isn't known
* New data_source supports search filters/pagination/sorting to retrieve a very specific Cloud Router or a list of Cloud Routers
* Supporting API documentation: https://app.swaggerhub.com/apis/equinix-api/fabric/4.13#/Cloud%20Routers/searchCloudRouters


srushti-patl (approved):
LGTM

equinix/data_source_fabric_cloud_router.go
```diff
@@ -38,3 +140,15 @@ func dataSourceFabricCloudRouterRead(ctx context.Context, d *schema.ResourceData
 	d.SetId(uuid)
 	return resourceFabricCloudRouterRead(ctx, d, meta)
 }
+
+func dataSourceFabricGetCloudRouters() *schema.Resource {
```
ctreatma, line 144: I understand that there may be some shared code in play here, but it would be preferable to put the new data source in its own file.

equinix/resource_fabric_cloud_router.go
```diff
@@ -307,16 +308,144 @@ func resourceFabricCloudRouterRead(ctx context.Context, d *schema.ResourceData,
...
+		}
+		if property, ok := pageMap["property"]; ok {
+			sortItem.SetProperty(fabricv4.CloudRouterSortBy(property.(string)))
+		}
+		sortCriteria[index] = sortItem
+	}
+	return sortCriteria
+}
+
+func resourceFabricCloudRoutersSearch(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
ctreatma, line 404: It seems like all of this new code in the resource file is only used by the new multi-result data source; in that case this code could live alongside the rest of the new data source code in, e.g., `data_source_fabric_cloud_routers.go`

equinix/resource_fabric_port.go
```diff
@@ -208,46 +208,6 @@ func FabricPortResourceSchema() map[string]*schema.Schema {
...
 
-func readFabricPortResourceSchemaUpdated() map[string]*schema.Schema {
-	sch := FabricPortResourceSchema()
-	sch["uuid"].Computed = true
-	sch["uuid"].Optional = false
-	sch["uuid"].Required = false
-	return sch
-}
-
-func readFabricPortsResponseSchema() map[string]*schema.Schema {
```
ctreatma, line 219: It looks like all of this stuff was moved into the port data source file; does that need to happen as part of introducing a new cloud routers data source?

go.sum
```diff
@@ -68,10 +68,6 @@ cloud.google.com/go/compute v1.6.0/go.mod h1:T29tfhtVbq1wvAPo0E3+7vhgmkOYeXjhFvz
 cloud.google.com/go/compute v1.6.1/go.mod h1:g85FgpzFvNULZ+S8AYq87axRKuf2Kh7deLqV/jJ3thU=
 cloud.google.com/go/compute v1.7.0/go.mod h1:435lt8av5oL9P3fv1OEzSbSUe+ybHXGMPQHHZWZxy9U=
 cloud.google.com/go/compute v1.10.0/go.mod h1:ER5CLbMxl90o2jtNbGSbtfOpQKR0t15FOtRsugnLrlU=
-cloud.google.com/go/compute v1.24.0 h1:phWcR2eWzRJaL/kOiJwfFsPs4BaKq1j6vnpZrc1YlVg=
```
ctreatma, line 71: Renovate should be taking care of `go mod tidy` for us now, so this may be taken care of with a rebase

equinix/data_source_fabric_cloud_router.go
thogarty: Solid feedback. Updated.

equinix/resource_fabric_cloud_router.go
thogarty: Agreed. The remaining changes in this file do relate to code that was already there. A refactor was made so that we can set the data response easily for a GET by UUID or a POST with search Parameters that has the same response schema as elements inside of the data list.

equinix/resource_fabric_port.go
thogarty: Nope, too much with one change. Not necessary.

Was going towards the same destination with all data source code belonging together but it should be with a different PR.

go.sum
thogarty: Done 👍

srushti-patl (approved):

srushti-patl (dismissed):

ctreatma (dismissed):

srushti-patl (approved):

ctreatma (approved):

ctreatma:
The ship may have sailed on this one, but in the future when adding a new resource or data source, I encourage you to add it in a separate, dedicated package under `internal/resources/fabric/` rather than adding more stuff to the `equinix` package.  (If you do that, I would recommend continuing to use terraform-plugin-sdk/v2 and saving the migration to terraform-plugin-framework for later.)  That will avoid adding new things that need to be refactored in the future.

thogarty:
> The ship may have sailed on this one, but in the future when adding a new resource or data source, I encourage you to add it in a separate, dedicated package under `internal/resources/fabric/` rather than adding more stuff to the `equinix` package. (If you do that, I would recommend continuing to use terraform-plugin-sdk/v2 and saving the migration to terraform-plugin-framework for later.) That will avoid adding new things that need to be refactored in the future.

@ctreatma , I'm following up with a few more so I'll keep that in mind. 👍

PR #693, opened 2024-06-04
feat: Fabric Connections SEARCH data source

thogarty:
* Refactored Fabric Connections resource/data_source/tests to internal package
* Added `data "equinix_fabric_connections"` to product suite
* Added tests and documentation for new data source
* Duplicated and decoupled Fabric Connections resource from other Fabric resource during refactor

When duplicating Fabric Connection methods into internal package for the new data source it kept having dependencies to the point where bringing the whole resource/data_source/schemas with it to internal made more sense than having a duplicate.


equinix/data_source_fabric_port_acc_test.go
```diff
@@ -1,33 +1,17 @@
...
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 )
 
-const (
-	FabricDedicatedPortEnvVar = "TF_ACC_FABRIC_DEDICATED_PORTS"
-)
-
-type EnvPorts map[string]map[string][]fabricv4.PortResponse
-
-func GetFabricEnvPorts(t *testing.T) EnvPorts {
```
ctreatma, line 20: The removal of this function is causing a couple errors in the unit test workflow: https://github.com/equinix/terraform-provider-equinix/actions/runs/9375209054/job/25812858456?pr=693#step:6:32

It looks like there are still a couple references to this function; you could update those here if you like, but if you want to minimize the diff for this PR, you could also keep this function in place and continue to use it within the `equinix` package.  If you keep it, I recommend adding a `// Deprecated: <details>` comment that directs the reader to prefer the new, equivalent function in `testing_helpers`.

internal/resources/fabric/connection/datasources.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+func dataSourceFabricConnectionRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	uuid, _ := d.Get("uuid").(string)
+	d.SetId(uuid)
+	return resourceFabricConnectionRead(ctx, d, meta)
+}
+
+func DataSourceSearch() *schema.Resource {
```
ctreatma, line 30: I'm a wary of having two datasources implemented in a single file, but that's mitigated somewhat by the fact that this is now an isolated package.

ctreatma (approved):

srushti-patl (approved):

thogarty:
@ctreatma , had to rebase to resolve conflicts. Provider.go had import conflicts when Networks Data Source was merged. This is the same with both imports in provider.go.

ctreatma (approved):

PR #699, opened 2024-06-15
fix: resource_fabric_routing_protocol.go

jkallem-equinix:
fix: resource_fabric_routing_protocol.go added more compute fields to the file resource_fabric_routing_protocol.go since  Routing Protocol is updating when no Changes are made for Routing Protocol Modules while updating Connection_Name and Bandwidth.


srushti-patl (approved):
LGTM

PR #700, opened 2024-06-18
feat: convert virtual circuit resource & datasource to equinix-sdk-go

ctreatma:
This updates the `metal_virtual_circuit` resource and data source to use `equinix-sdk-go` instead of `packngo`.  As mentioned in #402, the `packngo` SDK has been deprecated, and moving to `equinix-sdk-go` makes it possible for these resources to gain new features in the future.

As an added bonus, the affected code has also been moved to `internal/resources/metal` to align with #106.

### DIFF
--- a/equinix/resource_metal_virtual_circuit.go
+++ b/equinix/resource_metal_virtual_circuit.go
@@ -3,22 +3,21 @@ package equinix
 import (
 	"context"
 	"log"
-	"reflect"
 	"regexp"
 	"strconv"
 	"time"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	"github.com/equinix/terraform-provider-equinix/internal/converters"
 
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
-	equinix_schema "github.com/equinix/terraform-provider-equinix/internal/schema"
 
 	"github.com/equinix/terraform-provider-equinix/internal/config"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/packethost/packngo"
 )
 
 func resourceMetalVirtualCircuit() *schema.Resource {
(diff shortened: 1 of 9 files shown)


equinix/resource_metal_virtual_circuit.go
```diff
@@ -146,141 +145,173 @@ func resourceMetalVirtualCircuit() *schema.Resource {
...
 	if len(matches) == 3 {
 		connectionID = matches[1]
 		portID = matches[2]
 	} else {
-		log.Printf("[DEBUG] Could not parse connection and port ID from port href %s", vc.Port.Href.Href)
+		log.Printf("[DEBUG] Could not parse connection and port ID from port href %s", portHref)
 	}
+	errs := &multierror.Error{}
 
-	err = equinix_schema.SetMap(d, map[string]interface{}{
```
ctreatma, line 235: The more granular type mapping means we can't just look up whatever fields we want and shove whatever we get into a map; we have to know if we're working with a VLAN or VRF virtual circuit and look up only the appropriate fields based on the type.

This could be seen as a shortcoming of strict schema compliance but it's also possible this means that there should be separate resources for `vlan_virtual_circuit` and `vrf_virtual_circuit`.

equinix/resource_metal_virtual_circuit.go
```diff
@@ -146,141 +145,173 @@ func resourceMetalVirtualCircuit() *schema.Resource {
...
 		Pending: pending,
 		Target:  target,
 		Refresh: func() (interface{}, string, error) {
-			vc, _, err := client.VirtualCircuits.Get(id, nil)
+			vc, _, err := client.InterconnectionsApi.GetVirtualCircuit(ctx, id).Execute()
 			if err != nil {
 				return 0, "", err
 			}
-			return vc, string(vc.Status), nil
+			return vc, string(vc.VlanVirtualCircuit.GetStatus()), nil
```
ctreatma, line 314: This is definitely wrong; we might have gotten a `VrfVirtualCircuit` instead

equinix/resource_metal_virtual_circuit.go
```diff
@@ -289,71 +320,117 @@ func getVCStateWaiter(client *packngo.Client, id string, timeout time.Duration,
...
 		return diag.FromErr(equinix_errors.FriendlyError(err))
 	}
 
 	deleteWaiter := getVCStateWaiter(
+		ctx,
 		client,
 		d.Id(),
 		d.Timeout(schema.TimeoutDelete)-30*time.Second,
-		[]string{string(packngo.VCStatusDeleting)},
+		[]string{string(metalv1.VLANVIRTUALCIRCUITSTATUS_DELETING)},
```
ctreatma, line 428: Status checks are using the values defined in the VLAN VC schema but that's probably OK for now since the previous SDK didn't even distinguish between VC types

equinix/resource_metal_virtual_circuit.go
ctreatma: Updated to handle both VLAN and VRF virtual circuits

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -0,0 +1,459 @@
...
+
+			"vnid": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "VNID VLAN parameter, see https://metal.equinix.com/developers/docs/networking/fabric/",
+			},
+			"nni_vnid": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "Nni VLAN ID parameter, see https://metal.equinix.com/developers/docs/networking/fabric/",
```
displague, line 136: There are a few metal.equinix.com URLs referenced here. I could ignore these as migrated and a separate problem, but this is effectively 90% new code.

This link should be: https://deploy.equinix.com/developers/docs/metal/interconnections/introduction/

The same link is above and may appear elsewhere.

ctreatma: 90% is a high estimate; this is [mostly the same code](https://github.com/equinix/terraform-provider-equinix/pull/700/commits/0447951f245151a5b35c5d23f9a225ffcd4421d3#diff-a559b25bbfeecdf929180963d85161e3d214577dcc1f671f09d74e2a31f4c838) but moving it made GitHub lose the plot.

That said, I'm fully on board with fixing a couple outdated URLs while I'm in here.

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	errs := &multierror.Error{}
+
+	if connectionID != "" {
+		multierror.Append(errs, d.Set("connection_id", connectionID))
+	}
+	d.Set("port_id", portID)
+
+	if vc.VlanVirtualCircuit != nil {
+		multierror.Append(errs, d.Set("project_id", vc.VlanVirtualCircuit.Project.GetId()))
```
displague, line 279: multierror.Append is variadic: https://github.com/hashicorp/go-multierror/blob/main/append.go#L14

This could be a comma list of d.Sets.

I don't know if d.Set will include the fieldname in the error messages, like (I think) SetMap does.
We may have unnamed error messages that are hard to pin down in this list.

ctreatma: SetMap does not do anything special to the error it gets from `d.Set`.  It just passes it to `multierror.Append` as was done here.  I've since switched to `append` and `errors.Join` from the standard library.  That is also variadic but IMO the repetition is worthwhile here and reduces the chance that someone copy-pastes a line to add support for a new attribute and happens to paste it outside of the `append()` call.

displague:
Some sweeper errors persist:
```
2024/06/19 16:46:55 Sweeper Tests for region (all) ran unsuccessfully:
2024/06/19 16:46:55 	- equinix_metal_project: 1 error occurred:
	* Error deleting project DELETE https://api.equinix.com/metal/v1/projects/f096688e-166d-42f1-9959-9acf97ebc8a1: 422 cannot delete when associated to a virtual circuit 


2024/06/19 16:46:55 	- equinix_metal_vlan: Error deleting vlan DELETE https://api.equinix.com/metal/v1/virtual-networks/851f2fc7-bfc1-4a9b-b8a7-4a8ec4e125f1: 422 Cannot delete Virtual Network when associated to a virtual circuit 
FAIL	github.com/equinix/terraform-provider-equinix/internal/sweep	7.564s
FAIL
```

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -0,0 +1,459 @@
...
+		Delay:      10 * time.Second,
+		MinTimeout: 5 * time.Second,
+	}
+}
+
+func resourceMetalVirtualCircuitUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*config.Config).NewMetalClientForSDK(d)
+	needsUpdate := false
+
+	ur := metalv1.VirtualCircuitUpdateInput{}
```
displague, line 342: I want to think there is some pattern to be used across this resource that would have collapsed the `if vrf {} else vlan {}` blocks in this PR.

That could look like this:

```go
type VCCommon interface {
  GetHref() string
  GetID() string
  Set...
}
```

Where that quickly falls apart is in the custom enum inputs/outputs and the one-off differences between VirtualCircuit types, such as `int` reference vs `*int` for NNID fields.

Withing this Update CRUD, we could apply that pattern. We did so in metal-cli: https://github.com/equinix/metal-cli/blob/main/internal/virtualcircuit/update.go#L12-L22

I only mention this for posterity. Not looking for this PR to adopt this pattern.

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -0,0 +1,459 @@
...
+				ur.VrfVirtualCircuitUpdateInput.Tags = sts
+			default:
+				return diag.Errorf("garbage in tags: %s", ts)
+			}
+		}
+	}
+
+	if needsUpdate {
+		if _, _, err := client.InterconnectionsApi.UpdateVirtualCircuit(ctx, d.Id()).VirtualCircuitUpdateInput(ur).Execute(); err != nil {
+			return diag.FromErr(equinix_errors.FriendlyError(err))
```
displague, line 425: Should this be: `equinix_errors.FriendlyErrorForMetalGo`? There are a few other instances in this PR.

ctreatma: IMO it should just be `diag.FromErr(err)`

displague: I didn't check all of the uses. I was considering how we noticed FriendlyError vs. FriendlyErrorForMetalGo usage in other locations, which could potentially lead to improper 404 detection.

ctreatma: I changed the `FriendlyError` calls to remove the conversion entirely, since in one case there's no status check happening and in the other cases the error has already been (begrudgingly) converted in order to fit into the existing filtering and avoid a larger refactor.

ctreatma:
The remaining sweeper errors seem to pre-date this PR: https://github.com/equinix/terraform-provider-equinix/actions/runs/9503755717/job/26194913301#step:7:146

The connection in question does not have any virtual circuits associated, so it appears they were deleted, just not in time for the subsequent sweeper.  That's a problem that impacts our sweepers in general:  at a glance, approximately none of them wait for swept resources to disappear; they just assume that if the delete request passed, all is well.

cprivitere (approved):

displague (approved):

PR #702, opened 2024-06-18
fix: ignore HTTP errors based only on status

ctreatma:
The `equinix_errors.IgnoreHttpResponseErrors` function was introduced to support migration of Metal resources from `packngo` to `equinix-sdk-go`.  However, the check functions that were passed in to it (such as `equinix_errors.IsNotFound`) only work correctly for errors from `equinix-sdk-go` if the error is first converted to an `ErrorResponse` by calling `equinix_errors.FriendlyErrorForMetalGo(resp, err)`.

This updates `equinix_errors.IgnoreHttpResponseErrors` to only check the response status code.

### DIFF
--- a/equinix/resource_metal_device.go
+++ b/equinix/resource_metal_device.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
 	"path"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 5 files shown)


ctreatma:
Converting to draft because there is some convoluted usage of state waiters for deletions that only works if the waiter returns an error object that includes the HTTP response, and this change doesn't help with that problem.

ctreatma:
Closing; I think ultimately the ErrorResponse type and the `FriendlyError` methods should go away, since they don't seem to add much, but that's a bigger question and the existing code seems to be working well enough for now.

PR #704, opened 2024-06-21
fix: Adding "SUBMITTED_FOR_APPROVAL" attribute as Target string in connectionUpdate and Updating documentation

srushti-patl:
- Added "SUBMITTED_FOR_APPROVAL" as Target String in connectionUpdate method
- Updated fabric_connection_resource by adding IBM provider use case note


thogarty (dismissed):
LGTM

thogarty (approved):
LGTM

PR #706, opened 2024-06-21
feat: convert metal_gateway to equinix-sdk-go

ctreatma:
This converts the metal_gateway resource and data source from packngo to equinix-sdk-go to enable the later introduction of IPv6 support.

### DIFF
--- a/internal/resources/metal/gateway/resource.go
+++ b/internal/resources/metal/gateway/resource.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
 	"github.com/equinix/terraform-provider-equinix/internal/framework"
 	"github.com/hashicorp/terraform-plugin-framework-timeouts/resource/timeouts"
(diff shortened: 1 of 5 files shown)


internal/resources/metal/gateway/resource.go
```diff
@@ -144,34 +166,37 @@ func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp
 	id := state.ID.ValueString()
 
 	// API call to delete the Metal Gateway
-	deleteResp, err := client.MetalGateways.Delete(id)
+	_, deleteResp, err := client.MetalGatewaysApi.DeleteMetalGateway(ctx, id).Include(includes).Execute()
```
ocobles, line 169: Although it doesn't affect the result, is there any reason to specify Includes in the delete request?

ocobles (approved):
It looks good, I added a comment to understand if that was on purpose but AFAIK it shouldn't affect the result

internal/resources/metal/gateway/resource.go
ctreatma: Good catch; the includes are needed with the generated SDK because it validates responses based on the spec, and the response will fail validation without these included fields.  I added a comment to document that as well.

PR #711, opened 2024-06-25
fix: zSide.serviceToken.type. Unexpected value '' error message

d-bhola:
This checks that the serviceTokenType is only set when it is not an empty String value.


thogarty (approved):
LGTM. Ship it.

PR #713, opened 2024-06-27
fix: Fix provider crash when creating a metal gateway tied to an ipv6 reservation

RaptorGandalf:
This PR fixes a bug when creating metal gateways associated with IPv6 reservations. Previously the provider would assume the gateway was tied to an IPv4 reservation, which would result in an error when the `calculateSubnetSize` function was called. 

To fix this, an additional check was added to the `calculateSubnetSize` function which ensures the address is actually `IPv4` before executing the private ipv4 subnet logic.

Additionally an acceptance test was added which creates a metal gateway associated with an IPv6 reservation and VRF.


ctreatma (approved):
Nice fix!  New test looks good and passes.

PR #717, opened 2024-07-01
feat: Support IPv6 parameters when creating / updating interconnection virtual circuits

RaptorGandalf:
This PR adds support for the newly available IPv6 parameters when creating / updating interconnection virtual circuits. To allow this feature to also be used with shared interconnections, some changes to the create logic of the interconnection virtual circuit resource were made. 

With shared interconnections, the customer does not create virtual circuits, but rather they are created for them by the metal API. Attempting to create a virtual circuit on a shared connection will result in an error from the Metal API. 

To get around this, an optional `virtual_circuit_id` parameter has been added to the virtual circuit resource. If this value is present, the terraform create function will issue an update call to the metal API for the existing specified virtual circuit, rather than trying to create a new one. 

Additionally I have added support for the already existing IPv4 parameters in the update function which seemed to be missing.


internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -145,95 +167,104 @@ func Resource() *schema.Resource {
...
-
-	if _, ok := d.GetOk("vlan_id"); ok {
-		vncr.VlanVirtualCircuitCreateInput = &metalv1.VlanVirtualCircuitCreateInput{
-			ProjectId:   projectId,
-			Name:        &name,
-			Description: metalv1.PtrString(d.Get("description").(string)),
-			Speed:       metalv1.PtrString(d.Get("speed").(string)),
-			Vnid:        metalv1.PtrString(d.Get("vlan_id").(string)),
-		}
+	if _, ok := d.GetOk("virtual_circuit_id"); ok {
```
ctreatma, line 170: I think we'll also need some logic in `resourceMetalVirtualCircuitDelete` so that, if `virtual_circuit_id` is specified, it does not attempt to delete the virtual circuit (I think the API will error if you try to delete a shared connection's virtual circuit).

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -37,6 +37,11 @@ func Resource() *schema.Resource {
 				Description: "UUID of Connection where the VC is scoped to",
 				ForceNew:    true,
 			},
+			"virtual_circuit_id": {
```
ctreatma, line 40: This should also be `ForceNew: true`, because removing this attribute from config implies the user is changing from a shared connection VC to a dedicated connection VC.

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -111,12 +119,33 @@ func Resource() *schema.Resource {
 				Optional:     true,
 				RequiredWith: []string{"vrf_id"},
 				Description:  "The Metal IP address for the SVI (Switch Virtual Interface) of the VirtualCircuit. Will default to the first usable IP in the subnet.",
+				Computed:     true,
```
ctreatma, line 122: I think we could either add this and remove `RequiredWith` on line 120, which would make `metal_ip` optional for VRF VCs, or we could leave this attribute as-is (the `RequiredWith` means the user will always be forced to specify a value for `metal_ip` for a VRF VC).

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -111,12 +119,33 @@ func Resource() *schema.Resource {
...
 			},
 			"customer_ip": {
 				Type:         schema.TypeString,
 				Optional:     true,
 				RequiredWith: []string{"vrf_id"},
 				Description:  "The Customer IP address which the CSR switch will peer with. Will default to the other usable IP in the subnet.",
+				Computed:     true,
```
ctreatma, line 129: Same as above, we only need one of `RequiredWith` or `Computed`.  Adopting `Computed` may require changes elsewhere in the resource.

internal/resources/metal/virtual_circuit/resource.go
RaptorGandalf: Makes sense. I removed the required_with attribute since according to Metal API the `customer_ip` and `metal_ip` params are optional. Only `subnet` is actually required when working with a VRF VC.

internal/resources/metal/virtual_circuit/resource_test.go
```diff
@@ -130,3 +130,75 @@ func TestAccMetalVirtualCircuit_dedicated(t *testing.T) {
...
+            project_id = equinix_metal_project.test.id
+            port_id = equinix_metal_connection.test.ports[0].id
+            vrf_id = equinix_metal_vrf.test1.id
+			subnet = "10.0.0.0/31"
+			subnet_ipv6 = "2604:1380:4641:a00::4/126"
+        }`,
+		randstr, randstr, randstr, randstr, randstr)
+}
+
+func TestAccMetalConnection_sharedVrfWithVC(t *testing.T) {
```
ctreatma, line 176: Test looks good and I see it passing, but the name doesn't line up with others in the file (the name prefix should reference virtual circuit rather than connection, since the focus of the test is validating the VC resource).

```suggestion
func TestAccMetalVirtualCircuit_sharedVrf(t *testing.T) {
```

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -145,95 +172,104 @@ func Resource() *schema.Resource {
...
+				Vrf:         d.Get("vrf_id").(string),
+				// TODO: woof
+				Md5:          *metalv1.NewNullableString(metalv1.PtrString(d.Get("md5").(string))),
+				PeerAsn:      int64(d.Get("peer_asn").(int)),
+				Subnet:       d.Get("subnet").(string),
+				CustomerIp:   metalv1.PtrString(d.Get("customer_ip").(string)),
+				MetalIp:      metalv1.PtrString(d.Get("metal_ip").(string)),
+				SubnetIpv6:   metalv1.PtrString(d.Get("subnet_ipv6").(string)),
+				CustomerIpv6: metalv1.PtrString(d.Get("customer_ipv6").(string)),
+				MetalIpv6:    metalv1.PtrString(d.Get("metal_ipv6").(string)),
```
ctreatma, line 211: If the user doesn't specify values for these attributes, this code will send, e.g., `metal_ip: ""`.  Does the API accept an empty string for these parameters or will it return an error in that case?

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -422,6 +462,42 @@ func resourceMetalVirtualCircuitUpdate(ctx context.Context, d *schema.ResourceDa
...
+		}
+
+		if d.HasChange("subnet_ipv6") {
+			needsUpdate = true
+			subnet_ipv6 := metalv1.PtrString(d.Get("subnet_ipv6").(string))
+			ur.VrfVirtualCircuitUpdateInput.SubnetIpv6 = subnet_ipv6
+		}
+
+		if d.HasChange("customer_ipv6") {
+			needsUpdate = true
```
ctreatma, line 491: We should also add a step to the new test that covers this functionality (the new step would pass in a config that is identical to the config used in test step 1, but does not include ipv6 settings)

internal/resources/metal/virtual_circuit/resource.go
RaptorGandalf: It will accept an empty string and calculate those fields itself the same as if the parameters were not sent.

internal/resources/metal/virtual_circuit/resource.go
RaptorGandalf: Looks like I found a bug in the API when adding this step. Its tripping a validation error when subnet_ipv6 is sent as empty/nil rather than removing those details. Will work on getting a fix out for that.

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -37,6 +37,12 @@ func Resource() *schema.Resource {
 				Description: "UUID of Connection where the VC is scoped to",
 				ForceNew:    true,
 			},
+			"virtual_circuit_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "UUID of an existing VC to configure. Used in the case of shared interconnections where the VC has already been created.",
```
displague, line 43: Is `virtual_circuit_id` somewhat redundant, able to be inferred, from the required `connection_id`?

`virtual_circuit_id` can be confusing compared to the resource's `id`. I think users will naturally expect `virtual_circuit_id` to be set when a new VC is created and be confused when the value is empty.

Some alternative ways to approach/name this: (just thinking out loud)
* `behaviors { connection_prepared = true }`
* `use_shared_connection_virtual_circuit = true`

If the upstream extended the service to offer HA pairs in shared virtual circuits (similar to what Network Edge has introduced in the service over time), the `virtual_circuit_id` approach would be preferable. If we can't think of an alternative name, perhaps the description could add something like: "`virtual_circuit_id` is an optional parameter, not to be confused with the computed `id` attribute.".

RaptorGandalf: I believe there can be multiple virtual circuits associated with the shared connection, so I don't think we can infer which one to use. 

What about a name like `shared_connection_virtual_circuit_id` in addition to some details in the description help to help make it more clear?

ctreatma: There's precedent for the `virtual_circuit_id` approach in the `metal_port` resource, which accepts a `port_id` attribute.  That usage is a little different because a port can't be created directly, but I'm not sure we need much more than a longer description that more explicitly calls out that `virtual_circuit_id` is not the `id` attribute.

If we start getting into behavior flags and long attribute names to differentiate this, that brings me back to either having separate `metal_virtual_circuit` and `metal_shared_virtual_circuit` resources or moving the BGP settings to a `metal_virtual_circuit_bgp_config` resource.

internal/resources/metal/virtual_circuit/resource.go
ctreatma: I see the new test step and it looks to be passing in CI; resolving this thread.

internal/resources/metal/virtual_circuit/resource.go
ctreatma: Looking at this again, IMO the current description of the `virtual_circuit_id` attribute is sufficient for now, but the requirement to provide a `connection_id` in all cases could be potentially confusing.  I think we only need `connection_id` when creating dedicated connection VCs, in which case we should update the `connection_id` attribute definition as follows:

```
"connection_id": {
				Optional:    true,
                                 Computed: true,
                                 ExactlyOneOf: []string{"connection_id", "virtual_circuit_id"},
				Description: "UUID of Connection where the VC is scoped to.  Only used for dedicated connections",
				ForceNew:    true,
			}
```

ctreatma (approved):

PR #719, opened 2024-07-02
docs: use default template for metal_virtual_circuit

ctreatma:
This removes the custom templates for the `metal_virtual_circuit` resource and data source so that we can generate those docs from code and better ensure that docs stay up-to-date as we add, change, and remove functionality.

This required:
- Deleting the custom templates for the `metal_virtual_circuit` resource and data source
- Renaming examples directories and files that were created by `tfplugindocs migrate` to align with the [conventional paths](https://github.com/hashicorp/terraform-plugin-docs?tab=readme-ov-file#conventional-paths).
  - **NOTE:** in particular, example files go in directories called, e.g., `equinix_metal_virtual_circuit`, even though the `equinix_` prefix is omitted for all other files
- Moving the resource import script out of the template and into the examples directory


ctreatma:
FYI @equinix/governor-digin-fabric @equinix/governor-ne-network-edge-engineering  this PR doesn't require your review, but it is an example of how to switch from using a custom template for documentation to using the default templates.  The default templates have support for one TF example file (although that file could contain multiple example usages) and the default resource template can pull in an example import script as well.  The bullet list in the issue description lays out the steps for conversion; in particular there's a note in there about a mismatch between the file paths that were used by `tfplugindocs migrate` when I created the templates from existing docs and the conventional file paths for examples that are used by `make docs` (which is eventually equivalent to `tfplugindocs generate`).

ctreatma:
@RaptorGandalf when you have a moment, please review this PR.  This updates the virtual circuit resource and data source docs to cover the new attributes that were added for IPv6 support.  Until recently the provider docs were all/mostly hand-written; as the docs begin to leverage more generation features, it will be easier to update them alongside the code.  The files under `docs` are the important changes here; the rest the changes were needed to generate the docs instead of hand-writing them, and those changes won't appear in future PRs for virtual circuit docs.

RaptorGandalf (approved):

PR #720, opened 2024-07-03
test: Fabric Connections test sweeper

thogarty:
* Add test sweeper for fabric connections
* Modify internal/sweep package with constants, list var, and methods to support test sweeping for fabric
* Add config for fabric test client

### DIFF
--- a/internal/sweep/sweep_test.go
+++ b/internal/sweep/sweep_test.go
@@ -1,9 +1,9 @@
 package sweep_test
 
 import (
-	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/vlan"
 	"testing"
 
+	fabric_connection "github.com/equinix/terraform-provider-equinix/internal/resources/fabric/connection"
 	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/connection"
 	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/device"
 	"github.com/equinix/terraform-provider-equinix/internal/resources/metal/organization"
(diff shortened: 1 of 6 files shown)


ctreatma (commented):
The sweeper isn't getting called in CI because the GitHub Actions workflow is currently set up to only look for sweepers in the equinix directory.  You can update the Fabric acctest workflow to find sweepers in both the old and new locations by copying the command sweeper step of the Metal acctest workflow: https://github.com/equinix/terraform-provider-equinix/blob/main/.github/workflows/acctest.yml#L115-L124

thogarty:
> The sweeper isn't getting called in CI because the GitHub Actions workflow is currently set up to only look for sweepers in the equinix directory. You can update the Fabric acctest workflow to find sweepers in both the old and new locations by copying the command sweeper step of the Metal acctest workflow: https://github.com/equinix/terraform-provider-equinix/blob/main/.github/workflows/acctest.yml#L115-L124

@ctreatma , it's been updated. I also changed the filter from some type of sweeper suffix to all fabric sweepers because they have been made to search for any resource that matches fabric suffixes. We just run it with different users to ensure all our created resources are swept properly.

internal/sweep/sweep_test.go
```diff
@@ -25,6 +27,7 @@ func TestMain(m *testing.M) {
 func addTestSweepers() {
 	connection.AddTestSweeper()
 	device.AddTestSweeper()
+	fabric_connection.AddTestSweeper()
```
ctreatma, line 30: Not relevant for this PR or even for the Fabric team but the existing Metal sweepers in this file should all be renamed to `metal_<resource>`.

ctreatma:
I was wondering why CI was still not running the new sweeper for this PR and then I remembered that the acceptance test workflow is configured with the `pull_request_target` trigger, so GHA will not use the updated workflow file from the PR for security.

I ran the workflow manually and confirmed that the sweeper is now being run in CI: https://github.com/equinix/terraform-provider-equinix/actions/runs/9959400532

ctreatma (approved):
Workflow and sweeper wiring looks good to me, but someone on the Fabric team should validate that the sweeper job did what it was expected to do.

srushti-patl (approved):

jkallem-equinix (approved):
LGTM

PR #721, opened 2024-07-08
fix: convert metal_connection delete to equinix-sdk-go

ctreatma:
This issue came out of investigation into Metal acceptance test failures in which the tests appear to be failing because [a connection that was deleted came back from the API with status `deleted`](https://github.com/equinix/terraform-provider-equinix/actions/runs/9844086716/job/27176733992#step:6:2177)

> === NAME  TestAccMetalConnection_sharedVlan
>     resource_test.go:256: Error running post-test destroy, there may be dangling resources: exit status 1
>         
>         Error: Failed to delete Metal Connection 3e4601f8-5aa3-4696-bfc9-363dbb14554e
>         
>         Connection 3e4601f8-5aa3-4696-bfc9-363dbb14554e is in undexpected state
>         deleted
> --- FAIL: TestAccMetalConnection_sharedVlan (25.25s)

The metal_connection resource and data source were previously migrated from the deprecated packngo SDK to equinix-sdk-go, but in that migration the Delete function for the metal_connection resource was missed.

This updates the resource Delete function to use equinix-sdk-go instead of packngo.  This requires adding a state waiter to replicate the wait functionality that was built in to packngo.  As part of this migration, the waiter is updated to allow the `deleted` status that can be returned by the API for a deleted connection.

### DIFF
--- a/internal/resources/metal/connection/resource.go
+++ b/internal/resources/metal/connection/resource.go
@@ -3,7 +3,10 @@ package connection
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
+	"slices"
+	"time"
 
 	"github.com/equinix/equinix-sdk-go/services/metalv1"
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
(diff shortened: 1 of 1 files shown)


cprivitere (approved):
LGTM

PR #726, opened 2024-07-10
chore(ci): lint with golangci-lint

ctreatma:
This introduces linting in CI to help maintain code quality and identify issues that may not be caught by tests.  To avoid a big PR that resolves all existing linting issues, the CI workflow is configured to only look at files that are changed in a particular PR or commit; we'll need to keep an eye on this to make sure it's working as expected.  This will allow us to chip away at existing issues over time.  Eventually (ideally within a few months), the CI workflow should be updated to always lint the entire codebase.

This PR also introduces a `make lint` task for local execution, but this task lints the entire codebase and, for the time being, will catch some issues that are not flagged for a particular PR or commit in CI.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -25,6 +25,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          args: --whole-files # Remove alogn with only-new-issues by 2025
+          args: --whole-files # Remove along with only-new-issues by 2025
           version: ${{ steps.golangcilint.outputs.version }}
           only-new-issues: true # Remove along with pull-requests: read permission by 2025
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Determine golangci-lint version
         id: golangcilint
         run: |
-          line=$(grep GOLANGCI_LINT_VERSION= Makefile)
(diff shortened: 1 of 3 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Determine golangci-lint version
+        id: golangcilint
+        run: |
+          line=$(grep GOLANGCI_LINT_VERSION= GNUmakefile)
+          version=$(echo ${line} | cut -d = -f2)
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          args: --whole-files # Remove alogn with only-new-issues by 2025
```
displague, line 28:
```suggestion
          args: --whole-files # Remove along with only-new-issues by 2025
```

displague (commented):
> limit ci to checking files that are changed in a PR/commit

cool

GNUmakefile
```diff
@@ -11,6 +11,12 @@ ACCTEST_COUNT       ?= 1
 GOFMT_FILES         ?=$$(find . -name '*.go' |grep -v vendor)
 PKG_NAME            =equinix
 
+CRI=docker
```
displague, line 14: Rather than add a dependency on `docker` which devs may not have, perhaps we could just use `go run`:

```
 go run github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.1 run
```

displague (approved):

PR #727, opened 2024-07-10
feat: PA-VM Integration with Panorama Server

kpdhulipala:
feat: PA-VM Integration with Panorama Server
Allow additional parameters in vendor config (panoramaAuthKey, panoramaIpAddress)

### DIFF
--- a/equinix/data_source_network_device.go
+++ b/equinix/data_source_network_device.go
@@ -661,6 +661,17 @@ func createDataSourceVendorConfigurationSchema() map[string]*schema.Schema {
 			Sensitive:   true,
 			Description: neDeviceVendorConfigDescriptions["LicenseID"],
 		},
+		neDeviceVendorConfigSchemaNames["PanoramaIPAddress"]: {
+			Type:        schema.TypeString,
+			Computed:    true,
+			Description: neDeviceVendorConfigDescriptions["PanoramaIPAddress"],
+		},
+		neDeviceVendorConfigSchemaNames["PanoramaAuthKey"]: {
+			Type:        schema.TypeString,
+			Sensitive:   true,
+			Computed:    true,
+			Description: neDeviceVendorConfigDescriptions["PanoramaAuthKey"],
+		},
 	}
 }
 
--- a/examples/resources/network_device/example_9.tf
+++ b/examples/resources/network_device/example_9.tf
@@ -1,0 +1,45 @@
+# Create PA-VM firewall cluster with Panorama Server Integration
+# with Panorama Server IP and Panorama Auth Key in vendor Configuration
+
+data "equinix_network_account" "sv" {
+  metro_code = "SV"
+}
+
+resource "equinix_network_device" "panw-cluster" {
+  name            = "tf-panw"
+  metro_code      = data.equinix_network_account.sv.metro_code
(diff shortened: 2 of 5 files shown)


equinix/data_source_network_device.go
```diff
@@ -661,6 +661,19 @@ func createDataSourceVendorConfigurationSchema() map[string]*schema.Schema {
 			Sensitive:   true,
 			Description: neDeviceVendorConfigDescriptions["LicenseID"],
 		},
+		neDeviceVendorConfigSchemaNames["PanoramaIPAddress"]: {
+			Type:        schema.TypeString,
+			ForceNew:    true,
```
rling-equinix, line 666: We don't need forceNew tag for computed attributes.

equinix/data_source_network_device.go
```diff
@@ -661,6 +661,19 @@ func createDataSourceVendorConfigurationSchema() map[string]*schema.Schema {
...
+			Computed:    true,
+			Description: neDeviceVendorConfigDescriptions["PanoramaIPAddress"],
+		},
+		neDeviceVendorConfigSchemaNames["PanoramaAuthKey"]: {
+			Type:        schema.TypeString,
+			Sensitive:   true,
+			ForceNew:    true,
```
rling-equinix, line 673: Same here.

rling-equinix (changes requested):
Are all the doc changes generated automatically?

equinix/data_source_network_device.go
kpdhulipala: addressed

equinix/data_source_network_device.go
kpdhulipala: addressed

kpdhulipala:
> Are all the doc changes generated automatically?

added separate example file to cover PA-VM cluster with panorama specific fields in vendor config

examples/resources/network_device/example_9.tf
```diff
@@ -0,0 +1,45 @@
+# Create PA-VM firewall cluster with Panorama Server Integration
```
ctreatma, line 1: This won't automatically be picked up by the docs generator.  You will need to update `templates/resources/network_device.md` to add this example to the template.

To have the new attributes show up in the generated docs, you would have to either add them to the template manually or replace [these lines in the template](https://github.com/equinix/terraform-provider-equinix/blob/main/templates/resources/network_device.md.tmpl#L41-L176) with `{{ .SchemaMarkdown | trimspace }}`.  A similar change would need to be made in `templates/data-sources/network_device.md` in order to automatically generate attributes in the docs for that data source.

Once the templates are updated you will need to run `make docs` and commit the changes.

kpdhulipala: thanks @ctreatma , looks like now these are generated rightly and Go Tests / Validate docs (pull_request)  check is passed.

rling-equinix (approved):
Lgtm.

schadalawada (approved):

ctreatma (approved):
Rubber-stamping since this is all specific to Network Edge and it's already approved by that team.  Will need to follow up separately to adjust codeowners so that DRE isn't tagged for review of PRs like this one.

PR #728, opened 2024-07-11
chore(ci): rename acctest workflow since it only applies to Metal

ctreatma:


.github/workflows/metal_acctest.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
         TF_ACC_METAL_DEDICATED_CONNECTION_ID: ${{ secrets.TF_ACC_METAL_DEDICATED_CONNECTION_ID }}
       run: |
-        go test ./... -v -coverprofile coverage.txt -covermode=atomic -count 1 -parallel 8 -run "(Metal)|(Provider)|(Migration)" -timeout 180m
+        go test ./... -v -coverprofile coverage.txt -covermode=atomic -count 1 -parallel 8 -run "(Metal)" -timeout 180m
```
displague, line 113: I was curious if there were tests like `FCR...Metal`. There are not.

```
$ git grep 'func\ Test.*Metal' |  grep -E -v -e 'TestAcc((Pre)?Check)?(?i:(Data|Re)Source)?Metal'
internal/resources/metal/vlan/datasource_test.go:func TestMetalVlan_matchingVlan(t *testing.T) {
```

.github/workflows/test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       run: go build -v .
 
     - name: TF tests
-      run: go test -v --coverprofile coverage.txt -covermode=atomic -parallel 4 $(go list ./... |grep -v 'tests')
+      run: go test -v --coverprofile coverage.txt -covermode=atomic -parallel 4 ./...
```
displague, line 53: Perhaps the intent of the old code was to ignore coverage of the tests themselves (or testing_helpers). Should be fine.

displague (approved):

.github/workflows/test.yml
ctreatma: The `go list` subcommand was added to avoid running the terragrunt tests in the `tests` directory since there was no way to make them work in CI.  Those tests have since been deleted, so the subcommand was effectively a no-op after that.

Issue #724, opened 2024-07-10
[Feature]: Increase usage of generation for provider docs


ctreatma:
### Description

In v2 we adopted `terraform-plugin-docs` to enable generation of provider docs from provider code.  In the initial migration we created templates from the existing docs to minimize the amount of change introduced in a single PR.  This means that even though we use `terraform-plugin-docs`, we are still generally using it with hand-written, hard-coded templates that are simply copied from `templates/` to `docs/`.  This results in gaps between the attributes that are supported by a given resource or data source and the attributes that are mentioned in the docs.

To reduce this gap we should do one of the following for every custom template in `templates/data-sources` and `templates/resources`:
- Remove the template and use the default template instead
- Update the template to replace the schema documentation with a reference to `{{ .SchemaMarkdown | trimspace }}`

An example of the first approach can be seen in #719.

### New or Affected Terraform Resources

This affects documentation for all resources and data sources

### Potential Terraform Configuration

This is a documentation change and does not impact terraform configuration

PR #733, opened 2024-07-18
docs: updating documentation for NIMf connection example in Metal Connection Resource

srushti-patl:
- Updated following doc files for NIMF connection:
- resources/fabric_connection.md
- resources/metal_connection.md
- datasoruces/metal_connection.md


ctreatma (changes requested):
These changes need to be made in the `templates` directory; once the templates are updated, you can run `make docs` to regenerate the docs from the updated templates.

In the templates, you should consider replacing the hard-coded attribute listing with `{{ .SchemaMarkdown | trimspace }}` (see #724).

Issue #619, opened 2024-03-11
refactor: externalize NewFrameworkProvider as we do with SDKv2 Equinix Provider for easier downstream Pulumi provider


ocobles:
>  Comparing our Pulumi and Terraform provider to the Talos provider and their effort to work with an `internal/` provider, https://github.com/siderolabs/terraform-provider-talos/issues/130, I see that Talos had internalized the provider and needed to revert that to simplify the shimming.
> 
> Looking at how the Pulumi providers wrap the SDKv2 and Framework providers, it appears that this is the only resource from TF that they depend on -- the provider objects.
> 
> For example: 
> https://github.com/pulumiverse/pulumi-talos/blob/main/provider/resources.go#L35
> 
> Our SDKv2 Provider is exposed today as `equinix.Provider`. We intentionally left this object in an exported package because of the dependency that Pulumi and Crossplane (at the time) had on this object.
> 
> Looking at the Equinix TF provider, I think within `equinix/provider.go` we can export a framework provider function (name doesn't matter -- `equinix.NewFrameworkProvider(version, ua string)`?) that effectively calls on https://github.com/equinix/terraform-provider-equinix/blob/main/internal/provider/provider.go#L27C1-L32C1 
> 
> The Pulumi Equinix shim would take the return value of that function. With the `New...` function exported and taking user-agent as an argument, we wouldn't need the submodule, shim patch, or the UA patch.

_Originally posted by @displague in https://github.com/equinix/pulumi-equinix/issues/79#issuecomment-1980010449_

ctreatma:
Based on the difficulties we ran into with refactoring provider resources into isolated packages, we probably don't want to (and maybe can't) tie the framework provider back into the equinix package/directory.

I'm hesitant to introduce a public interface to satisfy Pulumi--and thus establish that there is some broader validity in consuming this provider as a Go module rather than a plugin binary--but if we _do_ go that route, we should introduce a new package for that purpose and hide both the SDK and framework providers behind it.

displague:
> _if we do go that route_, we should introduce a new package for that purpose and hide both the SDK and framework providers behind it.

Cyclical dependencies have been a concern with the refactor. That would make sense.

PR #735, opened 2024-07-19
feat: expose the FrameworkProvider type and functions via equinix/provider

displague:
In development of Terraform dependent providers, such as Crossplane and Pulumi, access to the Terraform Framework Provider is necessary in order to access the metadata and implementation of this provider directly.

The Equinix (SDK) Provider is currently exposed in a similar way, as equinix/provider.go (equinix.Provider).  This PR exposes equinix/provider/framework.go (moving it from internal/provider/provider.go) as `equinix.provider.FrameworkProvider`.

This branch can be used directly in the go.mod of Crossplane Provider PRs to test that this minor Terraform refactor has impact there.

I would have moved equinix/provider.go's `Provider` and related tests to equinix/provider/sdkv2.go and sdkv2_test.go, but that change would have been more substantial. This would be simpler to apply after #622 and perhaps after a similar internal/ PR for Metal where the remaining private types and functions in the equinix package are no longer needed.

Alternative approaches to take advantage of the implementation involve copying the Terraform provider code into those projects. This creates downstream build and sync work that can be easily avoided by exposing the FrameworkProvider.

Another alternative approach, taken by some downstream tools, is to fetch the Terraform provider binary, run the schema export command, and then process that JSON through conversion libraries specific to that downstream tool. This approach is also burdensome for downstream tools and maintainers.

Additional minor changes introduced in this PR:
- go1.22 fmt changes
- go1.22.0 go.mod fix for toolchain


ctreatma:
IMO keeping this private--and making the SDK provider private too--and requiring the downstreams to do more work is preferable to establishing an expectation that this provider can be used as a Go module and that we are going to provide a stable public interface.  I believe this aligns with how HashiCorp-managed providers (AWS, Google, Azure) are structured--and probably others--although I don't know which or how many of those providers are used as the basis for downstream integrations.

displague:
> provide a stable public interface

The public interface would be restricted to the Provider/FrameworkProvider types and their constructors. The implementation details remain private. The public interfaces of this provider/ package would be tied to the upstream FrameworkProvider changes, which could change. The SDK Provider interface stayed consistent for many years with new interfaces being exposed as new packages (SDKv2).

> to provide a stable public interface.

We are in v2+ currently and haven't adopted v2/ paths. We have many public functions in equinix/ that we intend to move to internal. We don't have a strong argument around this today, IMHO.

> aligns with how HashiCorp-managed providers (AWS, Google, Azure) are structured

There are other _potential_ motives for this as Terraform providers became desirable as surrogates for Go CRUD implementations, translations of APIs to declarative syntax. :tinfoil_hat:

displague:
moving equinix/provider.go's `Provider` to `equinix/provider/sdk.go` would be a breaking change from the Go SDK perspective. I withdrawal that suggestion.

ctreatma:
These 2 comments seem to be in direct conflict:

> moving equinix/provider.go's Provider to equinix/provider/sdk.go would be a breaking change from the Go SDK perspective. I withdrawal that suggestion.

> We are in v2+ currently and haven't adopted v2/ paths. We have many public functions in equinix/ that we intend to move to internal. We don't have a strong argument around this today, IMHO.

If the provider can be importable and not follow backwards compatibility guarantees, then it seems like it should be fine to move the SDK provider.

I maintain my concerns about introducing the potential overhead of ensuring that provider releases don't break arbitrary downstream tools; the existing downstream maintenance burden is a clear indicator of where support responsibilities lie.  That said, it sounds like we're taking it as a given that the provider should be importable, and if that's the case I'll reiterate my suggestion from https://github.com/equinix/terraform-provider-equinix/issues/619#issuecomment-1988674690:

> we should introduce a new package for that purpose and hide both the SDK and framework providers behind it

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/equinix/terraform-provider-equinix
 
-go 1.22
+go 1.22.0
```
ctreatma, line 3: We don't specify patch versions in other Go things (whether shipped as modules or binaries).  What makes it necessary here?

displague: When tools try to fetch the toolchain for this project they find `1.22`, which is not a release version -- there is no toolchain defined without full semver. 

```sh
$ go version
go: downloading go1.22 (linux/amd64)
go: download go1.22 for linux/amd64: toolchain not available
```

* https://github.com/golang/go/issues/62278#issuecomment-1693538776
* https://go.dev/doc/toolchain

An alternative approach is to specify `toolchain 1.22.0`.

ctreatma: Does that error happen for you in `metal-cli` and `equinix-sdk-go` as well, or only here?  And it's preventing you from running `go mod tidy` or you're trying to do something else?  That doesn't seem to be impacting me or Renovate, but I'm not clear exactly how to reproduce this problem.

PR #736, opened 2024-07-22
feat: remove credential validation on provider startup

ctreatma:
Previously we validated that the user specified at least one kind of Equinix credential during provider startup.  This behavior is unusual among Terraform providers and its utility is debatable, because it does not and cannot protect against these situations:

- Invalid credentials were provided
- Credentials were omitted for a service that is used (e.g., a Metal auth token was specified, but a client ID and secret were needed for the resources in the user's configuration)

In addition to the above shortcomings, validating credentials in the Config type that is shared between SDKv2 and framework providers forces unnecessary changes to the SDKv2 provider if a new service is onboarded to the framework provider.

This removes the credential validation from the config object so that any and all validation of the provider configuration is specified in the provider schema.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -102,6 +102,7 @@ func (c *Config) Load(ctx context.Context) error {
 	}
 
 	authClient.Timeout = c.requestTimeout()
+	//nolint:staticcheck // We should move to subsystem loggers, but that is a much bigger change
 	authClient.Transport = logging.NewTransport("Equinix", authClient.Transport)
 	c.authClient = authClient
 	neClient := ne.NewClient(ctx, c.BaseURL, authClient)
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -121,6 +102,7 @@ func (c *Config) Load(ctx context.Context) error {
 	}
 
 	authClient.Timeout = c.requestTimeout()
+	//nolint:staticcheck // We should move to subsystem loggers, but that is a much bigger change
```
ctreatma, line 105: Punted on this to avoid setting up subsystem loggers, which requires adding new subsystems to `ctx` and may also require defining new environment variables to configure log level per subsystem.

We could also consider just using [NewLoggingHTTPTransport](https://developer.hashicorp.com/terraform/plugin/sdkv2/logging/http-transport#newlogginghttptransport-transport-http-roundtripper) for now; that would avoid the need to add new subsystems to the context on provider startup but we would lose the log tagging we have now (e.g., at the moment packngo logs are logged with an `Equinix Metal (packngo)` prefix).

cprivitere (approved):

ocobles:
I agree that validation was not necessary here. However, Equinix provider adds a bit of complexity by using different types of authentication for different services. I think it might be nice to intercept 401 errors to display a message pointing users to the provider config documentation

PR #745, opened 2024-07-31
feat: add equinix_fabric_precision_time resource and data source

thogarty:
* Add precision_time resource/data source to internal package
* Use terraform-plugin-framework for development
* Add doc templates for the new resource/data_source


internal/resources/fabric/precision_time/models.go
```diff
@@ -0,0 +1,308 @@
...
+import (
+	"context"
+	"github.com/equinix/equinix-sdk-go/services/fabricv4"
+	fwtypes "github.com/equinix/terraform-provider-equinix/internal/framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+type ResourceModel struct {
```
ctreatma, line 12: IMO it's preferable to have a single, shared `Model` if the implementations are identical (these ones look the same to me but I might be missing a difference somewhere).

internal/resources/fabric/precision_time/resource_schema.go
```diff
@@ -0,0 +1,280 @@
...
+				},
+			},
+			"project": schema.SingleNestedBlock{
+				Description: "An object that contains the Equinix Fabric project_id used for linking the Time Precision Service to a specific Equinix Fabric Project",
+				CustomType:  fwtypes.NewObjectTypeOf[ProjectModel](ctx),
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.UseStateForUnknown(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"project_id": schema.StringAttribute{
```
ctreatma, line 225: Nit: it looks like there's only one of these project blocks and it has a single attribute; why not make `project_id` a top-level attribute of this resource?

examples/data-sources/fabric_precision_time/get_by_uuid.tf
```diff
@@ -0,0 +1,3 @@
+data "equinix_fabric_precision_time" "time_service" {
```
ctreatma, line 1: Note that you can avoid adding a custom doc template for this datasource by moving this file to `examples/data-sources/equinix_fabric_precision_time/data-source.tf`

examples/resources/fabric_precision_time/ntp_time_service.tf
```diff
@@ -0,0 +1,17 @@
+resource "equinix_fabric_precision_time" "ntp" {
```
ctreatma, line 1: There's more of a tradeoff for these examples than the data source example, but you could merge these resource examples into a single file, with `# comments` describing each block if necessary, put the merged file at `examples/resources/equinix_fabric_precision_time/resource.tf` and remove the custom doc template for the resource as well.

ctreatma:
It looks like this branch will need to be rebased on the latest from main to ensure that all necessary validation workflows run for this PR.

examples/resources/fabric_precision_time/ntp_time_service.tf
thogarty: I tried out this documentation adjustment you suggested for Data Source and Resource and didn't get the results that were expected. I still feel the templates are necessary based on what Fabric team expects before the schema information, and the examples weren't included following that pattern.

internal/resources/fabric/precision_time/resource_test.go
```diff
@@ -0,0 +1,268 @@
+package precision_time
```
ctreatma, line 1: We encourage black box testing rather than tying tests to private implementation details.  The `_test` package is a Go idiom that allows you to keep tests in the same directory as the code under test, but force those tests to use the code as a consumer would.  (Ultimately your team are the owners of this code and the decision as to whether to use black box testing for each resource is entirely up to you.)

```suggestion
package precision_time_test
```

templates/data-sources/fabric_precision_time.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+  Fabric V4 API compatible data resource that allow user to fetch Equinix Fabric Precision Time Service by Uuid
+---
+
+# equinix_fabric_precision_time (Data Source)
+
+Fabric V4 API compatible data resource that allow user to fetch Equinix Fabric Precision Time Service by Uuid
+
+Additional documentation:
+* Getting Started: https://docs.equinix.com/en-us/Content/Edge-Services/EPT/EPT.htm
+* API: https://developer.equinix.com/dev-docs/fabric/api-reference/fabric-v4-apis#precision-time
```
ctreatma, line 15: You could instead put this in the `Description` or `MarkdownDescription` attribute of the data source schema and reference `{{ .Description | trimspace }}` in the docs template.  This would keep the description in the docs while also making available to other tools.

templates/data-sources/fabric_precision_time.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+
+## Example Usage
+
+{{tffile "examples/data-sources/fabric_precision_time/get_by_uuid.tf"}}
+
+
+{{ .SchemaMarkdown }}
```
ctreatma, line 22: I've usually seen that as `{{ .SchemaMarkdown | trimspace }}`, which would I think eliminate some of the empty lines at the end of the generated docs.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	// Parse API response into the Terraform state
+	resp.Diagnostics.Append(plan.parse(ctx, eptChecked.(*fabricv4.PrecisionTimeServiceCreateResponse))...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Set state to fully populated data
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```
ctreatma, line 95: Is there a risk of overwriting API response data with corresponding plan data here?  I'm assuming that any intersections between request data fields and response data fields are expected to have the same value, but is that guaranteed to be the case?

ctreatma, line 95: Is this where you're running into issues due to fields that are only present in the request and not in the response?

internal/config/config.go
```diff
@@ -122,8 +124,9 @@ func (c *Config) Load(ctx context.Context) error {
 
 // NewFabricClientForSDK returns a terraform sdkv2 plugin compatible
 // equinix-sdk-go/fabricv4 client to be used to access Fabric's V4 APIs
+// Deprecated: migrate to NewFabricClientForFramework instead
 func (c *Config) NewFabricClientForSDK(d *schema.ResourceData) *fabricv4.APIClient {
-	client := c.newFabricClient()
+	client := c.newFabricClient(context.WithValue(context.Background(), ctxKey("fabricSDKv2"), "no_oauth_reload"))
```
ctreatma, line 129: I'd recommend waiting to make this change until you can update all of the functions calling `NewFabricClientForSDK` to pass in the correct context.  The background context is probably not the right one to latch on to.

internal/config/config.go
```diff
@@ -134,18 +137,37 @@ func (c *Config) NewFabricClientForSDK(d *schema.ResourceData) *fabricv4.APIClie
 // Shim for Fabric tests.
 // Deprecated: when the acceptance package starts to contain API clients for testing/cleanup this will move with them
 func (c *Config) NewFabricClientForTesting() *fabricv4.APIClient {
-	client := c.newFabricClient()
+	client := c.newFabricClient(context.WithValue(context.Background(), ctxKey("fabricSDKv2"), "no_oauth_reload"))
```
ctreatma, line 140: This is probably fine since it's only used in tests, but this could be an opportunity to define a client in the `acceptance` package for testing/cleanup instead.

internal/config/config.go
```diff
@@ -134,18 +137,37 @@ func (c *Config) NewFabricClientForSDK(d *schema.ResourceData) *fabricv4.APIClie
...
+	return client
+}
+
 // newFabricClient returns the base fabricv4 client that is then used for either the sdkv2 or framework
 // implementations of the Terraform Provider with exported Methods
-func (c *Config) newFabricClient() *fabricv4.APIClient {
+func (c *Config) newFabricClient(ctx context.Context) *fabricv4.APIClient {
 	//nolint:staticcheck // We should move to subsystem loggers, but that is a much bigger change
 	transport := logging.NewTransport("Equinix Fabric (fabricv4)", c.authClient.Transport)
+	if v := ctx.Value(ctxKey("fabricSDKv2")); v == nil {
```
ctreatma, line 161: IMO it'd be better to have two functions, one that takes a context and includes authClient creation, and the old one that didn't take a context.

I'm curious what is the benefit of choosing not to create a new authClient, though?

internal/resources/fabric/precision_time/resource.go
thogarty: Yea, this is the spot. It's related to advance_configuration only at this point. I'm considering separating ntp and ptp into their own attributes: 

`advance_configuration_ntp` and `advance_configuration_ptp`. 

NTP will be in request only and not in response. So it could be marked optional to allow for it being set early and not set later. 

PTP will be in request and response so it could stay as optional/computed. 

The nested object structure in terraform-plugin-framework along with it's strict checking of attribute behavior has proven difficult for me. Hoping this uncovers an appropriate path forward that doesn't actually involve modifying the schemas to be so different from the API schema similar to how `project_id` was handled. It can backfire if the `project` object gets extended or if `advance_configuration` gets extended.

thogarty: This also happens during Read and Update. Read when importing a resource, and Update when the response doesn't contain the details it expects. Though I haven't been able to test those as much because the Create has gotten a plan error every time `advance_configuration` is used. When `advance_configuration` is not used then we're able to proceed without plan errors based on parsing method putting some zero values there.

ctreatma: It looks like the data source and resource have the same attributes.  Is the data source able to load the `advance_configuration` details on read?  Or does that attribute not work at all for data sources?

Is there some other data in the API response that corresponds to the `advance_configuration` settings?  Or a different API endpoint that can be used fetch the advanced configuration details?

thogarty:
Resource is being reworked. Closing open PR.

PR #755, opened 2024-08-09
fix: [MAPI-24472]: cleanup logo from org

sbhatnagar-equinix:
logo attribute was removed as a request param from Metal APIs:
- POST /organizations
- PATCH /organizations/:id
This is to reflect the same in terraform


sbhatnagar-equinix:
> Thanks for opening this, @sbhatnagar-equinix.
> 
> We discussed elsewhere that this PR is in response to https://github.com/equinix/equinix-sdk-go/pull/71/files#diff-ff225eefc4189773d61a31b71d59b068352082b3eea9b9ab21388c78aea1365bL12, the removal of the `logo` field from Equinix Metal organizations. Additionally, the field has has not been configurable for over a year.
> 
> While the changes in this PR look accurate for removal of the field, I think we should apply the Terraform deprecation best practices and hold off on complete removal until the next major release.
> 
> https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-attribute-removal
> 
> What does that look like for this PR?
> 
> 1. Restore the Logo field in the resource and datasource schemas, adding the "Deprecated" boolean, and noting in the description that the default, empty string, is the only valid value and that support for this field has been removed from the API. (I believe `make docs` would add a deprecated note into the markdown, but because we have a `.tmpl` file we have to manually add `(Deprecated)` in the description)
>    
>    * https://github.com/equinix/terraform-provider-equinix/pull/755/files#diff-ae9524605e3007ee06d6316053eece6e05fb1ec9e5b99082995d450401d3f74cL42-L47
>    * https://github.com/equinix/terraform-provider-equinix/pull/755/files#diff-03570daed908ba323e249952233723f0d3df91144dd9f74e21a42fd4c981e9a0L48-L51
> 2. Setting the `parse` value to empty string, since the field can not be determined by the API.
>    
>    * https://github.com/equinix/terraform-provider-equinix/pull/755/files#diff-c4f9c071e0fec3ac370badc0f9a8439beef2dffe2394809bda88960ed53a6a2aL39
>    * https://github.com/equinix/terraform-provider-equinix/pull/755/files#diff-c4f9c071e0fec3ac370badc0f9a8439beef2dffe2394809bda88960ed53a6a2aL69
> 
> When the next semver major is to be released, the remnants of the logo field will then be removed. That PR could be opened now (ideally after the deprecation PR is merged to avoid rebase needs), with the majority of what is present in this PR.

Thanks Marques for pointing to the next steps. As discussed, this PR is only good for clean up of logo and since we need to first go through a cycle of deprecation I have created another branch  https://github.com/equinix/terraform-provider-equinix/tree/fix-deprecate-org-logo and will raise a pr soon based on instructions.
We can abandon this PR to be addressed again during the next release cycle. I would like to understand the timelines for that and when can I raise a PR for cleaning logo.

sbhatnagar-equinix:
Please review https://github.com/equinix/terraform-provider-equinix/pull/756 for deprecation of logo

PR #756, opened 2024-08-15
fix: [BMS-948]: deprecate logo for Metal Orgs

sbhatnagar-equinix:
This is to deprecate logo from metal orgs since logo was sunset from Metal API. 
Jira link: https://equinixjira.atlassian.net/browse/BMS-948


displague (approved):

displague:
I missed the linter error:
```
  Running [/home/runner/golangci-lint-1.56.2-linux-amd64/golangci-lint run --new-from-patch=/tmp/tmp-1803-XQQN5N3cmhOT/pull.patch --new=false --new-from-rev= --whole-files] in [/home/runner/work/terraform-provider-equinix/terraform-provider-equinix] ...
  internal/resources/metal/organization/datasource_schema.go:50: File is not `goimports`-ed (goimports)
  				Description: "**Deprecated** Logo URL",
  				Computed:    true,
  internal/resources/metal/organization/resource_schema.go:44: File is not `goimports`-ed (goimports)
  				Description: "**Deprecated** Logo URL",
  				Optional:    true,
  				Computed:    true,
  				Default:     stringdefault.StaticString(""),
  
  level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
  ```

We've enabled this check only on updated files. No need to worry about the formatting outside of the files that you have updated:

`goimports -w ./internal/resources/metal/organization/*schema.go`

File: internal/resources/metal/organization/models.go [hunk 1]
```diff
@@ -66,7 +66,7 @@ func (m *DataSourceModel) parse(ctx context.Context, org *packngo.Organization)
 	m.Description = types.StringValue(org.Description)
 	m.Website = types.StringValue(org.Website)
 	m.Twitter = types.StringValue(org.Twitter)
-	m.Logo = types.StringValue(org.Logo)
+	m.Logo = types.StringValue("")
```
ctreatma, line 69: Just to confirm: this field is no longer showing up in API responses, so the attribute is already effectively `""` for someone using the current version of the terraform provider?  If the logo attribute is still available in any API responses, then this would be a breaking change in the terraform provider.

ctreatma: Tested with curl and I see that the attribute is absent, so this should be a non-breaking change in the provider.

ctreatma (approved):
Test failures are due to machine availability and unrelated to this change.

PR #758, opened 2024-08-22
feat: use go-retryable's default retry policy

ctreatma:
Previously, Metal and Fabric clients [used a custom retry policy](https://github.com/equinix/terraform-provider-equinix/blob/4b6af14d184c8dff4c438f42520ac2347deb6551/internal/config/config.go#L259-L280) that had extremely limited retry scenarios.  This removes the custom retry policy for those clients so that they will use the [default retry policy for the `go-retryablehttp` client](https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L495-L542).

The custom retry policy in this provider was inherited from `terraform-provider-metal`.  It was [added to that repo 5 years ago](https://github.com/equinix/terraform-provider-metal/commit/c9b838763f6ed36ac7720760039baf203d2f6f9a) to avoid retrying on 5xx errors.

For comparison, the default retry policy of `go-retryablehttp` also aligns with [the retry policy for the AWS provider](https://github.com/hashicorp/terraform-provider-aws/blob/67c89ea17c8ee0ec8d86af82a9ffbe6ffc2318e3/docs/retries-and-waiters.md?plain=1#L58-L66).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2,14 +2,12 @@ package config
 
 import (
 	"context"
-	"crypto/x509"
 	"fmt"
 	"log"
 	"net/http"
 	"net/url"
 	"os"
 	"path"
-	"regexp"
 	"runtime/debug"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


cprivitere (approved):
lgtm

PR #762, opened 2024-08-27
feat: Adding Marketplace Subscription capability to Fabric Cloud Router Resource

srushti-patl:
- Added marketplace_subscription schema to fabric_cloud_router resource
- Marked 'account' schema as optional and computed because this parameter is optional while creating Fabric Cloud Router using Marketplace Subscription id.


equinix/resource_fabric_cloud_router.go
```diff
@@ -273,6 +313,10 @@ func resourceFabricCloudRouterCreate(ctx context.Context, d *schema.ResourceData
 	package_ := packageCloudRouterTerraformToGo(schemaPackage)
 	createCloudRouterRequest.SetPackage(package_)
 
+	schemaMarketplaceSubscription := d.Get("marketplace_subscription").(*schema.Set).List()
+	marketplaceSubscription := marketplaceSubscriptionCloudRouterTerraformToGo(schemaMarketplaceSubscription)
+	createCloudRouterRequest.SetMarketplaceSubscription(marketplaceSubscription)
```
thogarty, line 318: Should only set the subscription if the value is provided. Do a reflect deep equal check on empty return struct to avoid empty value in JSON request body.

equinix/resource_fabric_cloud_router.go
```diff
@@ -243,6 +272,17 @@ func projectCloudRouterTerraformToGo(projectTerraform []interface{}) fabricv4.Pr
 
 	return project
 }
+func marketplaceSubscriptionCloudRouterTerraformToGo(marketplaceSubscriptionTerraform []interface{}) fabricv4.MarketplaceSubscription {
+	if len(marketplaceSubscriptionTerraform) == 0 {
+		return fabricv4.MarketplaceSubscription{}
+	}
+	marketplaceSubscription := fabricv4.MarketplaceSubscription{}
+	marketplaceSubscriptionMap := marketplaceSubscriptionTerraform[0].(map[string]interface{})
+	subscriptionUUID := marketplaceSubscriptionMap["uuid"].(string)
```
thogarty, line 281: You added type as Optional/Computed but you aren't accounting for type as an input here. Additionally, do empty string checks before setting, and use setters instead of direct assignment on attributes.

equinix/resource_fabric_cloud_router.go
```diff
@@ -56,6 +57,23 @@ func fabricCloudRouterProjectSch() map[string]*schema.Schema {
 	}
 }
 
+func fabricMarketplaceSubscriptionSch() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"type": {
+			Type:        schema.TypeString,
+			Computed:    true,
```
thogarty, line 64: Just double checking, you can send type and uuid in the request AND they will be returned in the response if not provided, correct? We need to be spot on with attribute behaviors to prepare for terraform-plugin-framework migration when we move to tf-plugin-protocol v6

equinix/resource_fabric_cloud_router.go
```diff
@@ -132,9 +150,20 @@ func fabricCloudRouterResourceSchema() map[string]*schema.Schema {
...
+			Description: "Equinix Fabric Entity for Marketplace Subscription",
+			MaxItems:    1,
+			Elem: &schema.Resource{
+				Schema: fabricMarketplaceSubscriptionSch(),
+			},
+		},
 		"account": {
 			Type:        schema.TypeSet,
-			Required:    true,
+			Optional:    true,
```
thogarty, line 165: Same question about account. I believe this is correct but double checking.

thogarty (changes requested):
This PR also needs the document changes to the schema for the cloud router resource. We should also update the Cloud Router data source schemas for the same changes to allow users to grab that data as well.

We should merge Charles' docs PR first

equinix/resource_fabric_cloud_router.go
srushti-patl: account number is required in the account filed and I have marked it as required in schema definition.

templates/resources/fabric_cloud_router.md.tmpl
```diff
@@ -0,0 +1,29 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
```
thogarty, line 2: nit: Unnecessary change. You can remove lines 2, 3, 5, and 6. They aren't being added by the default implementation and aren't needed.

templates/resources/fabric_cloud_router.md.tmpl
```diff
@@ -0,0 +1,29 @@
...
+Fabric V4 API compatible resource allows creation and management of [Equinix Fabric Cloud Router](https://docs.equinix.com/en-us/Content/Interconnection/FCR/FCR-intro.htm#HowItWorks).
+
+Additional documentation:
+* Getting Started: https://docs.equinix.com/en-us/Content/Interconnection/FCR/FCR-intro.htm#HowItWorks
+* API: https://developer.equinix.com/dev-docs/fabric/api-reference/fabric-v4-apis#fabric-cloud-routers
+
+## Example Usage
+
+Fabric Cloud Router
+{{tffile "examples/resources/equinix_fabric_cloud_router/example_1.tf"}}
```
thogarty, line 24: So I don't know how the `resources/fabric_cloud_router.md` got updated correctly based on how you've oriented the examples in this template. The Code shows that example_1.tf is for the marketplace subscription and `example_2.tf` is for the standard cloud router. Could you run this again and see if it shows a diff? I think there's a disconnect here.

thogarty (changes requested):
Last changes requested. Other than that it looks good.

templates/resources/fabric_cloud_router.md.tmpl
srushti-patl: yes, I will rename example files.

thogarty (approved):
LGTM. Ship it

PR #764, opened 2024-08-28
chore(docs): use default template where possible for Fabric docs

ctreatma:
This removes custom templates for some Fabric data sources and resources to avoid missed docs changes due to hard-coded docs templates, which partially addresses #724.

Changes made here are:  in cases where there is a single example file,
  1. rename that example to align with [conventional paths for examples in `tfplugindocs`](https://github.com/hashicorp/terraform-plugin-docs?tab=readme-ov-file#conventional-paths)
  2. remove the custom template so that the default template is used
  3. update the resource/data source description in code to match what was in the previous docs


ctreatma:
@equinix/governor-digin-fabric for your reference (and/or review).  I noticed that #762 makes changes to the schema but doesn't include the corresponding docs updates because these templates are all currently hard-coded.

ocobles (dismissed):
lgtm! However there few [golangci-lint errors](https://github.com/equinix/terraform-provider-equinix/actions/runs/10604211199/job/29390307606?pr=764#step:5:38) that we may want to address in a different PR

ctreatma:
> lgtm! However there few [golangci-lint errors](https://github.com/equinix/terraform-provider-equinix/actions/runs/10604211199/job/29390307606?pr=764#step:5:38) that we may want to address in a different PR

Rather than bypass the required status check, since it's already limited to files that are changed in the PR, I fixed the issues.

Took a little more back-and-forth than I would have liked due to linter config.  I have a separate PR to adjust linter config here: #766

ocobles (approved):

docs/data-sources/fabric_cloud_routers.md
```diff
@@ -1,9 +1,5 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
-page_title: "equinix_fabric_cloud_routers Data Source - terraform-provider-equinix"
 subcategory: "Fabric"
-description: |-
```
displague, line 5: Can we add description into the resource definition to get matching markdown results? This is a small description, but we could miss some handy documentation links or greater details if they were specified on other resources (I'll keep an eye out in this review).

I presume the default page_title is as good or better.

docs/data-sources/fabric_connection.md
```diff
@@ -129,14 +123,13 @@ Read-Only:
...
 
 <a id="nestedobjatt--a_side--access_point"></a>
-
 ### Nested Schema for `a_side.access_point`
 
 Read-Only:
 
 - `account` (Set of Object) (see [below for nested schema](#nestedobjatt--a_side--access_point--account))
 - `authentication_key` (String)
-- `gateway` (Set of Object, Deprecated) **Deprecated** `gateway` Use `router` attribute instead (see [below for nested schema](#nestedobjatt--a_side--access_point--gateway))
```
displague, line 139: we lose a documented deprecation flag. Is the field not marked as deprecated? (also in z_side docs)

docs/resources/fabric_cloud_router.md
```diff
@@ -60,7 +55,7 @@ resource "equinix_fabric_cloud_router" "new_cloud_router"{
 
 - `description` (String) Customer-provided Fabric Cloud Router description
 - `href` (String) Fabric Cloud Router URI information
-- `order` (Block Set, Min: 1, Max: 1) Order information related to this Fabric Cloud Router (see [below for nested schema](#nestedblock--order))
```
displague, line 63: Min:1 will be lost. Not included in spec? Intended? (perhaps the documentation template isn't reflecting it)

docs/data-sources/fabric_cloud_routers.md
displague: The descriptions are in the body, they are not being rendered in the markdown frontmatter.  

https://github.com/equinix/terraform-provider-equinix/pull/764/files?diff=split&w=1#diff-329a375d2de207d4b7529b502ca8930f5636802bc6aa8934c848ed2972a7260aR32-R36

Is this intended? The first line of non-title markdown body text ends up getting the same treatment as a description frontmatter so it would be redundant to capture it as front-matter?

displague (commented):
Aside from some callouts, the move to a fully generated doc brought all-around improvement. Enums and fields that were missed are now documented and will stay documented through schema changes.

docs/data-sources/fabric_cloud_routers.md
ctreatma: The `description` frontmatter is not used by the registry: https://developer.hashicorp.com/terraform/registry/providers/docs#yaml-frontmatter

The `page_title` is used, but based on current page titles in the registry, the value we are specifying in data source and resource pages is equivalent to not specifying that attribute at all.

docs/data-sources/fabric_connection.md
thogarty: It is marked appropriately, but because it is a nested field the descriptions and further details aren't picked up by `tfplugindocs`. Just the name of the attribute and the type.

thogarty: Actually, there may be another reason for this. The data_source isn't showing descriptions for any of the nested attributes even though the resource schema displays them. Might be because of the way we're changing attribute behavior to computed and then setting the data_source value.

Resource attributes are showing nested descriptions.

docs/resources/fabric_cloud_router.md
thogarty: Order is optional so it shouldn't have a minimum of 1 anyway. This is intended.

thogarty (approved):
LGTM. Can be shipped.

PR #766, opened 2024-08-30
chore(ci): tell golangci-lint to report all issues

ctreatma:
By default, golangci-lint will only report a certain number of issues per linter and a certain number of the same issue per run.  That adds unnecessary turnaround time when fixing issues.

This PR updates our golangci-lint config file to tell the linter to report all errors.

The golangci-lint config structure has changed quite a bit since v1.56, so rather than fight their docs I'm also upgrading the linter to the latest.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,5 @@
 run:
   modules-download-mode: readonly
-  skip-dirs:
-    - cmd/migration-tool
 linters:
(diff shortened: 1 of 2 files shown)


ctreatma:
Ran into this config limitation while addressing linter issues on #764

ocobles (approved):

PR #767, opened 2024-08-30
chore(ci): skip Metal E2E tests on PRs that don't touch code

ctreatma:


ctreatma:
Made this change because I noticed the Metal E2E tests running on #766 even though that PR doesn't make any changes that would be relevant to the behavior of the Metal resources and data sources.

.github/workflows/metal_acctest.yml
```diff
@@ -14,12 +14,10 @@ name: Metal Acceptance Tests
...
     types: [opened, synchronize, reopened, ready_for_review]
-    paths-ignore:
-      - 'LICENSE'
-      - '**.md'
-      - 'website/**'
-      - 'docs/**'
-      - '.github/ISSUE_TEMPLATE/**'
+    paths:
+      - 'go.mod'
+      - 'go.som'
```
displague, line 19:
```suggestion
      - 'go.sum'
```

.github/workflows/metal_acctest.yml
```diff
@@ -14,12 +14,10 @@ name: Metal Acceptance Tests
...
+      - '**/*.go'
```
displague, line 20: Is there an option to combine this with a path-ignore of fabric/ne/...? If a PR only modified an existing NE resource, we could skip Metal tests.

ctreatma: I think it's safer to assume the code might be shared

displague (approved):

PR #773, opened 2024-09-10
feat: officially deprecate the metal_device_network_type resource

ctreatma:


docs/resources/metal_device_network_type.md
```diff
@@ -13,27 +13,26 @@ To learn more about Layer 2 networking in Equinix Metal, refer to
 
 If you are attaching VLAN to a device (i.e. using equinix_metal_port_vlan_attachment), link the device ID from this resource, in order to make the port attachment implicitly dependent on the state of the network type. If you link the device ID from the equinix_metal_device resource, Terraform will not wait for the network type change. See examples in [equinix_metal_port_vlan_attachment](port_vlan_attachment).
```
ctreatma, line 14: Was hoping that the addition of `DeprecationMessage` would feed into generated docs, but alas.  Should keep the hand-written template for now and delete it all when the resource is eventually removed.

docs/resources/metal_device_network_type.md
```diff
@@ -2,7 +2,10 @@
 subcategory: "Metal"
 ---
 
-# equinix_metal_device_network_type (Resource)
+# equinix_metal_device_network_type (Resource, Deprecated)
```
ctreatma, line 5: There doesn't seem to be a standard for marking entire resources as deprecated in the docs.  The [HashiCorp recommendations](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal) are focused only on the code.

ocobles: Actually the differentiation between resource / datasource in parentheses seems to be a thing of the past, I don't see it in the documentation of any other provider. For me it's fine like this, but the documentation generation plugin won't put things like that either if we stop using templates in the future.

ocobles: So the **Note** below might be enough, but again, I think the way you put it seems clarifying

PR #774, opened 2024-09-10
feat: Fabric route filter resource and data sources

thogarty:
* Addition of:
   * data equinix_fabric_route_filter
   * data equinix_fabric_route_filters
   * equinix_fabric_route_filter
 * Includes testing and documentation


internal/resources/fabric/route_filter/resource_test.go
```diff
@@ -0,0 +1,95 @@
+package route_filter_test
+
+import (
+	"context"
+	"fmt"
+	"github.com/equinix/equinix-sdk-go/services/fabricv4"
```
ctreatma, line 6: Not sure why it didn't show up as a comment, but the linter is failing because the imports in this file don't match what's expected by `goimports`

equinix/provider.go
```diff
@@ -90,6 +91,8 @@ func Provider() *schema.Provider {
 			"equinix_fabric_networks":                  fabric_network.DataSourceSearch(),
 			"equinix_fabric_port":                      dataSourceFabricPort(),
 			"equinix_fabric_ports":                     dataSourceFabricGetPortsByName(),
+			"equinix_fabric_route_filter":              fabric_route_filter.DataSource(),
+			"equinix_fabirc_route_filters":             fabric_route_filter.DataSourceSearch(),
```
ctreatma, line 95: There's a typo in the name of the search data source
```suggestion
			"equinix_fabric_route_filters":             fabric_route_filter.DataSourceSearch(),
```

internal/resources/fabric/route_filter/sweeper.go
```diff
@@ -0,0 +1 @@
+package route_filter
```
ctreatma, line 1: Does this resource need to be swept?  If not, this file can be omitted.

ctreatma (commented):
This may need to be rebased on the latest from `main`

thogarty:
@ctreatma , I made the changes for the comments. I won't be able to get the Validate Docs PR to pass because I do want to leave the docs out until this is publishable. Though I do need the rest of the code for the other resource additions coming after this one.

Do you have some guidance on how to proceed in scenarios like that?

ctreatma:
My recommendation is to document the resources as they are added to the provider.  Once a PR is merged to `main` it can be released at any time by any user with access to the release job, so we don't have a notion of a thing that is in `main` but is not exposed to users.  Given our on-demand release process, I don't think it's a big problem to roll out related resources across multiple releases, as long as the resources in each release work correctly.

I suppose you could disable the tests and remove the `provider.go` changes so that the resource and data source are unreachable, and then add the provider wiring and enable the tests later to actually expose the resources to customers, but then you won't get test feedback until much later and that enablement PR could become unmanageable if it contains a bunch of fixes for test failures that could have been caught earlier.

thogarty:
> My recommendation is to document the resources as they are added to the provider. Once a PR is merged to `main` it can be released at any time by any user with access to the release job, so we don't have a notion of a thing that is in `main` but is not exposed to users. Given our on-demand release process, I don't think it's a big problem to roll out related resources across multiple releases, as long as the resources in each release work correctly.
> 
> I suppose you could disable the tests and remove the `provider.go` changes so that the resource and data source are unreachable, and then add the provider wiring and enable the tests later to actually expose the resources to customers, but then you won't get test feedback until much later and that enablement PR could become unmanageable if it contains a bunch of fixes for test failures that could have been caught earlier.

Copy that. Will include docs and make this releasable. It won't be much use to customers until the other two are included, but it will work completely fine on its own.

thogarty:
@ctreatma , this is ready to go now.

ctreatma (approved):

srushti-patl (approved):

PR #780, opened 2024-09-19
docs: updated equinix_metal_connection examples

### LINKED ISSUE #777: [Docs]: some `equinix_fabric_connection` using deprecated `equinix_ecx_l2_connection` fields
I found at least three examples using resource `equinix_fabric_connection` that replaced `equinix_ecx_l2_connection` but without updating the fields, so now most of those fields are different

https://github.com/equinix/terraform-provider-equinix/blob/f303197fdec11f69584257a46900534adae2ce4c/examples/resources/equinix_metal_connection/shared_no_token_metal_to_fabric.tf#L28-L36

https://github.com/equinix/terraform-provider-equinix/blob/f303197fdec11f69584257a46900534adae2ce4c/examples/resources/equinix_metal_connection/shared_token_fabric_port_to_metal.tf#L24-L32

https://github.com/equinix/terraform-provider-equinix/blob/f303197fdec11f69584257a46900534adae2ce4c/examples/resources/equinix_metal_connection/shared_token_metal_to_csp.tf#L17-L31

ocobles:
- some examples with ecx resource were updated with the fabric resource but without using the correct fabric fields
- renaming examples to use the same terminology as that available in the documentation https://deploy.equinix.com/developers/docs/metal/interconnections/fabric-billed-fabric-vc/ https://deploy.equinix.com/developers/docs/metal/interconnections/metal-billed-fabric-vc/
- fixed errors in examples (tested)
- removed note "Service Token A-side / Z-side (service_token_type) is not generally available" since it's been GA for more than a year
- Removed example using legacy not automated workflow since it was deprecated
- fixes #777


templates/resources/metal_connection.md.tmpl
```diff
@@ -46,7 +44,7 @@ The following arguments are supported:
 * `metro` - (Optional) Metro where the connection will be created.
 * `facility` - (**Deprecated**) Facility where the connection will be created. Use metro instead; read the [facility to metro migration guide](https://registry.terraform.io/providers/equinix/equinix/latest/docs/guides/migration_guide_facilities_to_metros_devices)
 * `redundancy` - (Required) Connection redundancy - redundant or primary.
-* `type` - (Required) Connection type - dedicated or shared.
+* `type` - (Required) Connection type - dedicated, shared or shared_port_vlan.
```
ctreatma, line 47: While we're in here, it would be good to replace this whole `Argument Reference` section with `{{ .SchemaMarkdown | trimspace }}`.  That might require adding/updating property descriptions in the resource schema.

ctreatma (approved):

ctreatma:
Given the current codeowner mappings, this PR requires review from @equinix/governor-digin-fabric before GitHub will allow it to be merged.

If that review requirement is not acceptable/desirable, we should update the codeowner mappings in a separate PR so that docs file names for each Equinix service can reference other Equinix services without requiring review from the relevant service teams.

srushti-patl (approved):
LGTM

PR #781, opened 2024-09-20
fix: equinix_metal_organization.project_ids data

billf:


displague:
Thanks for opening this PR, @billf.

The changes make sense. What error did you receive that triggered this PR?

billf:
```
data "equinix_metal_project" "production" {
  name = "whatever"
}

// This panics.
data "equinix_metal_organization" "org" {
  organization_id = data.equinix_metal_project.production.organization_id
}
```

```
│   with data.equinix_metal_organization.org,
│   on main.tf line 19, in data "equinix_metal_organization" "org":
│   19: data "equinix_metal_organization" "org" {
│
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please
│ report the following to the provider developer:
│
│ Expected framework type from provider logic: basetypes.StringType / underlying type: tftypes.String
│ Received framework type from provider logic: types.ListType[!!! MISSING TYPE !!!] / underlying type: tftypes.List[tftypes.DynamicPseudoType]
│ Path: project_ids[0]
╵
```

ctreatma:
@billf would you be able to add a test covering this behavior so we can avoid reintroducing the bug in the future?

The test would look a lot like the existing organization data source test: https://github.com/equinix/terraform-provider-equinix/blob/e158a616ca5ed96ff0aed1ce18f9a396637c1973/internal/resources/metal/organization/datasource_test.go

billf:
> @billf would you be able to add a test covering this behavior so we can avoid reintroducing the bug in the future?

not a problem; it's actually why this PR has remained in draft mode. 

> The test would look a lot like the existing organization data source test: https://github.com/equinix/terraform-provider-equinix/blob/e158a616ca5ed96ff0aed1ce18f9a396637c1973/internal/resources/metal/organization/datasource_test.go, but modified to ensure that a project is created for the test organization before the data source runs:

i'll take a look, thanks for the pointer.

displague:
Closing this because Equinix Metal was removed from the provider in v5.0.0

PR #707, opened 2024-06-21
refactor: move device resource and datasource to separate package

ctreatma:
This moves the device resource and data sources out of the `equinix` package and into an isolated directory & package to align with #106.

PR #782, opened 2024-09-23
refactor:  migrate metal_vlan to equinix-sdk-go

ctreatma:


ctreatma:
I left the device linter errors in place here because they are already fixed in #707 which is explicitly focused on the device code.

displague:
#707 has been merged - rebase away!

internal/resources/metal/vlan/datasource_test.go
```diff
@@ -244,10 +244,10 @@ func TestMetalVlan_matchingVlan(t *testing.T) {
 		{
 			name: "ComplexMatch",
 			args: args{
-				vlans: []packngo.VirtualNetwork{{VXLAN: 987, FacilityCode: "fac", MetroCode: "skip"}, {VXLAN: 123, FacilityCode: "fac", MetroCode: "met"}, {VXLAN: 456, FacilityCode: "fac", MetroCode: "nope"}},
+				vlans: []metalv1.VirtualNetwork{{Vxlan: metalv1.PtrInt32(987), AdditionalProperties: map[string]interface{}{"facility_code": "fac"}, MetroCode: metalv1.PtrString("skip")}, {Vxlan: metalv1.PtrInt32(123), AdditionalProperties: map[string]interface{}{"facility_code": "fac"}, MetroCode: metalv1.PtrString("met")}, {Vxlan: metalv1.PtrInt32(456), AdditionalProperties: map[string]interface{}{"facility_code": "fac"}, MetroCode: metalv1.PtrString("nope")}},
```
displague, line 247: nit: breaking this up with newlines may help readers without line-wrapping views

internal/resources/metal/vlan/models.go
```diff
@@ -21,29 +21,34 @@ type DataSourceModel struct {
...
+func (m *DataSourceModel) parse(vlan *metalv1.VirtualNetwork) (d diag.Diagnostics) {
+	m.ID = types.StringValue(vlan.GetId())
+	m.VlanID = types.StringValue(vlan.GetId())
+	m.Vxlan = types.Int64Value(int64(vlan.GetVxlan()))
 	m.Facility = types.StringValue("")
 
-	if vlan.Description != "" {
-		m.Description = types.StringValue(vlan.Description)
+	if vlan.GetDescription() != "" {
+		m.Description = types.StringValue(vlan.GetDescription())
```
displague, line 31: Empty or not, when would we not want the model description to match the upstream description?

internal/resources/metal/vlan/models.go
```diff
@@ -68,28 +73,33 @@ type ResourceModel struct {
...
-	m.ID = types.StringValue(vlan.ID)
-	m.Vxlan = types.Int64Value(int64(vlan.VXLAN))
+func (m *ResourceModel) parse(vlan *metalv1.VirtualNetwork) (d diag.Diagnostics) {
+	m.ID = types.StringValue(vlan.GetId())
+	m.Vxlan = types.Int64Value(int64(vlan.GetVxlan()))
 	m.Facility = types.StringValue("")
 
-	if vlan.Description != "" {
-		m.Description = types.StringValue(vlan.Description)
+	if vlan.GetDescription() != "" {
```
displague, line 81: same comment as above, why wouldn't we set description unconditionally?

internal/resources/metal/vlan/resource.go
```diff
@@ -9,9 +9,10 @@ import (
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
 	"github.com/equinix/terraform-provider-equinix/internal/framework"
 
+	"github.com/equinix/equinix-sdk-go/services/metalv1"
+	packngo "github.com/equinix/equinix-sdk-go/services/metalv1"
```
displague, line 13: I don't see many references to "packngo" below that would justify keeping the alias. Also - metalv1 is imported twice because of the alias.

displague (approved):

internal/resources/metal/vlan/models.go
ctreatma: https://github.com/equinix/terraform-provider-equinix/pull/642

ctreatma: There may be other ways to address the bug fixed by that previous PR; I forget what else, if anything, I tried at that time, but I think I was aiming to maintain backwards compatibility.  The effect of this if block is that, if the description is null or `""` in the response, it is null in the model.  I don't recall if the API meaningfully differentiates between a null VLAN description and an empty VLAN description.

internal/resources/metal/vlan/models.go
```diff
@@ -21,29 +21,34 @@ type DataSourceModel struct {
...
+		m.Description = types.StringValue(vlan.GetDescription())
 	}
 
-	if vlan.Project.ID != "" {
-		m.ProjectID = types.StringValue(vlan.Project.ID)
+	if vlan.AdditionalProperties["project"] != nil {
```
ctreatma, line 34: This will require a change to the OpenAPI spec to declare the `project` attribute for VLANs

internal/resources/metal/vlan/models.go
```diff
@@ -21,29 +21,34 @@ type DataSourceModel struct {
...
+		project := vlan.AdditionalProperties["project"].(map[string]interface{})
+		project_id := project["id"].(string)
+		m.ProjectID = types.StringValue(project_id)
 	}
 
 	if vlan.Facility != nil {
-		m.Facility = types.StringValue(strings.ToLower(vlan.Facility.Code))
-		m.Metro = types.StringValue(strings.ToLower(vlan.Facility.Metro.Code))
+		facility_code := vlan.Facility.AdditionalProperties["code"].(string)
+		metro := vlan.Facility.AdditionalProperties["metro"].(map[string]interface{})
```
ctreatma, line 42: This _could_ be improved by updating the OpenAPI spec but probably not worth it in this case.

internal/resources/metal/vlan/resource.go
```diff
@@ -104,19 +104,18 @@ func (r *Resource) Create(ctx context.Context, request resource.CreateRequest, r
...
+	client := r.Meta.NewMetalClientForFramework(ctx, request.ProviderMeta)
 
 	var data ResourceModel
 	response.Diagnostics.Append(request.State.Get(ctx, &data)...)
 	if response.Diagnostics.HasError() {
 		return
 	}
 
-	vlan, _, err := client.ProjectVirtualNetworks.Get(
+	vlan, _, err := client.VLANsApi.GetVirtualNetwork(
```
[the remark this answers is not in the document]
ctreatma: Looks like the missing `href` is in the `facility` object, and there is no way to include `facility.href`, so the API spec will have to be corrected (it says `facility` is an `Href`, but that is impossible).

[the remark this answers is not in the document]
ctreatma: Since facility-based VLANs have long been deprecated, this feels unworthy of a fix in the API spec.  Since `facility.href` is defined as long as `facility` is not in the include list, I opted to instead load the facility details with a separate API request.

internal/resources/metal/vlan/models.go
ctreatma: On closer inspection the `Project` field in `packngo` came from the `assigned_to` attribute in the API response.  That attribute is defined in the API spec and is available via `equinix-sdk-go`.

Issue #233, opened 2022-07-27
equinix_metal_device 403s on failed provision are not detected


displague:
Metal devices may fail to provision due to hardware failures, provisioning timeouts, or other reasons. When this happens the metal instance may no longer be available through the API and the user may receive a 403 response.  Terraform should treat this as a failed provision and the device should be considered deleted.

This is generally done here: https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_device.go#L812-L831

But there may be some other scenarios, [such as deprovision](https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_device.go#L791-L794), where a device is moved into this 403 state. In these events, the "Read" function should be prepared to handle 403s and treat them as deletes, especially if the delete that resulted in a deprovision failure happened outside of Terraform's control loops.
 
https://github.com/equinix/terraform-provider-equinix/blob/master/equinix/resource_metal_device.go#L586-L591

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
* Please do not leave _+1_ or _me too_ comments, they generate extra noise for issue followers and do not help prioritize the request.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

### Affected Resource(s)

* equinix_metal_device

### Terraform Configuration Files

```tf
# Copy-paste your Terraform configurations here.
#
# For large Terraform configs, please use a service like Dropbox and share a link to the ZIP file.
#
# If reproducing the bug involves modifying the config file (e.g., apply a config,
# change a value, apply the config again, see the bug), then please include both:
# * the version of the config before the change, and
# * the version of the config after the change.
```

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

displague:
@ocobleseqx points out https://github.com/hashicorp/terraform-provider-aws/issues/16796 may be relevant here

PR #784, opened 2024-09-25
refactor: ignore HTTP response errors by status code

ctreatma:
This was broken out of #782 since it involves a widely-used helper function and will impact more than just the Metal VLAN resource.

Previously, the `IgnoreHttpResponseErrors` helper function accepted as arguments a number of other, existing helper functions that are meant to evaluate whether a particular API response has a certain HTTP status.  Here's [an example helper function for checking if the response is a 403](https://github.com/equinix/terraform-provider-equinix/blob/b1ed71b389705513308c1e47e7ac6ff0c46891cd/internal/errors/errors.go#L119-L130):

```go
func HttpForbidden(resp *http.Response, err error) bool {
	if resp != nil && (resp.StatusCode != http.StatusForbidden) {
		return false
	}

	switch err := err.(type) {
	case *ErrorResponse, *packngo.ErrorResponse:
		return IsForbidden(err)
	}

	return false
}
```

This function immediately returns false if the response status code _is not_ 403; otherwise it tries to convert the error argument to an `ErrorResponse` or a `packngo.ErrorResponse` and then use a different helper function to determine if that error represents a 403.  This assumes that the error object contains a copy of the response, which is only certain to be the case for code that uses `packngo`.  In practice we almost never did the necessary conversion to `ErrorResponse` for non-`packngo` resources and data sources, so in many cases we were probably not successfully ignoring the desired response codes.

This refactors  `IgnoreHttpResponseErrors` to inspect HTTP status codes directly in the response instead of relying on helper functions for specific custom error types.

This also removes the `FriendlyErrorForMetalGo` function which only existed in order to convert a separate HTTP response and error into an `ErrorResponse`, which is no longer necessary.  Some usage of `FriendlyError` is also removed in cases where that function had no or minimal effect.

### DIFF
--- a/internal/resources/fabric/cloud_router/sweeper.go
+++ b/internal/resources/fabric/cloud_router/sweeper.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
 
 	"github.com/equinix/equinix-sdk-go/services/fabricv4"
 	equinix_errors "github.com/equinix/terraform-provider-equinix/internal/errors"
@@ -64,7 +65,7 @@ func testSweepCloudRouters(region string) error {
 		if sweep.IsSweepableFabricTestResource(cloudRouter.GetName()) {
 			log.Printf("[DEBUG] Deleting Cloud Routers: %s", cloudRouter.GetName())
 			resp, err := fabric.CloudRoutersApi.DeleteCloudRouterByUuid(ctx, cloudRouter.GetUuid()).Execute()
-			if equinix_errors.IgnoreHttpResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(resp, err) != nil {
+			if equinix_errors.IgnoreHttpResponseErrors(http.StatusForbidden, http.StatusNotFound)(resp, err) != nil {
 				errs = append(errs, fmt.Errorf("error deleting fabric Cloud Router: %s", err))
 			}
 		}
--- a/internal/resources/metal/gateway/resource.go
+++ b/internal/resources/metal/gateway/resource.go
@@ -172,27 +172,26 @@ func (r *Resource) Delete(ctx context.Context, req resource.DeleteRequest, resp 
 	// only match one of MetalGateway or VrfMetalGateway if the included
 	// fields are present in the response
 	_, deleteResp, err := client.MetalGatewaysApi.DeleteMetalGateway(ctx, id).Include(includes).Execute()
-
-	if err != nil {
-		if deleteResp != nil {
-			err = equinix_errors.FriendlyErrorForMetalGo(err, deleteResp)
-		}
+	if equinix_errors.IgnoreHttpResponseErrors(http.StatusForbidden, http.StatusNotFound)(deleteResp, err) != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Failed to delete Metal Gateway %s", id), err.Error(),
+		)
 	}
+
 	if err == nil {
 		// Wait for the deletion to be completed
 		deleteTimeout := r.DeleteTimeout(ctx, state.Timeouts)
-		deleteWaiter := getGatewayStateWaiter(
+		deleteWaiter := getGatewayDeleteWaiter(
 			ctx,
 			client,
 			id,
 			deleteTimeout,
 			[]string{string(metalv1.METALGATEWAYSTATE_DELETING)},
-			[]string{},
 		)
 		_, err = deleteWaiter.WaitForStateContext(ctx)
 	}
 
-	if equinix_errors.IgnoreHttpResponseErrors(http.StatusForbidden, http.StatusNotFound)(nil, err) != nil {
+	if err != nil {
 		resp.Diagnostics.AddError(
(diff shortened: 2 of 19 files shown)


internal/resources/fabric/cloud_router/sweeper.go
```diff
@@ -64,7 +65,7 @@ func testSweepCloudRouters(region string) error {
 		if sweep.IsSweepableFabricTestResource(cloudRouter.GetName()) {
 			log.Printf("[DEBUG] Deleting Cloud Routers: %s", cloudRouter.GetName())
 			resp, err := fabric.CloudRoutersApi.DeleteCloudRouterByUuid(ctx, cloudRouter.GetUuid()).Execute()
-			if equinix_errors.IgnoreHttpResponseErrors(equinix_errors.HttpForbidden, equinix_errors.HttpNotFound)(resp, err) != nil {
```
displague, line 67: Are the function definitions of`HttpNotFound` and `HttpForbidden` being removed too?

displague: There's still a bit of packngo code remaining. I can see how it makes sense to wait before removing those functions entirely.

ctreatma: No, those functions are still used in calls to the `IgnoreResponseErrors` function which was initially written for packngo responses and is still used by packngo-based code.

displague:
This was an extra flaky CI AccTest run, mostly around network resources but also devices that timed out during provisioning and projects not being cleaned up as a result.

Triggering a retest.

---

(out of pr scope)
We could avoid these in the future with a test fuzzing (ignore attribute changes within a few seconds).

```
=== NAME  TestAccMetalDevice_importBasic
    resource_test.go:591: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        - 	"updated": "2024-09-25T21:58:29Z",
        + 	"updated": "2024-09-25T21:58:30Z",
          }
```

displague:
Looking at the failures again, they seem relevant to the change:
```
=== NAME  TestAccMetalGateway_privateIPv4
    resource_test.go:15: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: Failed to delete Metal Gateway 00448d03-8c3c-446c-86b2-d8082f0a7c3d
        
        API Error HTTP 404 404 Not Found Not found
--- FAIL: TestAccMetalGateway_privateIPv4 (22.67s)
```

404s should be considered successful deletes. It wasn't picked up. This happens for a few different resource types.

A device_timeout test fails with a 403, which may be exactly what we would have expected.

ctreatma:
> 404s should be considered successful deletes. It wasn't picked up. This happens for a few different resource types.

internal/resources/metal/virtual_circuit/resource.go
```diff
@@ -354,19 +355,24 @@ func resourceMetalVirtualCircuitRead(ctx context.Context, d *schema.ResourceData
...
-					// The resource delete function uses this waiter and relies
-					// on it to return an ErrorResponse error so it can treat
-					// a 404 as success.  This conversion is done here for now
-					// to avoid a larger refactoring.
-					err = equinix_errors.FriendlyErrorForMetalGo(err, resp)
+				// Per current usage, when waiting for delete we pass in an empty target
+				// list; therefore, if the target list is empty we assume that a 404 or 403
+				// from the API is a success rather than a failure.
+				if len(target) == 0 {
+					if equinix_errors.IgnoreHttpResponseErrors(http.StatusNotFound, http.StatusForbidden)(resp, err) == nil {
```
ctreatma, line 373: I suspect that we've been ignoring far more HTTP 403 errors than we should.  My understanding is that there are limited times when we might want to ignore a 403 for a Metal device, but that other resources do not exhibit the same behavior: https://github.com/equinix/terraform-provider-equinix/issues/233

ctreatma: I'm creating a separate issue for the above comment; the aim of this PR is to check for the same statuses, and there's already plenty of tech debt cleanup coming in through that narrow scope.

thogarty (approved):
All good from Fabric standpoint. Thanks @ctreatma !

internal/resources/metal/gateway/resource.go
```diff
@@ -214,17 +214,25 @@ func getGatewayAndParse(ctx context.Context, client *metalv1.APIClient, state *R
 	return diags, nil
 }
 
-func getGatewayStateWaiter(ctx context.Context, client *metalv1.APIClient, id string, timeout time.Duration, pending, target []string) *retry.StateChangeConf {
+func getGatewayDeleteWaiter(ctx context.Context, client *metalv1.APIClient, id string, timeout time.Duration, pending []string) *retry.StateChangeConf {
+	// deletedMarker is a terraform-provider-only value that is used by the waiter
+	// to indicate that the gateway appears to be deleted successfully based on
+	// status code
+	deletedMarker := "tf-marker-for-deleted-gateway"
```
ocobles, line 221: ✅

ocobles (approved):
lgtm!

PR #794, opened 2024-10-04
feat: Add BGP Metrics fields for fabric cloud router routing protocols

thogarty:
* Added additional optional fields to routing protocols bgp_ipv4 and bgp_ipv6 objects
* Included acceptance testing

Acceptance Tests for Fabric will be failing intermittently because in UAT the Vlan Tags are not being released quickly (or possibly at all). I've included evidence of local passing tests in a screenshot below:

image


srushti-patl (approved):
LGTM!
Need to fix Lint Validation and Go build

equinix/resource_fabric_routing_protocol.go
```diff
@@ -712,6 +748,16 @@ func routingProtocolBgpIpv4TerraformToGo(routingProtocolBgpIpv4Request []interfa
 	enabled := bgpIpv4Map["enabled"].(bool)
 	rpBgpIpv4.SetEnabled(enabled)
 
+	if outboundAsPrependCount := bgpIpv4Map["outbound_as_prepend_count"].(int); outboundAsPrependCount > 0 {
```
d-bhola, line 751: @thogarty Are we checking this condition `outboundAsPrependCount > 0` to validate that its in this range -  0, 1, 3, 7 ([wiki](https://equinixjira.atlassian.net/wiki/spaces/IC/pages/145495563542/Routing+Protocol+-+NextGen+API))

d-bhola: I realized that 0 won't be included then. Is there a different reason for `outboundAsPrependCount > 0` ?

thogarty: Ah, good find on the wiki. This creates a problem though. SDKv2 always assigns the zero value of the type and doesn't have the concept of `null` for an unassigned value. Zero value for int is always 0, and I wasn't trying to avoid sending the value in the API request if the user didn't explicitly include it in the config.

Might have to rework how we can handle this case in SDKv2; or we don't support this feature until we can move to terraform-plugin-framework.

thogarty: I have a plan that can work though. We can use strings for these values and convert internally during mapping.

d-bhola: That could work. 

In that case, if it's a string "0" (in SDKv2) it would mean an actual value of `outboundAsPrependCount` and if it is int 0, then it would mean 0 (for unassigned value/null).

d-bhola: Let me rethink this. 

You said we can use strings and convert during mapping, but then what about int 0 values (unassigned values)

d-bhola: Never mind, I understood how you can implement this in the code.

thogarty:
@d-bhola , updated it to string type to account for user possibly not adding the parameter and to account for allowing the user to choose zero explicitly.

New test run after changes:
Screenshot 2024-10-04 at 6 44 28 PM

srushti-patl (approved):
LGTM

PR #796, opened 2024-10-10
feat: Nfv 29975 - make ssh public key name optional in Create Virtual device request

kpdhulipala:
feat: Nfv 29975 - make ssh public key name optional in Create Virtual device request


rling-equinix (approved):
Lgtm.

PR #799, opened 2024-10-16
fix: Update Fabric Terraform Resources and Data sources for latest Fabric …

thogarty:
* Upgrade equinix-sdk-go to v0.46.0
* Update Port models to match changes in new Fabric SDK
* Update SimplifiedAccount to match changes in new Fabric SDK


srushti-patl (approved):

thogarty:
Closes #798

thogarty:
@ctreatma this should be released immediately to give customers a version that they can upgrade to once 4.17 API changes rollout. We do expect that early upgrades will break, but we would rather have that then not have a version available to customers that they can upgrade to for fixing the breaking issues.

The 4.17 API release is tomorrow evening (10/18)

displague (approved):
The release notes will need to include a note that all configurations with `equinix_fabric_port` and `equinix_fabric_ports` will need to updated to the latest version.

Migration notes will need to be included if the state will not be safely updated in-place.

thogarty:
> The release notes will need to include a note that all configurations with `equinix_fabric_port` and `equinix_fabric_ports` will need to updated to the latest version.
> 
> Migration notes will need to be included if the state will not be safely updated in-place.

Agree. The state should be able to be updated with a refresh with no other changes than an upgrade. Just modifying internal functionality and not the Terraform interface. Additionally, for ports we have no resources.

Merging this in.

PR #806, opened 2024-10-29
feat!: switch from tfprotov5 to tfprotov6

### LINKED ISSUE #738: [Feature]: adopt terraform protocol v6
### Description

Currently this provider uses terraform protocol v5.  This prevents us from using some nicer features of terraform-plugin-framework, especially around nested schemas.

The impact of moving from terraform protocol v5 to v6 should largely be internal to the provider, but it does mean dropping support for terraform <1 and should be done in a major release unless we can effectively demonstrate that this provider is not actively used with terraform <1.

### New or Affected Terraform Resources

_No response_

_No response_

ctreatma:
Closes #738 

This adopts [Terraform Plugin Protocol version 6](https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol#protocol-version-6) so that this provider can leverage newer features of the protocol, enabling provider developers to easily reference the latest plugin development docs and provide a better end-user experience.

Upgrading the Terraform Plugin Protocol means dropping support for Terraform < 1.0, which constitutes a breaking change.  No resources or datasources have been updated in this PR to take advantage of the features enabled by the latest protocol.  The new protocol features are only available to framework-based resources and datasources, and as of this PR, only Metal has those.

### DIFF
--- a/equinix/data_source_metal_device_bgp_neighbors_acc_test.go
+++ b/equinix/data_source_metal_device_bgp_neighbors_acc_test.go
@@ -14,7 +14,7 @@ func TestAccDataSourceMetalDeviceBgpNeighbors(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
 		ExternalProviders:        testExternalProviders,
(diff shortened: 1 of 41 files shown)


thogarty (approved):
LGTM. In favor of shipping it.

PR #813, opened 2024-11-04
chore: upgrade go to 1.23

ctreatma:
Renovate seems to be tripping on the new Go version as part of upgrading Go dependencies.


cprivitere (approved):
lgtm

PR #815, opened 2024-11-06
feat: Updating Service Token Resource for Aside Port Service Token

srushti-patl:
- Update Service Token Resource to support Aside Port Service Token Creation


internal/resources/fabric/service_token/resource_test.go
```diff
@@ -63,7 +63,54 @@ func TestAccFabricServiceToken_PNFV(t *testing.T) {
 	})
 }
 
-func testAccFabricServiceTokenConfig(serviceTokenName string, serviceTokenDescription string, virtualDeviceUuid string) string {
+func TestAccFabricAsidePortServiceToken_PNFV(t *testing.T) {
```
thogarty, line 66: This tests needs to be with PPDS suffix if we're going to use the ppds port on line 70. Is PPDS used for this test?

internal/resources/fabric/service_token/resource_test.go
```diff
@@ -63,7 +63,54 @@ func TestAccFabricServiceToken_PNFV(t *testing.T) {
...
+	ports := testing_helpers.GetFabricEnvPorts(t)
+	var portUuid string
+	if len(ports) > 0 {
+		portUuid = ports["ppds"]["dot1q"][0].GetUuid()
+	}
+	serviceTokenName, serviceTokenUpdatedName := "token_port_PNFV", "UP_Token_port_PNFV"
```
thogarty, line 72: Update suffixes to match actual user account used. Looks like a different user is being used here based onthe portUuid being pulled from the PPDS sub portion.

thogarty (changes requested):
Need to take a second look at suffixes for test if the PPDS user is the one being used to test.

internal/resources/fabric/service_token/resource_test.go
srushti-patl: We are going to use pnfv user, I will update this line:
portUuid = ports["ppds"]["dot1q"][0].GetUuid()

thogarty (approved):
Approving based on response to feedback. Ship it.

PR #819, opened 2024-11-15
feat: Provide examples for create device types -  ZSCALER APPC and PSE 

kpdhulipala:
feat: Provide examples for create device types -  ZSCALER APPC and PSE


displague (approved):
lgtm. Not sure we need the _account datasourc e to get the metro code, but it doesn't hurt.

shyamchalla (approved):

PR #821, opened 2024-11-22
feat: Add Terraform Provider Docs for Alibaba Integration

d-bhola:


templates/resources/fabric_connection.md.tmpl
```diff
@@ -94,6 +94,49 @@ EVPLAN Port to Network Connection:
...
+  resource "random_string" "random" {
+    length  = 12
+    special = false
+  }
+  ```
+  * A unique `connection_name` is generated by appending a 12-character random string to the variable name given by the user
+  * A Fabric Port to Alibaba Profile Connection is successfully created and displayed in the Fabric Portal
+  * An Express Connect Physical Connection resource is created and becomes visible in the Alibaba Portal
+
+#### 2. Formulate main.tf and output.tf Files
```
d-bhola, line 122: @thogarty - When we discussed, were you referring to changing the words in this step? (Currently, it is the same as in `README.md` of `terraform-equinix-fabric`)

templates/resources/fabric_connection.md.tmpl
```diff
@@ -94,6 +94,49 @@ EVPLAN Port to Network Connection:
 
 {{tffile "examples/resources/equinix_fabric_connection/fcr_to_metal.tf"}}
 
+Port to Alibaba Connection EVPL_VC Connection:
+
+{{tffile "examples/resources/equinix_fabric_connection/port_to_alibaba.tf"}}
+
+### Step by Step Instructions for Fabric Port to Fabric Alibaba Profile Connection
```
thogarty, line 101: To provide even more clarity you could rephrase this to `Step by Step Instructions for the Fabric Port to Fabric Alibaba Profile Connection Example Given Below`

thogarty (changes requested):
Small changes requested. Thanks for adding this!

templates/resources/fabric_connection.md.tmpl
```diff
@@ -94,6 +94,49 @@ EVPLAN Port to Network Connection:
 
 {{tffile "examples/resources/equinix_fabric_connection/fcr_to_metal.tf"}}
 
+Port to Alibaba Connection EVPL_VC Connection:
+
+{{tffile "examples/resources/equinix_fabric_connection/port_to_alibaba.tf"}}
```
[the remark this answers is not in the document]
d-bhola: Done ✔️

templates/resources/fabric_connection.md.tmpl
d-bhola: Done ✔️

examples/resources/equinix_fabric_connection/port_to_alibaba.tf
```diff
@@ -0,0 +1,39 @@
...
+      profile {
+        type = "L2_PROFILE"
+        uuid = "<service_profile_uuid>"
+      }
+      location {
+        metro_code = "SV"
+      }
+    }
+  }
+}
```
thogarty, line 39: I missed this; the instructions only work if you include the alicloud resources in this example. I had assumed you had included them but I was mistaken. 

Please select an option from the following and update the PR to match:
1. You update the instructions to only refer to what needs to be done with just a connection example to Alibaba without Alicloud provider integration (I prefer this to avoid needing to update our Terraform registry if Alicloud updates)
  * If this approach is taken (which I would suggest) you only need to include points 3 and 5 with rewording to make them appropriate for running the example file, accepting from Alibaba Portal, then deleting, and finishing the deletion from Alibaba portal.
3. You include the VBR and supporting information in the example so that the example makes sense

Choice is yours, but it does need to be modified to reflect what is actually in the example.

Let me know if this is confusing and we'll jump on a short call.

thogarty (changes requested):
I made an assumption about the PR and was incorrect. Please see comment.

examples/resources/equinix_fabric_connection/port_to_alibaba.tf
thogarty: After processing this, we most definitely want to go with option #1

thogarty: Would also be good to add a trailing newline in the example.tf as well. You can configure your editors to automatically do that for you. I would recommend that as well.

d-bhola: @thogarty  - 
Approach 1 - I think points 1, 3 and 5 would be needed. Point 1 can also mention about running the example.

thogarty: You can merge them into 1. Create and Accept connection

d-bhola: Okay.👍

docs/resources/fabric_connection.md
```diff
@@ -731,6 +731,79 @@ resource "equinix_fabric_connection" "fcr2metal" {
...
 
+Port to Alibaba Connection EVPL_VC Connection:
+
+### Step by Step Instructions for the Fabric Port to Fabric Alibaba Profile Connection Example Given Below
+
+#### 1. Create and Accept the Connection
+
+Create the connection using the Equinix Terraform provider. Then manually accept the connection request in the Alibaba Portal for the created physical connection.
+
+**Note:** The `connection_name` must follow either one of these patterns:
```
thogarty, line 742: I wasn't perfectly clear. We don't need any of the details related to connection name. That's only for when you're integrating with alicloud via the data source

thogarty (changes requested):
Change requested.

thogarty (approved):
LGTM. Ship it.

PR #826, opened 2024-12-03
feat: create device without default password(supported for only C8000V Autonomous and Fortinet firewall devices)

kpdhulipala:
feat: create device without default password(supported for only C8000V Autonomous and Fortinet firewall devices)


rling-equinix (approved):
Lgtm.

PR #833, opened 2024-12-18
feat: Updating Service Token Resource for network based token

srushti-patl:
- Updated Service Token Resource Schema and Model
- Updated docs for Service Token to add zside Network based token

### DIFF
--- a/internal/resources/fabric/service_token/models.go
+++ b/internal/resources/fabric/service_token/models.go
@@ -2,7 +2,6 @@ package service_token
 
 import (
 	"fmt"
-	"log"
 	"reflect"
 	"sort"
 	"time"
@@ -30,7 +29,9 @@ func buildCreateRequest(d *schema.ResourceData) fabricv4.ServiceToken {
 	serviceTokenRequest.SetExpirationDateTime(expirationTime)
 
 	descriptionConfig := d.Get("description").(string)
-	serviceTokenRequest.SetDescription(descriptionConfig)
+	if descriptionConfig != "" {
+		serviceTokenRequest.SetDescription(descriptionConfig)
+	}
 
 	connectionConfig := d.Get("service_token_connection").(*schema.Set).List()
 	connection := connectionTerraformToGo(connectionConfig)
@@ -121,8 +122,6 @@ func buildUpdateRequest(d *schema.ResourceData) ([][]fabricv4.ServiceTokenChange
 			if bandwidth, ok := oldBandwidthLimitMap["bandwidth_limit"]; ok {
 				if bandwidthLimitValue, ok := bandwidth.(int); ok {
 					oldAsideBandwidthLimit = bandwidthLimitValue
-				} else {
-					log.Printf("[DEBUG] Expected bandwidthLimit to be an integer, but got %T", bandwidth)
 				}
 			}
 		}
@@ -135,8 +134,6 @@ func buildUpdateRequest(d *schema.ResourceData) ([][]fabricv4.ServiceTokenChange
 			if bandwidth, ok := newBandwidthLimitMap["bandwidth_limit"]; ok {
 				if bandwidthLimitValue, ok := bandwidth.(int); ok {
 					newAsideBandwidthLimit = bandwidthLimitValue
-				} else {
-					log.Printf("[DEBUG] Expected bandwidthLimit to be an integer, but got %T", bandwidth)
 				}
 			}
 		}
@@ -180,12 +177,11 @@ func buildUpdateRequest(d *schema.ResourceData) ([][]fabricv4.ServiceTokenChange
 	}
 
 	if !areSlicesEqual(oldZsideBandwidth, newZsideBandwidth) {
-		patches = append(patches, []fabricv4.ServiceTokenChangeOperation{
-			{
-				Op:    "replace",
-				Path:  "/connection/supportedBandwidths",
-				Value: newZsideBandwidth,
-			}})
+		patches = append(patches, []fabricv4.ServiceTokenChangeOperation{{
+			Op:    "replace",
+			Path:  "/connection/supportedBandwidths",
+			Value: newZsideBandwidth,
+		}})
 	}
 
 	return patches, nil
@@ -414,9 +410,7 @@ func portTerraformToGo(portList []interface{}) fabricv4.SimplifiedMetadataEntity
 	priority := portListMap["priority"].(string)
 	locationList := portListMap["location"].(*schema.Set).List()
 
-	if uuid != "" {
-		port.SetUuid(uuid)
-	}
+	port.SetUuid(uuid)
 	if href != "" {
 		port.SetHref(href)
 	}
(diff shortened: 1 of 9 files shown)


ryli17 (commented):
With recent change (for Feb release) https://github.com/equinix-product/interconnection_fabric-api-model/pull/1000#pullrequestreview-2518132243, the `connection.type` is not mandatory anymore. Should we revise `models.go` accordingly?

internal/resources/fabric/service_token/models.go
```diff
@@ -128,22 +132,22 @@ func buildUpdateRequest(d *schema.ResourceData) []fabricv4.ServiceTokenChangeOpe
...
-			if bandwidth, ok := newBandwidthLimitMap["bandwidthLimit"]; ok {
-				newBandwidthLimit := bandwidth.([]interface{})
-				if len(newBandwidthLimit) > 0 {
-					newAsideBandwidthLimits := converters.IfArrToIntArr(newBandwidthLimit)
-					newAsideBandwidthLimit = newAsideBandwidthLimits[0]
+			if bandwidth, ok := newBandwidthLimitMap["bandwidth_limit"]; ok {
+				if bandwidthLimitValue, ok := bandwidth.(int); ok {
+					newAsideBandwidthLimit = bandwidthLimitValue
+				} else {
+					log.Printf("[DEBUG] Expected bandwidthLimit to be an integer, but got %T", bandwidth)
```
thogarty, line 139: I believe Terraform already does a check for this for you so that you don't have to. Could just be unnecessary noise in the code. Did you run into issues with it?

internal/resources/fabric/service_token/models.go
```diff
@@ -176,14 +180,15 @@ func buildUpdateRequest(d *schema.ResourceData) []fabricv4.ServiceTokenChangeOpe
 	}
 
 	if !areSlicesEqual(oldZsideBandwidth, newZsideBandwidth) {
-		patches = append(patches, fabricv4.ServiceTokenChangeOperation{
-			Op:    "replace",
-			Path:  "/connection/supportedBandwidths",
-			Value: newZsideBandwidth,
-		})
+		patches = append(patches, []fabricv4.ServiceTokenChangeOperation{
+			{
```
thogarty, line 184: Unnecessary indentation. You can move the opening bracket up a line and remove the indent of the lines below it.

internal/resources/fabric/service_token/models.go
thogarty: Also, you're not returning an error. You're just doing a print. We should remove this type of "code noise" that isn't adding value to the system itself.

internal/resources/fabric/service_token/resource.go
```diff
@@ -127,16 +143,27 @@ func waitForStability(uuid string, meta interface{}, d *schema.ResourceData, ctx
...
 		MinTimeout: 30 * time.Second,
 	}
 
-	_, err := stateConf.WaitForStateContext(ctx)
+	inter, err := stateConf.WaitForStateContext(ctx)
+	var serviceToken *fabricv4.ServiceToken
 
-	return err
+	if err != nil {
+		log.Printf("[ERROR] Error while waiting for state: %v", err)
```
thogarty, line 158: Should improve the error message. It doesn't say what state the method is waiting for.

internal/resources/fabric/service_token/resource.go
```diff
@@ -115,7 +131,7 @@ func resourceDelete(ctx context.Context, d *schema.ResourceData, meta interface{
 	return diags
 }
 
-func waitForStability(uuid string, meta interface{}, d *schema.ResourceData, ctx context.Context, tieout time.Duration) error {
+func waitForStability(uuid string, meta interface{}, d *schema.ResourceData, ctx context.Context, timeout time.Duration) (*fabricv4.ServiceToken, error) {
 	log.Printf("Waiting for service token to be created, uuid %s", uuid)
 	stateConf := &retry.StateChangeConf{
 		Target: []string{
```
thogarty, line 137: The line right below this says `string(fabricv4.SERVICETOKENSTATE_INACTIVE)` is the expected target. Shouldn't it be `string(fabricv4.SERVICETOKENSTATE_ACTIVE)`?

internal/resources/fabric/service_token/resource_test.go
```diff
@@ -186,10 +227,10 @@ func testAccFabricZsideVirtualDeviceServiceTokenConfig(serviceTokenName string,
   			}
 
 		}
-    `, serviceTokenName, serviceTokenDescription, virtualDeviceUuid)
+    `, serviceTokenName, serviceTokenDescription, portUuid)
 }
 
-func testAccFabricAsidePortServiceTokenConfig(serviceTokenName string, serviceTokenDescription string, portUuid string) string {
+func testAccFabricZsidePortServiceTokenConfig(serviceTokenName string, serviceTokenDescription string, portUuid string) string {
```
thogarty, line 233: It's a little strange how the diff shows up in this file. Is this because you were attempting to organize the flow of the file?

internal/resources/fabric/service_token/models.go
```diff
@@ -508,9 +519,31 @@ func networkTerraformToGo(networkList []interface{}) fabricv4.SimplifiedTokenNet
...
 	uuid := networkListMap["uuid"].(string)
+	href := networkListMap["href"].(string)
 	type_ := networkListMap["type"].(string)
-	network.SetUuid(uuid)
-	network.SetType(fabricv4.SimplifiedTokenNetworkType(type_))
+	name := networkListMap["name"].(string)
+	scope := networkListMap["scope"].(string)
+	locationList := networkListMap["location"].(*schema.Set).List()
+
+	if uuid != "" {
```
thogarty, line 528: These checks are good, but they should follow the required/optional flow of the schema. If a field is required, you don't have to check for empty strings because Terraform will guarantee it has a value by throwing an error message to the user if they attempt to do a plan or apply without it.

internal/resources/fabric/service_token/models.go
```diff
@@ -114,11 +118,11 @@ func buildUpdateRequest(d *schema.ResourceData) []fabricv4.ServiceTokenChangeOpe
...
 
-			if bandwidth, ok := oldBandwidthLimitMap["bandwidthLimit"]; ok {
-				oldBandwidthLimit := bandwidth.([]interface{})
-				if len(oldBandwidthLimit) > 0 {
-					oldAsideBandwidthLimits := converters.IfArrToIntArr(oldBandwidthLimit)
-					oldAsideBandwidthLimit = oldAsideBandwidthLimits[0]
+			if bandwidth, ok := oldBandwidthLimitMap["bandwidth_limit"]; ok {
+				if bandwidthLimitValue, ok := bandwidth.(int); ok {
+					oldAsideBandwidthLimit = bandwidthLimitValue
+				} else {
```
thogarty, line 124: Same here. The else should be removed.

thogarty: You can assume the type assertion is safe so you don't need the if either.

internal/resources/fabric/service_token/models.go
```diff
@@ -28,6 +29,9 @@ func buildCreateRequest(d *schema.ResourceData) fabricv4.ServiceToken {
 	}
 	serviceTokenRequest.SetExpirationDateTime(expirationTime)
 
+	descriptionConfig := d.Get("description").(string)
+	serviceTokenRequest.SetDescription(descriptionConfig)
```
thogarty, line 33: Only set this without doing a check if the description is required. If it's optional (which I believe it is) you should check for empty string before setting. This is a source of plan diff errors that you've been running into.

thogarty (changes requested):
Large PR; I've added the notes for what I was able to see in the first pass. Will review again once we've resolved open discussions.

Please respond to Ryan Li's comment on this PR as well. Could be useful to include that change here as well.

internal/resources/fabric/service_token/resource.go
srushti-patl: The service token will be in ACTIVE status once it has been used to create a connection and that connection is active. When the token is ready to be used, it will be in INACTIVE status.

internal/resources/fabric/service_token/resource_test.go
srushti-patl: yes, I have re-organized the examples in alphabetical orders.

ryli17 (dismissed):

srushti-patl:
> With recent change (for Feb release) [equinix-product/interconnection_fabric-api-model#1000 (review)](https://github.com/equinix-product/interconnection_fabric-api-model/pull/1000#pullrequestreview-2518132243), the `connection.type` is not mandatory anymore. Should we revise `models.go` accordingly?

Confirmed with Ryan that we will change the connection type parameter to optional after the February release. Those changes are not in production right now.

thogarty (changes requested):
One small change. The mapping method is critical because it's what's being set to state. It's a good hub to check whenever the plan is showing differences. Very closely related to attribute behaviors.

thogarty (approved):
LGTM. Ship it.

PR #741, opened 2024-07-26
chore: remove migration-tool

ctreatma:
This removes the `migration-tool`, since old `terraform-provider-packet` and `terraform-provider-metal` have been deprecated for a few years now.

PR #834, opened 2024-12-18
chore(ci): delay release publish until after artifacts are attached

ctreatma:
We have intermittently seen provider releases fail to show up in the Terraform registry.  When that happens we can usually redeliver the GitHub release webhook event and get the registry to find the new release.

While working on changes to the Equinix Python SDK, it occurred to me that the webhook issues we are seeing might be related to how we do automatic releases.  The semantic-release-action creates the GitHub release, and then the provider artifacts are built and uploaded by GoReleaser after the GitHub release has been created.  The webhook event is sent to the Terraform registry when the release is created, not after the artifacts are uploaded; it's possible that the registry attempts to load the release only to find that it does not include a provider, and that that is why our provider releases sometimes do not show up in the registry right away.

This PR updates the semantic-release configuration so that semantic-release-action creates a draft release, and the release is not published until after GoReleaser builds and uploads the necessary artifacts.  The release and publish steps are split into separate jobs in the release workflow so that we can more easily re-run specific parts of the workflow and easily skip the publish step if no new release was created.  These changes may resolve our issues with releases not showing up in the registry, but even if they don't, this is still a worthwhile quality-of-life improvement for the release workflow.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -32,45 +32,15 @@ builds:
       - goos: darwin
         goarch: "386"
     binary: "{{ .ProjectName }}_v{{ .Version }}"
-  - id: migration-tool
-    dir: ./cmd/migration-tool/
-    env:
-      # goreleaser does not work with CGO, it could also complicate
-      # usage by users in CI/CD systems like Terraform Cloud where
-      # they are unable to install libraries.
-      - CGO_ENABLED=0
-    mod_timestamp: "{{ .CommitTimestamp }}"
-    flags:
-      - -trimpath
-    ldflags:
-      - "-s -w -X main.Version={{.Version}}"
-    goos:
-      - freebsd
-      - windows
-      - linux
-      - darwin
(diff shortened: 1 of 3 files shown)


displague (approved):

.goreleaser.yml
```diff
@@ -32,45 +32,15 @@ builds:
...
   - format: zip
     name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
     builds:
       - provider
-      - migration-tool
     files:
       - LICENSE*
       - README*
       - CHANGELOG*
-      - src: "cmd/migration-tool/*.md"
```
displague, line 71: This would be a good time to chase with a PR to remove the tool and any other remnants (including doc mentions).

displague: I noticed you started that in https://github.com/equinix/terraform-provider-equinix/pull/741/files

PR #840, opened 2025-01-13
chore(ci): perform full linting on all PRs

ctreatma:


ctreatma:
Pursuant to semi-related discussion on #844, I think it's actually preferable to keep these limited PR lint settings permanently rather than linting the whole codebase for every PR that is opened; if/when we adopt additional or different linter configuration, it would be handy to enforce those changes gradually rather than needing a single massive PR that brings the whole repo in line with the updated style rules.

PR #842, opened 2025-01-17
feat: Metros Data Source Development

d-bhola:
Addition of two data sources in Terraform Provider: 
   1. **Get Metro By Code** `/fabric/v4/metros/{metroCode}`
       - Data Source Read Method - `datasources_metro_code.go`
       
   2. **GET all Metros** `/fabric/v4/metros`
       - Data Source Read Method - `datasources_metros.go`
       
**Data Source Schema Addition**
- `datasources_schema.go` (same for both data sources)

**Model Translation and Data Setting**
- `models.go` (same for both data sources) 

**Acceptance Tests**
- `datasources_test.go`

**Documentation**
- `fabric_metro.md`
- `fabric_metros.md`


d-bhola:
@thogarty and @ctreatma -  I did a rebase now. This PR is ready for review again.

internal/resources/fabric/metro/datasources_test.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+		},
+	})
+}
+
+func testAccFabricMetroDataSourcesConfig(code string, limit, offset int) string {
+	return fmt.Sprintf(`
+
+       data "equinix_fabric_metro" "metro" {
+		  metro_code = "%[1]s"
```
thogarty, line 70: info: the [1] is used when you want to repeat parameters in Sprintf. Otherwise you can omit because it will automatically pick them up in the order they are given after the formatted string.

No change necessary. Just an educational comment.

thogarty (changes requested):
Only `parseMetro` needs to be updated because there are multiple `if` blocks that aren't adding any value. 

The rest of the comments should just be taken as education and you should avoid changing them to keep your PR review response as straight forward as possible.

Thanks for adding in the metros, @d-bhola !

internal/resources/fabric/metro/models.go
```diff
@@ -0,0 +1,205 @@
...
+		return diags
+	}
+
+	m.MetroModel = metroModel
+
+	return diags
+}
+
+func (m *MetroModel) parse(ctx context.Context, metro *fabricv4.Metro) diag.Diagnostics {
+	diags := parseMetro(ctx, metro, &m.Type, &m.Href, &m.Code, &m.Region, &m.Name, &m.EquinixASN, &m.LocalVCBandwidthMax, &m.GeoCoordinates, &m.ConnectedMetros, &m.GeoScopes)
```
ctreatma, line 114: Nit: delegating to a `parseMetro` function doesn't seem to add much here, especially since it isn't reused.  The logic from `parseMetro` could be moved up to this `parse` function, and you could directly do things like `m.Href = types.StringValue(metro.GetHref())` instead of getting all of the individual struct field addresses and then dereferencing them in `parseMetro`.

d-bhola: Addressed this change 👍

internal/resources/fabric/metro/models.go
```diff
@@ -0,0 +1,188 @@
...
+	"context"
+
+	"github.com/equinix/equinix-sdk-go/services/fabricv4"
+	fwtypes "github.com/equinix/terraform-provider-equinix/internal/framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+type Model struct {
```
thogarty, line 13: I would keep this as MetroModel. There was no need to rename it. Linter errors are generated because of that change also.

thogarty (changes requested):

internal/resources/fabric/metro/models.go
```diff
@@ -0,0 +1,205 @@
...
+	var diags diag.Diagnostics
+	*href = types.StringValue(metro.GetHref())
+	*tp = types.StringValue(metro.GetType())
+	*code = types.StringValue(metro.GetCode())
+	*region = types.StringValue(metro.GetRegion())
+	if metro.GetName() != "" {
+		*name = types.StringValue(metro.GetName())
+	}
+
+	if equinixAsn != nil {
```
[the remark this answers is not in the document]
d-bhola: Addressed this change 👍

internal/resources/fabric/metro/models.go
d-bhola: @thogarty - I made this change to fix [this lint error](https://github.com/equinix/terraform-provider-equinix/actions/runs/13190729790/job/36823082982?pr=842) I was getting 

Pasted Graphic

d-bhola: `metro.MetroModel` creates a stutter, since metro comes twice in the same term. This led me to change the name. Did you face this issue in streams or any other models ?

thogarty: Ah, ok. I did not. I named it `BaseStreamModel`. Which does still apply here because you could use `BaseMetroModel`.

The concern I have is that we don't use this package outside of it for anything other than the Terraform provider. To see an abstract `Model` name requires some cognitive overhead to cross-reference it and validate that it is actually the MetroModel in the API Spec. Though if aware of the pattern and the stutter then less energy is expended doing that.

Technically it isn't wrong, so you can keep as `Model`. We'll just have this note for posterity if we want to revisit it.

thogarty (approved):
It's good. Appreciate your response to comment with output from linter on stuttering in import naming from the package.

internal/resources/fabric/metro/models.go
ctreatma: > The concern I have is that we don't use this package outside of it for anything other than the Terraform provider. 

I think this is true for all of our framework resource/datasource models, and if so, then I'm not sure why we export any of them.  I don't think the framework depends on them being available outside of the package in which they are defined.  If we start defining models as unexported structs (e.g., `metroModel`) that would be fine by the linter, since it would be impossible to reference that as `metro.metroModel`.

d-bhola: I will leave it as `Model`

PR #847, opened 2025-01-29
chore: use original API request context for auth token requests

ctreatma:
#844 updated the `Config` object to create a new auth client every time a new Fabric API client was created for SDK resources.  If we stuck with this pattern, we would have to duplicate that behavior for every Equinix API client except for Metal.  In addition, this pattern means that the provider will make a separate auth token request per resource rather than reusing valid tokens across all resources that use the same provider (and therefore auth) config.

This reverts the auth client change by adopting a custom HTTP transport and OAuth token source that enables passing the original request context for an API request to the auth token client.  As a result, we no longer need to use `equinix/oauth2-go`.


thogarty (approved):
Awesome, @ctreatma ! 

I'm still working through the lower level details but your implementation looks solid from a decent high level understanding.

I'm curious if this removes the API Request/Response from the log output though with a different round tripper. If it does then I'll add it back in with a follow up after looking into it.

thogarty:
> Awesome, @ctreatma !
> 
> I'm still working through the lower level details but your implementation looks solid from a decent high level understanding.
> 
> I'm curious if this removes the API Request/Response from the log output though with a different round tripper. If it does then I'll add it back in with a follow up after looking into it.

Answered my own question. It doesn't because we still wrap this in `logging.NewTransport` which handles that for us with an additional wrapper. Good to go 👍

ctreatma:
@displague also interested to get your take on this one.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/equinix/terraform-provider-equinix
 go 1.23
 
 require (
-	github.com/equinix/equinix-sdk-go v0.48.0
+	github.com/equinix/equinix-sdk-go v0.48.1-0.20250130155308-f4934ab22f00
 	github.com/equinix/ne-go v1.19.0
 	github.com/equinix/oauth2-go v1.0.0
```
displague, line 8: Can this be removed in this PR?

internal/authtoken/transport.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2014 The Go Authors. All rights reserved.
```
displague, line 1: Where is this sourced?

thogarty:
@ctreatma + @displague , what are your thoughts on this one? 

We've got an open Goalie ticket for auth related issues on fabric_service_profile data source. My theory is that they're hitting a rate limiting issue trying to get so many auth tokens for each call. 

This should fix that issue.

ctreatma:
I think we should be a bit more certain that the requests are being rate-limited for the customer in question before merging/releasing a change to auth logic for all customers.  While this should be safe for everyone, it's nonetheless a core change and comes with some risk.

In terms of pure mechanics, releasing this will require review/merge/release of the corresponding SDK changes, and I don't recall how hacked-in those changes are (I'll update this comment with a PR link once a PR exists).

go.mod
ctreatma: Done

internal/authtoken/transport.go
ctreatma: Added a code comment with a permalink to the original file & brief description of changes

ctreatma:
Taking out of draft because the test workflows ignore drafts now.

ctreatma:
@displague @thogarty The SDK has been updated with `accesstokenv1` support and the corresponding context-aware `TokenSource`, so this PR is ready for final review.

The Fabric test workflow was recently changed such that this PR did not trigger the workflow; rather than address that here I ran the test workflow manually against this branch:  https://github.com/equinix/terraform-provider-equinix/actions/runs/15980969812

PR #848, opened 2025-01-31
feat: Create Test Sweeper for Networks

d-bhola:
**Test Sweeper for Networks:** 
 -Added test sweeper in`internal/resources/fabric/network/sweeper.go`
 -Added `network.AddTestSweeper()` function call in `internal/sweep/sweep_test.go`
 
A screenshot of the sweeper test run: 
image

### DIFF
--- a/internal/resources/fabric/network/sweeper.go
+++ b/internal/resources/fabric/network/sweeper.go
@@ -21,7 +21,7 @@ func AddTestSweeper() {
 	})
 }
 
-func testSweeperNetworks(region string) error {
+func testSweeperNetworks(_ string) error {
 	var errs []error
 	log.Printf("[DEBUG] Sweeping Fabric Networks")
 	ctx := context.Background()
--- a/internal/resources/fabric/network/sweeper.go
+++ b/internal/resources/fabric/network/sweeper.go
@@ -16,7 +16,7 @@ import (
 
 func AddTestSweeper() {
 	resource.AddTestSweepers("equinix_fabric_network", &resource.Sweeper{
-		Name:         "equinix_fabric_connection",
+		Name:         "equinix_fabric_network",
 		Dependencies: []string{},
 		F:            testSweepNetworks,
 	})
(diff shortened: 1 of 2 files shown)


internal/resources/fabric/network/sweeper.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		log.Printf("Error loading meta: %v", err)
+		return err
+	}
+	fabric := meta.NewFabricClientForTesting()
+
+	name := fabricv4.NETWORKSEARCHFIELDNAME_NAME
+	likeOperator := fabricv4.NETWORKFILTEROPERATOR_LIKE
+	limit := int32(100)
+	offset := int32(10)
```
thogarty, line 42: Shouldn't be offsetting by 10 here. If we have networks within the first 10 this will skip them. Set offset to `0`.

thogarty (changes requested):
Just need to update offset. Otherwise it looks good.

nit (no need to update in this PR): in future PRs, make your output log comments specifically related to fabric. There are other network types in Equinix besides Fabric so having that differentiater will add more context if we ever need to come back to this based on log output.

internal/resources/fabric/network/sweeper.go
d-bhola: Addressed this change.

internal/resources/fabric/network/sweeper.go
```diff
@@ -0,0 +1,71 @@
...
+
+func AddTestSweeper() {
+	resource.AddTestSweepers("equinix_fabric_network", &resource.Sweeper{
+		Name:         "equinix_fabric_network",
+		Dependencies: []string{},
+		F:            testSweeperNetworks,
+	})
+}
+
+func testSweeperNetworks(_ string) error {
```
d-bhola, line 24: Removed `region` from the function parameters to avoid the lint error

thogarty (approved):
LGTM. Thanks @d-bhola !

PR #849, opened 2025-01-31
chore: Enable revive and stylecheck in golangci linters

thogarty:
* Enable `revive` and `stylecheck` linters

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,6 +10,8 @@ linters:
     - ineffassign
     - staticcheck
     - unused
+    - revive
+    - stylecheck
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -10,6 +10,8 @@ linters:
     - ineffassign
     - staticcheck
     - unused
+    - revive
+    - stylecheck
 issues:
   max-issues-per-linter: 0
-  max-same-issues: 0
\ No newline at end of file
+  max-same-issues: 0
```
ctreatma, line 17: The [docs for provider testing](https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/testcase) encourage the use of underscores in test function names, so we should exclude test files from that rule:

```suggestion
  max-same-issues: 0
  exclude-rules:
    # Allow underscores in names in test files because the
    # HashiCorp-recommended/documented convention uses
    # underscores in function names
    - path: _test\.go
      linters:
        - revive
      text: "^var-naming"

ctreatma (approved):

PR #851, opened 2025-02-04
chore: update local lint task to only lint changed files

ctreatma:
Since we might periodically change linter configuration, as done in #849, we should keep the existing PR validation behavior that only lints the files that were changed in a particular PR.  This will allow us to continue to avoid unnecessarily large PRs to fix linting failures if and when the linter configuration changes.

If we're keeping the current, limited PR validation behavior, we should use the same behavior for the `make lint` task to that developers can easily identify the minimum necessary changes locally.  This updates the `make lint` task to only lint files that have been changed since `origin/main`, which should closely approximate the behavior of the PR validation workflow.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -25,6 +25,6 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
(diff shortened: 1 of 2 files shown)


thogarty (approved):
LGTM. I tested locally as well. Thanks, @ctreatma !

PR #853, opened 2025-02-05
refactor: use embedded structs to simplify framework model code

ctreatma:


ctreatma:
@displague would like to get your thoughts on this when you have a moment to review

internal/resources/metal/gateway/models.go
```diff
@@ -8,54 +8,8 @@ import (
...
-		}
-
-		m.PrivateIPv4SubnetSize = calculateSubnetSize(gw.VrfMetalGateway.IpReservation)
-		m.State = types.StringValue(string(gw.VrfMetalGateway.GetState()))
-	}
-	return nil
-}
-
-type DataSourceModel struct {
+type BaseModel struct {
```
displague, line 11: Is there any reason to export BaseModel? Why not baseModel?

internal/resources/metal/gateway/models.go
```diff
@@ -64,9 +18,14 @@ type DataSourceModel struct {
 	State                 types.String `tfsdk:"state"`
 }
 
-func (m *DataSourceModel) parse(gw *metalv1.FindMetalGatewayById200Response) diag.Diagnostics {
+type ResourceModel struct {
+	BaseModel
+	Timeouts timeouts.Value `tfsdk:"timeouts"`
+}
+
+func (m *BaseModel) parse(gw *metalv1.FindMetalGatewayById200Response) diag.Diagnostics {
```
displague, line 26: Will we generally need a `parse` method in composite types to pickup the fields that are not defined in the embedded model?

 `func (m *DataSourceModel) DataSourceModel parse(...) { m.BaseModel.parse(...); m.GatewayID = ... }`

displague: I see Organization parse following that pattern, where the DataSourceModel relies on the ResourceModel's parse before adding additional fields. It seems this pattern may not have been consistently applied.

internal/resources/metal/gateway/models.go
ctreatma: Per https://github.com/equinix/terraform-provider-equinix/pull/842#discussion_r1947025206 there doesn't seem to be a reason to export any of the models

internal/resources/metal/gateway/models.go
ctreatma: The `parse` method is for copying API response values into a model, so it's only concerned with API-defined properties.  In the case of this `m.Timeouts`, that is a Terraform-defined property that has no meaning to the API; the `m.Timeouts` value is always set by the plugin framework based on the `tfsdk: "timeouts"` tag.

In cases where a composite type has additional API-defined properties, you are correct that there needs to be a `parse` method attached to the composite type that delegates to the underlying `parse` method(s) and also sets the additional properties.  There's an example of that in the `metal_organization` resource/datasource.

internal/resources/metal/gateway/models.go
ctreatma: Lower-cased all of the models affected by this PR.  The AWS provider does the same for its framework models.

internal/resources/metal/gateway/models.go
ctreatma: For reference: https://github.com/equinix/terraform-provider-equinix/pull/853/files#diff-c4f9c071e0fec3ac370badc0f9a8439beef2dffe2394809bda88960ed53a6a2aR54-R58

PR #854, opened 2025-02-06
feat: Create Test Sweeper for Service Profiles

d-bhola:
Test Sweeper for Service Profiles:
-Added test sweeper `internal/resources/fabric/service_profile/sweeper.go`
-Added `service_profile.AddTestSweeper()` function call in `internal/sweep/sweep_test.go`


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/equinix/terraform-provider-equinix
 go 1.23.0
 
 require (
-	github.com/equinix/equinix-sdk-go v0.49.1
+	github.com/equinix/equinix-sdk-go v0.49.2
```
ctreatma, line 6: Just to confirm, this SDK upgrade is required in order to add the service profile sweeper?

d-bhola: There was a defect in POST Service Profile Search API, that was fixed and updated in this new SDK version- v0.49.2. This SDK upgrade was needed to reflect the patch applied for the API changes.

PR #855, opened 2025-02-07
fix: Use planmodifiers on stream resource to avoid unnecessary update calls

thogarty:
* Computed fields trigger update call, but no API call needs to be made. Computed fields will get eventual consistency through the implicit `refresh` that occurs during `terraform plan` and `terraform apply` commands


srushti-patl (approved):

thogarty:
Found to be incorrect logic. Those values will change over time and we only use `UseStateForUnknown` on computed values that we know will not change after they've been set to state.

PR #881, opened 2025-03-31
feat: Adding Fabric EPT Resource and Data Source

srushti-patl:
- Added Resource `equinix_fabric_precision_time_service` 
- Added  Data Source `equinix_fabric_precision_time_service`
- Added Data Source `equinix_fabric_precision_time_services`


internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+	serviceID := state.ID.ValueString()
+	updateRequest, diags := buildUpdateRequest(ctx, state, plan)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	if len(updateRequest) > 1 {
+		resp.Diagnostics.AddError("Error updating Precision Time Service",
+			"This resource only accepts one attribute change at a time; please reduce changes and try again")
```
thogarty, line 151: This is an approach that works, but isn't feature parity aligned with other resources. We do a for loop of single update API calls to allow user to modify multiple fields at the same time and then check for their completions. You can view the connection resource for inspiration. That implementation could probably be improved as well but it's a good starting point to iterate from.

IMO, not a blocker for merge but more of an FYI and a possible follow up change. We can still move this through as is.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+
+	// Extract the ID of the resource from the state
+	id := state.ID.ValueString()
+
+	// API call to delete the Precision Time Service
+	_, deleteResp, err := client.PrecisionTimeApi.DeleteTimeServiceById(ctx, id).Execute()
+	if err != nil {
+		if deleteResp == nil || !slices.Contains([]int{http.StatusForbidden, http.StatusNotFound}, deleteResp.StatusCode) {
+			resp.Diagnostics.AddError(
+				fmt.Sprintf("Failed to delete Precision Time Service %s", id), err.Error())
```
thogarty, line 218: Need a return in here. Not exiting properly.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+	deleteWaiter := getDeleteWaiter(ctx, client, id, deleteTimeout)
+	_, err = deleteWaiter.WaitForStateContext(ctx)
+
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Failed to delete Precision Time Service %s", id), err.Error())
```
thogarty, line 232: Need returns after Diagnostics errors to get the code to exit properly.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+	}
+}
+
+func buildUpdateRequest(ctx context.Context, state resourceModel, plan resourceModel) ([]fabricv4.PrecisionTimeChangeOperation, diag.Diagnostics) {
+	var resp *resource.UpdateResponse
```
thogarty, line 237: This is redundant for error handling. You can create one mDiags and use that instead of creating an entire `resource.UpdateResponse` object that is only being used to keep the error stack.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+		}
+
+		updateRequest = append(updateRequest, fabricv4.PrecisionTimeChangeOperation{
+			Op:    fabricv4.PRECISIONTIMECHANGEOPERATIONOP_REPLACE,
+			Path:  fabricv4.PRECISIONTIMECHANGEOPERATIONPATH_PTP_ADVANCED_CONFIGURATION,
+			Value: ptpAdvancedConfigurationValue,
+		})
+	}
+
+	return updateRequest, diags
```
thogarty, line 397: Also shows that your appending to the resource.UpdateResponse isn't even getting to your output so those errors won't bubble up to the right place. Need to fix the error handling in this method.

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+}
+func buildCreateRequest(ctx context.Context, plan resourceModel) (fabricv4.PrecisionTimeServiceRequest, diag.Diagnostics) {
+	var diags diag.Diagnostics
```
thogarty, line 400: View mDiags approach in #883 for error handling in sub methods

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+		if diags.HasError() {
+			return fabricv4.PrecisionTimeServiceRequest{}, diags
+		}
+		request.SetProject(fabricv4.Project{ProjectId: project.ProjectID.ValueString()})
+	}
+
+	return request, diags
+}
+
+func getInt32Pointer(value basetypes.Int32Value) *int32 {
```
thogarty, line 502: We shouldn't need this Pointer helper methods. Using the correct Set methods on Fabric API Model renders them obsolete.

internal/resources/fabric/precision_time/resource.go
thogarty: Probably good to implement with the other changes requested in the PR

internal/resources/fabric/precision_time/models.go
```diff
@@ -0,0 +1,511 @@
...
+
+func (m *dataSourceByIDModel) parse(ctx context.Context, ept *fabricv4.PrecisionTimeServiceResponse) diag.Diagnostics {
+	m.EptServiceID = types.StringValue(ept.GetUuid())
+	m.ID = types.StringValue(ept.GetUuid())
+	diags := m.basePrecisionTimeModel.parse(ctx, ept)
+	return diags
+}
+
+func (m *dataSourceAllEptServicesModel) parse(ctx context.Context, eptResponse *fabricv4.ServiceSearchResponse) diag.Diagnostics {
+	var diags diag.Diagnostics
```
thogarty, line 161: follow mDiags approach in #883 for error handling diagnostics

internal/resources/fabric/precision_time/models.go
```diff
@@ -0,0 +1,511 @@
...
+}
+
+func (m *resourceModel) parse(ctx context.Context, routeAggregation *fabricv4.PrecisionTimeServiceResponse) diag.Diagnostics {
+	m.ID = types.StringValue(routeAggregation.GetUuid())
+	diags := m.basePrecisionTimeModel.parse(ctx, routeAggregation)
+	return diags
+}
+
+func (m *basePrecisionTimeModel) parse(ctx context.Context, ept *fabricv4.PrecisionTimeServiceResponse) diag.Diagnostics {
+	var diags diag.Diagnostics
```
thogarty, line 199: mDiags approach #883

internal/resources/fabric/precision_time/resource.go
```diff
@@ -0,0 +1,561 @@
...
+}
+
+func (r *Resource) Create(
+	ctx context.Context,
+	req resource.CreateRequest,
+	resp *resource.CreateResponse,
+) {
+
+	var plan resourceModel
+	diags := req.Plan.Get(ctx, &plan)
```
thogarty, line 46: We should use req.Config for Create and Update instead of Plan. We're more focused on what is in the HCL than what Terraform thinks the Plan should be.

thogarty (changes requested):
Changes requested

internal/resources/fabric/precision_time/resource.go
srushti-patl: We cannot update multiple fields at the same time for EPT Resource. When we update one attribute, EPT service state will go to provisioning status and if we try to update other attribute at the same time than it will fail. That's why I have added this condition which restrict user to do multiple update at the same time.

thogarty (approved):
LGTM

PR #884, opened 2025-04-12
fix: Updating GO SDK version and stream subscription resource

srushti-patl:
- Updating equinix-sdk-go version to 0.51.0
- Removing 'filters' schema attribute from equinix_fabric_stream_subscription resource and data source


thogarty (approved):
LGTM.

PR #887, opened 2025-04-28
fix: Fixing parsePtpAdvancedConfiguration function in EPT resource

srushti-patl:
- Updated parsePtpAdvancedConfiguration function so that it will only set the PtpAdavncedConfiguration value when it's passed from the terraform config.


internal/resources/fabric/precision_time/models.go
```diff
@@ -393,35 +396,47 @@ func parseNtpAdvanceConfiguration(ctx context.Context, ntp []fabricv4.Md5) (fwty
...
-func parsePtpAdvancedConfiguration(ctx context.Context, ptp fabricv4.PtpAdvanceConfiguration) (fwtypes.ObjectValueOf[ptpAdvanceConfigurationModel], diag.Diagnostics) {
+func parsePtpAdvancedConfiguration(ctx context.Context, ptp fabricv4.PtpAdvanceConfiguration, userPTPAdvancedConfig bool) (fwtypes.ObjectValueOf[ptpAdvanceConfigurationModel], diag.Diagnostics) {
+	if !userPTPAdvancedConfig {
+		return fwtypes.NewObjectValueOfNull[ptpAdvanceConfigurationModel](ctx), nil
+	}
 	result := ptpAdvanceConfigurationModel{}
 	hasValue := false
 	if ptp.GetTimeScale() != "" {
 		result.TimeScale = types.StringValue(string(ptp.GetTimeScale()))
+		hasValue = true
```
thogarty, line 407: You can get rid of this boolean being set and just use reflection to check the result vs a null total value to avoid potential bugs with missing this switch.

internal/resources/fabric/precision_time/models.go
```diff
@@ -393,35 +396,47 @@ func parseNtpAdvanceConfiguration(ctx context.Context, ntp []fabricv4.Md5) (fwty
 	return fwtypes.NewListNestedObjectValueOfValueSlice(ctx, emptySlice), diags
 }
 
-func parsePtpAdvancedConfiguration(ctx context.Context, ptp fabricv4.PtpAdvanceConfiguration) (fwtypes.ObjectValueOf[ptpAdvanceConfigurationModel], diag.Diagnostics) {
+func parsePtpAdvancedConfiguration(ctx context.Context, ptp fabricv4.PtpAdvanceConfiguration, userPTPAdvancedConfig bool) (fwtypes.ObjectValueOf[ptpAdvanceConfigurationModel], diag.Diagnostics) {
+	if !userPTPAdvancedConfig {
```
thogarty, line 400: Redundant checking. This needs to be resolved clearly in one place instead of creating tons of boundaries around it.

internal/resources/fabric/precision_time/models.go
```diff
@@ -228,8 +228,11 @@ func (m *basePrecisionTimeModel) parse(ctx context.Context, ept *fabricv4.Precis
 		mDiags.Append(diags...)
 		return mDiags
 	}
-
-	m.PtpAdvanceConfiguration, diags = parsePtpAdvancedConfiguration(ctx, ept.GetPtpAdvancedConfiguration())
+	var userPTPAdvancedConfig bool
+	if !m.PtpAdvanceConfiguration.IsNull() && !m.PtpAdvanceConfiguration.IsUnknown() {
```
thogarty, line 232: You should get rid of all the extra bools, and do the reflection check in this method. Then the deeper method can handle it's job clearly returning the result without a double check, and this can assign it only if it doesn't have the Null value received from it.

thogarty (changes requested):
Changes requested. There are simpler ways to handle the same issues.

thogarty (approved):
LGTM. Thanks, @srushti-patl !

PR #891, opened 2025-05-07
fix: Replace EOL ubuntu_20_04 operating system with current ubuntu_24_04 os

lucasmpb:
Ubuntu 20.04 will reach end of life on May 31 on Equinix Metal. This PR will replace any mention of `ubuntu_20_04` with the newer `ubuntu_24_04` version of the operating system.

### DIFF
--- a/equinix/common_test.go
+++ b/equinix/common_test.go
@@ -10,7 +10,7 @@ import (
 var (
 	preferable_plans  = []string{"x1.small.x86", "t1.small.x86", "c2.medium.x86", "c3.small.x86", "c3.medium.x86", "m3.small.x86"}
 	preferable_metros = []string{"ch", "ny", "sv", "ty", "am"}
-	preferable_os     = []string{"ubuntu_20_04"}
+	preferable_os     = []string{"ubuntu_24_04"}
 )
 
 // Deprecated: use the identical TestDeviceTerminationTime from internal/acceptance instead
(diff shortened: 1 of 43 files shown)


stevemar (approved):
this looks reasonable to me, but i will defer to @displague and @ctreatma

displague (approved):
lgtm - we may discover through tests that the OS is not offered in the same plan and location that 20.04 was offered.

The code is accurate and existing users should be able to use the examples and any error messages received to navigate those changes.

If we see CI failures as a result of this change, we should chase those the deployment settings to keep the CI signal strong.

equinix/common_test.go
```diff
@@ -10,7 +10,7 @@ import (
 var (
 	preferable_plans  = []string{"x1.small.x86", "t1.small.x86", "c2.medium.x86", "c3.small.x86", "c3.medium.x86", "m3.small.x86"}
```
stevemar, line 11: this list should probably be trimmed down to just c3.small

displague: The intent of the list is to ensure that testing can find a matching combination of plan + os + metro. Matches earlier in the list should be prioritized (I don't recall from memory if that is the case)

displague:
The linter requirements will require some more search/replaces of existing code for this PR to be mergeable.

lucasmpb:
@displague let me work on those suggestions.

equinix/common_test.go
stevemar: it can go in another PR.

ctreatma (approved):

ctreatma:
I see a few failed tests but they look like standard test flakes (failed provisions, lack of capacity).  Good to go!

PR #895, opened 2025-05-16
docs: removed "id" from metro and made "uuid" required for virtual device and network connection

rfong-eqx:
- removed "id" from metro example
- made "uuid" a required field for connection/access_point/virtual_device and connection/access_point/network

### DIFF
--- a/docs/data-sources/fabric_metros.md
+++ b/docs/data-sources/fabric_metros.md
@@ -75,6 +75,7 @@ output "first_metro_geo_scopes" {
 ### Read-Only
 
 - `data` (Attributes List) Returned list of metro objects (see [below for nested schema](#nestedatt--data))
+- `id` (String) The unique identifier of the resource
 
 <a id="nestedatt--pagination"></a>
 ### Nested Schema for `pagination`
--- a/docs/data-sources/fabric_metros.md
+++ b/docs/data-sources/fabric_metros.md
@@ -20,10 +20,6 @@ output "number_of_returned_metros" {
   value = length(data.equinix_fabric_metros.metros.data)
 }
 
-output "first_metro_id" {
-  value = data.equinix_fabric_metros.metros.data.0.id
-}
-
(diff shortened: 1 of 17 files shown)


docs/data-sources/fabric_metros.md
```diff
@@ -75,7 +75,6 @@ output "first_metro_geo_scopes" {
 ### Read-Only
 
 - `data` (Attributes List) Returned list of metro objects (see [below for nested schema](#nestedatt--data))
-- `id` (String) The unique identifier of the resource
```
thogarty, line 78: Why are we removing the ID here?

IIRC the ID is critical for Terraform Resource Lifecycle management. Though I understand that for the Data Source we have the important information in the data list returned.

Need to make sure this is tested so that this removal doesn't break any changes because there could be code that updates this attribute in the models.go file as well.

internal/resources/fabric/connection/resource_schema.go
```diff
@@ -465,7 +465,7 @@ func networkSch() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"uuid": {
 			Type:        schema.TypeString,
-			Optional:    true,
+			Required:    true,
```
thogarty, line 468: Want to make sure these changes are tested. What were the results?

thogarty (commented):
Hope these ones come through. Questions about intent and testing.

docs/data-sources/fabric_metros.md
rfong-eqx: That was a mistake. I was supposed to remove the "id" output from the example.

internal/resources/fabric/connection/resource_schema.go
rfong-eqx: @srushti-patl helped me get the tests going and they passed.

srushti-patl (approved):
LGTM

PR #899, opened 2025-05-22
feat: Generate html report from xml for Fabric Acceptance Tests

d-bhola:


.github/workflows/fabric_acctest.yml
```diff
@@ -116,7 +116,16 @@ jobs:
           EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PNFV }}
           METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
         run: |
-          go test ./... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m
+          go test ./internal/resources/fabric/... ./equinix/... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m | tee pnfv_test_output.log
```
thogarty, line 119: Any reason we had to change the path of the tests? Was there an issue with `./...`?

.github/workflows/fabric_acctest.yml
```diff
@@ -186,7 +195,16 @@ jobs:
           EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PFCR }}
           METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
         run: |
-          go test ./... -v -coverprofile coverage_pfcr.txt -covermode=atomic -count 1 -parallel 8 -run "(PFCR)" -timeout 180m
+          go test ./internal/resources/fabric/... ./equinix/... -v -coverprofile coverage_pfcr.txt -covermode=atomic -count 1 -parallel 8 -run "(PFCR)" -timeout 180m | tee pfcr_test_output.log
```
thogarty, line 198: Same comment here about path. We want to avoid changing anything that is unnecessary.

.github/workflows/fabric_acctest.yml
```diff
@@ -206,3 +224,47 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage_pfcr.txt
+
+  upload-test-report:
```
thogarty, line 228: Test report looks good

thogarty (commented):
Just curious about path update

.github/workflows/fabric_acctest.yml
d-bhola: I realized I did this format while limiting the number of test cases (by making tweaks in this path),  I was running while testing this. I should keep the existing one itself. 👍

d-bhola: Updated the path.

.github/workflows/fabric_acctest.yml
```diff
@@ -116,7 +116,16 @@ jobs:
           EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PNFV }}
           METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
         run: |
-          go test ./... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m
+          go test ./... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m | tee pnfv_test_output.log
```
ctreatma, line 119: This seems like a lot more change than I would have expected for generating an HTML report from the test results.  I would've thought something like [gotestsum](https://github.com/gotestyourself/gotestsum) would give you the JUnit report generation out of the box (as well as the ability to [automatically re-run failed tests](https://github.com/gotestyourself/gotestsum?tab=readme-ov-file#re-running-failed-tests) if you see test flakes like the Metal team does); then it would just be a matter of feeding the JUnit report produced by gotestsum into a junit-to-HTML converter.

Did you try using `gotestsum` and run into issues with it?  I think that would be an easier pattern to replicate across teams but if there are things it doesn't handle well, that would be good to capture somewhere so other teams don't run into those issues as well.

ctreatma: The report merging explains a fair amount of the new structure; thanks for pointing that out.

The `junit2html` tool that you are using has a (poorly documented, but existent) `--merge` option that can be used to merge multiple JUnit files into a single report.  That merged file can then be passed in to a second execution of `junit2html` for conversion to HTML.  There are already 2 new dependencies here (`junit2html` and `go-junit-report`), so using `gotestsum` and `junit2html` would have the same impact in terms of adding dependencies (I don't know that the number of dependencies is or needs to be much of a concern for these workflows, though).  If, in the future, you want/need to re-run failed tests, I think it's worth revisiting the use of `go-junit-report` at that time; switching wouldn't make all that much difference here.

I don't think it's necessary to document the rationale behind these changes, but if you do, please make sure to mention the alternative tool selections.

.github/workflows/fabric_acctest.yml
```diff
@@ -116,7 +116,16 @@ jobs:
...
+
+      - name: Upload PNFV Testing Log
+        uses: actions/upload-artifact@v4
+        with:
+          name: pnfv_test_logs
+          path: pnfv_test_output.log
+
+      - name: Check if Tests are passed
+        run: sh scripts/check_tests.sh pnfv_test_output.log
```
ctreatma, line 128: Test runs are failing in part because this script file doesn't exist in the repo.  Can you add it?

PR #900, opened 2025-05-22
[Feature]: Allow update of VLAN description

### LINKED ISSUE #898: [Feature]: Allow update of VLAN description
### Description

Right now if you want to update the optional field of 'description' for 'vlan', terraform destroys and recreates the entity. As this is an updatable field in the GUI, can you support update instead.

Currently its set to `RequiresReplace()` on https://github.com/equinix/terraform-provider-equinix/blob/bad0ac7f60eede668216ac0eacab1d1064e509a5/internal/resources/metal/vlan/resource_schema.go#L38

```
"description": schema.StringAttribute{
	Description: "Description string",
	Optional:    true,
	PlanModifiers: []planmodifier.String{
		stringplanmodifier.RequiresReplace(),
	},
},
```

### New or Affected Terraform Resources

equinix_metal_vlan

_No response_

lucasmpb:
[closes #898]

This PR will update the metal vlan resource, to support in place update of the Description attribute. As [packngo](https://github.com/equinixmetal-archive/packngo/blob/master/virtualnetworks.go) doesn't support the update method, I had to use the [equinix-sdk-go](https://github.com/equinix/equinix-sdk-go/blob/main/services/metalv1/model_virtual_network.go) implementation of the virtual network resource.

### DIFF
--- a/internal/resources/metal/vlan/models.go
+++ b/internal/resources/metal/vlan/models.go
@@ -11,6 +11,9 @@ import (
 	"github.com/packethost/packngo"
 )
 
+// DataSourceModel represents the schema for reading VLAN data from the Equinix Metal API
+// in a Terraform data source context. It includes fields that describe the VLAN's identity,
+// location, and associated devices, and is used to populate Terraform state during read operations.
 type DataSourceModel struct {
 	ID                 types.String `tfsdk:"id"`
 	ProjectID          types.String `tfsdk:"project_id"`
@@ -19,7 +22,7 @@ type DataSourceModel struct {
 	Facility           types.String `tfsdk:"facility"`
 	Metro              types.String `tfsdk:"metro"`
 	Description        types.String `tfsdk:"description"`
-	AssignedDevicesIds types.List   `tfsdk:"assigned_devices_ids"`
+	AssignedDevicesIDs types.List   `tfsdk:"assigned_devices_ids"`
 }
 
 func (m *DataSourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostics) {
@@ -52,14 +55,17 @@ func (m *DataSourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostic
 		}
 	}
 
-	deviceIds := make([]types.String, 0, len(vlan.Instances))
+	deviceIDs := make([]types.String, 0, len(vlan.Instances))
 	for _, device := range vlan.Instances {
-		deviceIds = append(deviceIds, types.StringValue(device.ID))
+		deviceIDs = append(deviceIDs, types.StringValue(device.ID))
 	}
 
-	return m.AssignedDevicesIds.ElementsAs(context.Background(), &deviceIds, false)
+	return m.AssignedDevicesIDs.ElementsAs(context.Background(), &deviceIDs, false)
 }
 
+// ResourceModel defines the schema for managing VLAN resources in Terraform.
+// It is used during create, read, update, and delete operations to represent
+// the desired and actual state of a VLAN in Equinix Metal.
 type ResourceModel struct {
 	ID          types.String `tfsdk:"id"`
 	ProjectID   types.String `tfsdk:"project_id"`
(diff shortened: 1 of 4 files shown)


internal/resources/metal/vlan/resource.go
```diff
@@ -139,19 +161,69 @@ func (r *Resource) Read(ctx context.Context, request resource.ReadRequest, respo
...
 		return
 	}
 
-	if diag := resp.State.Set(ctx, &data); diag.HasError() {
-		resp.Diagnostics.Append(diag...)
+	// Extract the ID of the resource from the state
+	id := plan.ID.ValueString()
+
+	updateRequest := &metalv1.VirtualNetworkUpdateInput{}
+	if !state.Description.Equal(plan.Description) {
```
ctreatma, line 193: The `TestAccMetalVLAN_RemoveDescription` [test is failing](https://github.com/equinix/terraform-provider-equinix/actions/runs/15212419274/job/42792977565?pr=900#step:6:4162):

```
 === NAME  TestAccMetalVlan_RemoveDescription
    resource_test.go:172: Step 2/2 error: Error running apply: exit status 1
        
        Error: Error updating resource
        
          with equinix_metal_vlan.foovlan,
          on terraform_plugin_test.tf line 20, in resource "equinix_metal_vlan" "foovlan":
          20: resource "equinix_metal_vlan" "foovlan" {
        
        Could not update resource with ID d25dc518-7559-4205-8603-0f0213753e06: 422
        Unprocessable Entity Description can't be blank
```

Is there a different API request that should be used to remove a description?  If the API doesn't support removing a VLAN description once it is set and you don't want to change API behavior, the resource schema in this PR could be updated so that adding or changing a description is supported, but removing a description is `ForceNew`.  If you want to take the schema approach I can look for some examples of that.

ctreatma: The test is failing, but the error message is slightly different:

```
 === NAME  TestAccMetalVlan_RemoveDescription
    resource_test.go:172: Step 2/2 error: Error running apply: exit status 1
        
        Error: Error updating resource
        
          with equinix_metal_vlan.foovlan,
          on terraform_plugin_test.tf line 20, in resource "equinix_metal_vlan" "foovlan":
          20: resource "equinix_metal_vlan" "foovlan" {
        
        Could not update resource with ID 76e49a03-ef22-4322-a0e8-12970dc8dcf0: 422
        Unprocessable Entity Update params can't be blank
```

At a glance, I would expect this code to send a non-empty request body when the description is removed, something like `{ "description": "" }`.  It may not be doing that, or it may be sending a request body that we would expect to work, but the API is rejecting it.  One thing to try is running this test locally with debug logging enabled so that you can see what Terraform is sending to the API:

```bash
export TESTS_REGEXP='TestAccMetalVlan_RemoveDescription' # Limits testing to this one test
export METAL_AUTH_TOKEN='<your Metal token>' # Each test creates a project, so this needs to be a user key, not a project key
export EQUINIX_API_TOKEN='not_used_but_sometimes_needed_for_test_startup'
export TF_LOG='debug' # Turn on debug logging; enables wire logging for HTTP clients
make testacc
```

lucasmpb: Yes, I thought that we might not be sending the empty description when its removed. And running the test locally and checking logs, I confirm thats the issue. I'll see how I can fix that.

internal/resources/metal/vlan/models.go
```diff
@@ -96,5 +105,37 @@ func (m *ResourceModel) parse(vlan *packngo.VirtualNetwork) (d diag.Diagnostics)
...
-	return nil
+	return d
+}
+
+func (m *ResourceModel) parseMetalV1(vlan *metalv1.VirtualNetwork) (d diag.Diagnostics) {
+	m.ID = types.StringValue(vlan.GetId())
+	m.Vxlan = types.Int64Value(int64(vlan.GetVxlan()))
+	m.Facility = types.StringValue("")
+
+	if vlan.GetDescription() != "" {
```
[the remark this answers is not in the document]
lucasmpb: Thanks for all your help Charles!

ctreatma:
Note that I changed the Conventional Commit tag in the title from `refactor` to `feat`; this is introducing a new feature that wasn't previously supported, and IIRC our release automation doesn't consider `refactor` to be a releasable change type on its own (i.e., a release that only includes `refactor` commits will be skipped).

nebb00:
Thanks guys

PR #901, opened 2025-05-29
test: Remove spot_market acceptance tests, as Equinix Metal will remove this functionality

lucasmpb:
On June 20, 2025, Equinix Metal will drop support for the Spot Market functionality. To prevent the acceptance tests to fail when that happen, we will remove them proactively.

The spot market functionality will be removed on a subsequent PR closer to the spot market eol date.


stevemar (approved):

stevemar:
i updated the branch as it was out of sync

PR #907, opened 2025-06-10
feat: Fabric Acceptance Tests - Stream Alert Rules and Sweeper Test

d-bhola:
-Added following files:
 `resource_test.go`
`datasources_test.go`
`sweeper.go`

-Added Metric name in READ method
-Update `sweep_test.go`

### DIFF
--- a/internal/resources/fabric/streamalertrule/models.go
+++ b/internal/resources/fabric/streamalertrule/models.go
@@ -1,7 +1,8 @@
-package stream_alert_rule
+package streamalertrule
 
 import (
 	"context"
+
 	"github.com/equinix/equinix-sdk-go/services/fabricv4"
 	"github.com/equinix/terraform-provider-equinix/internal/fabric"
 	int_fw "github.com/equinix/terraform-provider-equinix/internal/framework"
--- a/internal/resources/fabric/streamalertrule/models.go
+++ b/internal/resources/fabric/streamalertrule/models.go
@@ -26,24 +26,24 @@ type streamAlertRuleResourceModel struct {
 }
 
 type baseStreamAlertRulesModel struct {
-	Type              types.String                         `tfsdk:"type"`
-	Name              types.String                         `tfsdk:"name"`
-	Description       types.String                         `tfsdk:"description"`
-	Enabled           types.Bool                           `tfsdk:"enabled"`
-	MetricName        types.String                         `tfsdk:"metric_name"`
-	ResourceSelector  fwtypes.ObjectValueOf[selectorModel] `tfsdk:"resource_selector"` // Object of selectorModel
-	WindowSize        types.String                         `tfsdk:"window_size"`
-	WarningThreshold  types.String                         `tfsdk:"warning_threshold"`
-	CriticalThreshold types.String                         `tfsdk:"critical_threshold"`
-	Operand           types.String                         `tfsdk:"operand"`
-	Href              types.String                         `tfsdk:"href"`
-	UUID              types.String                         `tfsdk:"uuid"`
-	State             types.String                         `tfsdk:"state"`
-	ChangeLog         fwtypes.ObjectValueOf[logModel]      `tfsdk:"change_log"` // Object of logModel
+	Type              types.String                          `tfsdk:"type"`
+	Name              types.String                          `tfsdk:"name"`
+	Description       types.String                          `tfsdk:"description"`
+	Enabled           types.Bool                            `tfsdk:"enabled"`
+	MetricName        types.String                          `tfsdk:"metric_name"`
+	ResourceSelector  fwtypes.ObjectValueOf[selectorModel]  `tfsdk:"resource_selector"` // Object of selectorModel
+	WindowSize        types.String                          `tfsdk:"window_size"`
+	WarningThreshold  types.String                          `tfsdk:"warning_threshold"`
+	CriticalThreshold types.String                          `tfsdk:"critical_threshold"`
+	Operand           types.String                          `tfsdk:"operand"`
+	Href              types.String                          `tfsdk:"href"`
+	UUID              types.String                          `tfsdk:"uuid"`
+	State             types.String                          `tfsdk:"state"`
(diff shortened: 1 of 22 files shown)


d-bhola:
`PASS: TestAccFabricStreamAlertRule_PFCR`
image

internal/resources/fabric/streamalertrule/sweeper.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	fabric := meta.NewFabricClientForTesting(ctx)
+	limit := int32(100)
+
+	streams, _, err := fabric.StreamsApi.GetStreams(ctx).Limit(limit).Execute()
+	if err != nil {
+		return fmt.Errorf("error getting streams list for sweeping fabric streams: %s", err)
+	}
+
+	for _, stream := range streams.GetData() {
```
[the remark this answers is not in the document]
srushti-patl: No, need to delete other resources

internal/sweep/sweep_test.go
```diff
@@ -44,6 +46,7 @@ func addTestSweepers() {
 	project.AddTestSweeper()
 	serviceprofile.AddTestSweeper()
 	ssh_key.AddTestSweeper()
+	streamalertrule.AddTestSweeper()
```
displague, line 49: nit: if the package import were `fabric_stream_alert_rule` this would be more consistent with lines 39-43. This sweeper would be listed in that grouping. 

None of the other sweepers follow that convention, so this is more of an observation rather than a recommendation for change.

d-bhola:
**PASS:** 
image

d-bhola:
Most recent Acceptance Test Passing Screenshot (Local):
image

internal/sweep/sweep_test.go
d-bhola: There are new linting rules in the repo, which do not support underscore in package name. Therefore, I could not adopt the naming convention.

internal/resources/fabric/streamalertrule/sweeper.go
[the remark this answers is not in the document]
d-bhola: I think adding `equinix_fabric_stream` dependency is a good idea!

srushti-patl (approved):

internal/resources/fabric/streamalertrule/models.go
```diff
@@ -79,32 +81,57 @@ type dataSourceAllStreamAlertRulesModel struct {
...
+func (m *streamAlertRuleResourceModel) parse(ctx context.Context, streamAlertRule *fabricv4.StreamAlertRule) diag.Diagnostics {
+	m.ID = types.StringValue(streamAlertRule.GetUuid())
+	diags := m.baseStreamAlertRulesModel.parse(ctx, streamAlertRule)
+	if diags.HasError() {
+		return diags
+	}
+	return diags
+}
+
+func parseAlertRule(ctx context.Context, streamAlertRule *fabricv4.StreamAlertRule,
```
thogarty, line 99: Why was this added? The baseStreamAlertRulesModel parse method handled this already. There's no need to create an additional wrapper around this.

internal/resources/fabric/streamalertrule/resource.go
```diff
@@ -207,22 +208,38 @@ func (r *Resource) Update(
...
 
+	needsUpdate := config.Name.ValueString() != state.Name.ValueString() ||
+		config.Description.ValueString() != state.Description.ValueString() ||
+		config.MetricName.ValueString() != state.MetricName.ValueString() ||
+		config.Operand.ValueString() != state.Operand.ValueString() ||
+		config.WindowSize.ValueString() != state.WindowSize.ValueString() ||
+		config.WarningThreshold.ValueString() != state.WarningThreshold.ValueString() ||
+		config.CriticalThreshold.ValueString() != state.CriticalThreshold.ValueString() ||
+		config.Enabled.ValueBool() != state.Enabled.ValueBool() ||
+		!config.ResourceSelector.IsNull() && !config.ResourceSelector.IsUnknown()
```
thogarty, line 231: What if no value for ResourceSelector is ever given? This last check will always cause an update to occur

internal/resources/fabric/streamalertrule/resource.go
```diff
@@ -258,41 +275,37 @@ func (r *Resource) Update(
 	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
 }
 
-func buildUpdateRequest(ctx context.Context, state streamAlertRuleResourceModel, plan streamAlertRuleResourceModel) (fabricv4.AlertRulePutRequest, diag.Diagnostics) {
+func buildUpdateRequest(ctx context.Context, plan streamAlertRuleResourceModel, request fabricv4.AlertRulePutRequest) (fabricv4.AlertRulePutRequest, diag.Diagnostics) {
```
thogarty, line 278: Why do we pass the request in here? Also, we should generate this with the config, not the plan. You don't need both.

internal/resources/fabric/streamalertrule/resource.go
```diff
@@ -258,41 +275,37 @@ func (r *Resource) Update(
...
+	request.SetName(plan.Name.ValueString())
+	request.SetDescription(plan.Description.ValueString())
+	request.SetMetricName(fabricv4.StreamAlertRuleMetricName(plan.MetricName.ValueString()))
+	request.SetOperand(fabricv4.StreamAlertRuleOperand(plan.Operand.ValueString()))
+	request.SetWindowSize(plan.WindowSize.ValueString())
+
+	if plan.CriticalThreshold.IsNull() && plan.WarningThreshold.IsNull() {
+		return request, diags
 	}
-	if !state.CriticalThreshold.Equal(plan.CriticalThreshold) {
```
thogarty, line 271: Why did we remove all of these checks? Are these all required values?

thogarty (changes requested):
Changes requested

thogarty (approved):
Thanks, @d-bhola

PR #908, opened 2025-06-17
feat: Workload identity federation support for Fabric using STS

rrajagopalan-equinix:


internal/config/config.go
```diff
@@ -357,3 +406,106 @@ func moduleVersionFromBuild(modulePath string) string {
...
+	if workloadIdentityToken == "" {
+		return "", 0, fmt.Errorf("Workload identity token cannot be empty for OIDC token exchange")
+	}
+
+	// Construct the full token exchange URL
+	baseURL, err := url.Parse(stsBaseURL)
+	if err != nil {
+		return "", 0, fmt.Errorf("Failed to parse STS base URL: %w", err)
+	}
+	baseURL.Path = path.Join(baseURL.Path, "/use/token")
```
ctreatma, line 452: There should be an OpenAPI spec for this endpoint.  Given that spec, we can onboard this service to https://github.com/equinix/equinix-sdk-go to encapsulate the request/response details & make it possible for customers to build similar integrations outside of the Terraform provider.

equinix/provider.go
```diff
@@ -83,6 +83,19 @@ func Provider() *schema.Provider {
...
+				DefaultFunc: schema.EnvDefaultFunc(config.AuthScopeEnvVar, ""),
+				Description: "The scope of the authentication token. This argument can also be specified with the `EQUINIX_STS_AUTH_SCOPE` shell environment variable.",
+			},
+			"sts_endpoint": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				DefaultFunc:  schema.EnvDefaultFunc(config.StsEndpointEnvVar, config.DefaultStsBaseURL),
+				ValidateFunc: validation.IsURLWithHTTPorHTTPS,
+				Description:  fmt.Sprintf("The STS API base URL to point out desired environment. This argument can also be specified with the `EQUINIX_STS_ENDPOINT` shell environment variable. (Defaults to `%s`)", config.DefaultStsBaseURL),
+			},
```
ctreatma, line 98: Note: there's a `make docs` task that will update the generated docs for you to match these provider changes

internal/config/config.go
```diff
@@ -150,13 +177,63 @@ func (c *Config) NewFabricClientForFramework(ctx context.Context, meta tfsdk.Con
...
 		ClientID:     c.ClientID,
 		ClientSecret: c.ClientSecret,
 		BaseURL:      c.BaseURL,
 	}
-	authClient := authConfig.New(ctx)
-	//nolint:staticcheck // We should move to subsystem loggers, but that is a much bigger change
+	return authConfig.New(ctx)
+}
+
+func (c *Config) configureHTTPClient(authClient *http.Client) *http.Client {
```
ctreatma, line 236: We can't address the linter's complaint about `logging.NewTransport` in this PR, so the `nolint` comment that used to precede this line should be brought back.

```suggestion
func (c *Config) configureHTTPClient(authClient *http.Client) *http.Client {
    //nolint:staticcheck // We should move to subsystem loggers, but that is a much bigger change
```

internal/config/config.go
```diff
@@ -150,13 +177,63 @@ func (c *Config) NewFabricClientForFramework(ctx context.Context, meta tfsdk.Con
...
+	return c.createFabricClient(httpClient)
+}
+
+func (c *Config) createAuthClient(ctx context.Context) *http.Client {
+	if c.StsAuthScope != "" && c.StsSourceToken != "" {
+		// If the StsAuthScope and StsSourceToken are set, use STS based authentication
+		return c.createStsIdentityClient(ctx, c.StsSourceToken)
+	}
+
+	return c.createOAuthClient(ctx)
```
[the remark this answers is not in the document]
displague: (Which seems fine to me, but we would need to declare that in documentation)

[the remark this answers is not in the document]
ctreatma: @displague I think you're saying the same thing I said?  The docs do not match the code; I stated a preference for what the code currently does rather than what the docs currently say, but regardless of which path is chosen, the docs should reflect what the code does.

internal/config/config.go
```diff
@@ -26,12 +28,15 @@ import (
...
-	ClientTimeoutEnvVar  = "EQUINIX_API_TIMEOUT"
-	MetalAuthTokenEnvVar = "METAL_AUTH_TOKEN"
+	EndpointEnvVar              = "EQUINIX_API_ENDPOINT"
+	ClientIDEnvVar              = "EQUINIX_API_CLIENTID"
+	ClientSecretEnvVar          = "EQUINIX_API_CLIENTSECRET"
+	ClientTokenEnvVar           = "EQUINIX_API_TOKEN"
+	ClientTimeoutEnvVar         = "EQUINIX_API_TIMEOUT"
+	MetalAuthTokenEnvVar        = "METAL_AUTH_TOKEN"
+	AuthScopeEnvVar             = "EQUINIX_STS_AUTH_SCOPE"
+	WorkloadIdentityTokenEnvVar = "TFC_WORKLOAD_IDENTITY_TOKEN"
```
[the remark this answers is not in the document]
displague: I don't know if `TFC` is conventional in HCP usage, but the TFC product is now called HCP Terraform. Terraform Cloud as a branding name has been replaced.

internal/config/config.go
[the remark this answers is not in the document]
ctreatma: Resolving this thread because the docs were updated to match what the code does (i.e., the docs now say that workflow ID settings cannot be used in conjunction with client ID/secret).

internal/provider/provider.go
```diff
@@ -79,6 +78,21 @@ func (p *FrameworkProvider) Schema(
 					int64validator.AtLeast(100),
 				},
 			},
+			"sts_auth_scope": schema.StringAttribute{
+				Optional:    true,
+				Description: "The scope of the authentication token. Must be an access policy ERN or a string of the form roleassignments:<org_id> This argument can also be specified with the `EQUINIX_STS_AUTH_SCOPE` shell environment variable.",
```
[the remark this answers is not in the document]
rrajagopalan-equinix: Addressed in af0a15d1761bea42b95b89d4cf7e61a087db5006

internal/config/config.go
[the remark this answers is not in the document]
ctreatma: Resolving this thread because the direct concern about tying the name of this feature to one specific token source has been addressed.  The TFC/HCP naming concern is a good callout and we should make sure that the latest docs have accurate references to any external systems and (if needed) their environment variables.

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
+# Workload Identity Federation (WIF) using Equinix STS
+
+This guide will walk you through the initial setup of Workload Identity Federation using STS, allowing your workloads to securely authenticate with Equinix services without using long-lived credentials.
```
lchiam, line 3: I would break this para into 2 sentences for readability.

This guide walks you through setting up Workload Identity Federation (WIF) using Equinix STS. It enables your workloads to securely authenticate with Equinix services without relying on long-lived credentials.

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Prerequisites
+
+- An Equinix account with an organization and project - [Sign up here](https://customerportal.equinix.com)
+- Access to the platform you want to use as an identity provider (e.g., Terraform HCP)
```
lchiam, line 8: Simplify by removing "platform" as it's not clear or necessary?

“- Access to your identity provider (e.g., Terraform HCP)”

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+- API credentials for Equinix platform - See [Generating Client ID and Client Secret key](https://developer.equinix.com/dev-docs/fabric/getting-started/getting-access-token#generating-client-id-and-client-secret) for more details
```
lchiam, line 9: Remove mention of platform

"Equinix API credentials - See [Generating Client ID and Client Secret key]"

ctreatma: Adding to Lynn's comment, the link for the client ID/secret docs should be https://docs.equinix.com/equinix-api/api-authentication#generate-client-id-and-client-secret

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+## Troubleshooting
+
+If your workloads fail to authenticate:
+
+1. Verify the trust relationship was established correctly
+2. Check that the workload's identity matches exactly what's specified in your access policies or role assignments
+3. Ensure the required permissions have been granted
+4. Look for any errors in the token exchange process
+
+For additional support, contact Equinix customer service.
```
ctreatma, line 113: Having a troubleshooting section is a good call.  Steps 2-4 look OK to me, but I'm not clear if we have enough detail in step (1) for customers to make progress there.  That may just be my lack of familiarity with the provider config and token exchange processes and related standards.  Is it straightforward to verify that a trust relationship was established correctly or is that a multi-step process that requires looking at specific settings?

Redirecting to customer service for further support sounds reasonable, but are there self-service things the customer can do if they find issues in the steps above?  For example, if they find that the trust relationship wasn't established correctly, or if they had a typo in the access policies or role assignments, can they fix that themselves or do they need customer service to do something for them?

docs/index.md
```diff
@@ -85,4 +100,7 @@ These parameters can be provided in [Terraform variable files](https://www.terra
 - `max_retry_wait_seconds` (Number) Maximum number of seconds to wait before retrying a request.
 - `request_timeout` (Number) The duration of time, in seconds, that the Equinix Platform API Client should wait before canceling an API request. Canceled requests may still result in provisioned resources. (Defaults to `30`)
 - `response_max_page_size` (Number) The maximum number of records in a single response for REST queries that produce paginated responses. (Default is client specific)
+- `sts_auth_scope` (String) The scope of the authentication token. Must be an access policy ERN or a string of the form `roleassignments:<org_id>`. This argument can also be specified with the `EQUINIX_STS_AUTH_SCOPE` shell environment variable. Please note that Equinix STS is an alpha feature and not available for all users.
+- `sts_endpoint` (String) The STS API base URL to point out desired environment. This argument can also be specified with the `EQUINIX_STS_ENDPOINT` shell environment variable. (Defaults to `https://sts.eqix.equinix.com`). Please note that STS is an alpha feature and not available for all users.
```
displague, line 104:
```suggestion
- `sts_endpoint` (String) The STS API base URL to point to the desired environment. This argument can also be specified with the `EQUINIX_STS_ENDPOINT` shell environment variable. (Defaults to `https://sts.eqix.equinix.com`). Please note that STS is an alpha feature and not available for all users.
```

Copilot offered this as a nitpick

docs/guides/sts_wif_setup.md
rrajagopalan-equinix: There are apis to verify and fix trust relationships, access policies and role assignments. I plan to add links to api docs when they are ready

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Step 3: Authorize Your Workloads
+
+You can authorize workloads using either role assignments or access policies:
+
+### Option A: Using Role Assignments
+
+```bash
+# First get a JWT token
+JWT=$(curl -s "https://api.equinix.com/oauth2/v1/userinfo" \
```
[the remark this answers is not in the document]
rrajagopalan-equinix: This has existing for a while as per my understanding

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+  --json '{
+    "accessPolicyId": "accesspolicy:your-policy-name",
+    "grants": [
+      "principal:'$ORG_ID':'${IDP_ID:4}':{subject}"
+    ],
+    "tags": {},
+    "permissions": [{
+      "serviceActions": [{
+        "serviceId": "Equinix Service ID", 
+        "actions": ["Action1", "Action2"]
```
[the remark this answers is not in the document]
rrajagopalan-equinix: There are apis to discover service ids and list actions for a given service, will add links to api docs when they are ready

internal/acceptance/acceptance.go
```diff
@@ -20,9 +24,13 @@ const (
 )
 
 var (
-	TestAccProvider          *schema.Provider
-	TestAccProviders         map[string]*schema.Provider
-	TestExternalProviders    map[string]resource.ExternalProvider
+	// TestAccProvider is the Equinix provider instance used for acceptance testing
+	TestAccProvider *schema.Provider
+	// TestAccProviders is a map of provider names to their schema.Provider instances used in acceptance tests.
+	TestAccProviders map[string]*schema.Provider
```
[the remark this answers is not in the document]
ctreatma: The map is already initialized in `init()`. https://github.com/equinix/terraform-provider-equinix/pull/908/files/9570aa47b0df53f336bb3bc09f203ee0f5a76417#diff-a7084a87154a2249e8cc364137ada5573520dc434b8138618a749ac4f41d545eL38

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,113 @@
...
+  -H "Authorization: Bearer $TOKEN" \
+  | jq -r '.jwt_token')
+
+# Create role assignment
+curl -s "https://api.equinix.com/am/v3/assignments" \
+  -H "Authorization: Bearer $JWT" \
+  --json '{
+    "principal": {
+      "type": "FEDERATED",
+      "name": "principal:'$ORG_ID':'${IDP_ID:4}':{subject}"
```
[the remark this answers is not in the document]
ctreatma: In the meantime, I think it'd be good to specifically mention that the `idp:` prefix needs to be removed in Equinix access policy principals.  You could add a sentence to the paragraph where you [mention the requirement for the `subject` value in the principal](https://github.com/equinix/terraform-provider-equinix/pull/908/files#diff-3863d4c3c06fdf2fd239a5ce5ecec057e3fb08e8d6a1036537fe4a344a72fafdR54).

I think it would also help to set an explicit environment variable for the un-prefixed IdP ID so you can call out the behavior again in the example code; something like:

```bash
PRINCIPAL_IDP_ID =  "${IDP_ID:4}" # Remove `idp:` prefix from IdP ID for use in policy principal
```

docs/guides/sts_wif_setup.md
```diff
@@ -0,0 +1,112 @@
...
+      "id": "'$ORG_ID'",
+      "type": "ORGANIZATION"
+    }
+  }'
+```
+
+### Option B: Using Access Policies
+
+```bash
+ACCESS_URL="https://access.equinix.com"
```
ctreatma, line 83: Similar to the comment about `STS_URL` on an earlier commit:  if we generally expect Equinix customers will use this URL and not a different one, we could remove this environment variable declaration and hard-code the hostname in the `curl` example below.

.github/workflows/fabric_acctest.yml
```diff
@@ -44,6 +46,79 @@ jobs:
     steps:
       - run: true
 
+  configure-oidc-template:
+    name: Configure OIDC Subject Claim Template
+    needs: authorize
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure OIDC Subject Claim Template
```
ctreatma, line 54: Is this something that only needs to be done once per repo, or does the template get reset periodically?  If it's a one-time action and doesn't have to be executed every time an OIDC token is requested, we should document the requirement somewhere but not include this step--or the `actions: write` permission--in the workflow.

.github/workflows/fabric_acctest.yml
```diff
@@ -196,7 +242,21 @@ jobs:
...
+      - name: TF Fabric PFCR acceptance tests STS creds
+        timeout-minutes: 180
+        env:
+          TF_ACC: "1"
+          TF_ACC_FABRIC_CONNECTIONS_TEST_DATA: ${{ secrets.TF_ACC_FABRIC_CONNECTIONS_TEST_DATA }}
+          TF_ACC_FABRIC_DEDICATED_PORTS: ${{ secrets.TF_ACC_FABRIC_DEDICATED_PORTS }}
+          TF_ACC_FABRIC_MARKET_PLACE_SUBSCRIPTION_ID: ${{ secrets.TF_ACC_FABRIC_MARKET_PLACE_SUBSCRIPTION_ID }}
+          TF_ACC_FABRIC_STREAM_TEST_DATA: ${{ secrets.TF_ACC_FABRIC_STREAM_TEST_DATA }}
+          EQUINIX_STS_SOURCE_TOKEN: ${{ steps.get_id_token.outputs.id_token }}
+          EQUINIX_STS_AUTH_SCOPE: "roleassignments:405153000683548"
```
d-bhola, line 255: Is there a way to get `EQUINIX_STS_AUTH_SCOPE `dynamically as well ?

.github/workflows/fabric_acctest.yml
```diff
@@ -196,7 +242,21 @@ jobs:
...
+          METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
+        run: |
```
d-bhola, line 257: Since we are already running the entire test suite through client id and secret approach, we can just add to run 1-2 tests here.

.github/workflows/fabric_acctest.yml
d-bhola: Yes, I meant moving to secrets. Its not a sensitive value, then its fine.

.github/workflows/fabric_acctest.yml
```diff
@@ -162,6 +166,20 @@ jobs:
...
     steps:
+      - id: get_id_token
+        name: Get GitHub OIDC Token for PFCR
+        uses: actions/github-script@v6
+        with:
+          script: |
+            try {
+              const idToken = await core.getIDToken('gha-fcr-client');
+              console.log('Token generated with audience: gha-fcr-client');
+              core.setOutput('id_token', idToken);
```
displague, line 177: Should this be using core.setSecret instead (which would influence how the value is referenced too)

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error

displague:
With #922 we have a major release to v4 queued up. I'm wondering if this PR should be queued for that same major.

(@ctreatma @sbhatnagar-equinix )

.github/workflows/fabric_acctest.yml
```diff
@@ -197,31 +215,26 @@ jobs:
...
-        if: ${{ always() }}
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
-        with:
-          name: pfcr_test_logs
-          path: pfcr_test_output.json
-
-      - name: Sweeper PFCR
-        if: ${{ always() }}
+      - name: TF Fabric PFCR acceptance tests STS creds
+        timeout-minutes: 180
```
ctreatma, line 219: The Fabric acceptance test workflow was recently updated so that the [TF Fabric PFCR acceptance tests](https://github.com/equinix/terraform-provider-equinix/actions/runs/17472658879/job/49624787652) step fails if the acceptance tests fail, which means that the STS creds test step is now being skipped:  https://github.com/equinix/terraform-provider-equinix/actions/runs/17472658879/job/49624787652

To allow the STS creds tests to run even if the other tests fail, you would need to add an if statement to this step.  Specifying `if: always()` is the fastest way to do that, but you could consider using a more specific conditional so that the STS creds tests only run if the `get_id_token` step succeeded.

ctreatma: An adjustment to my earlier suggestion:  rather than modifying how this test step runs, I recommend copying the `test-PFCR` job in its entirety, renaming it to `test-STS-creds` (or similar) and updating the acceptance test step to run the intended tests.  If/when all tests are able to run using STS token exchange, the Fabric team can choose to update and/or consolidate the test jobs as needed.

.github/workflows/fabric_acctest.yml
```diff
@@ -204,24 +206,70 @@ jobs:
...
 
-      - name: Sweeper PFCR
-        if: ${{ always() }}
-        env:
-          EQUINIX_API_CLIENTID: ${{ secrets.EQUINIX_API_CLIENTID_PFCR }}
-          EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PFCR }}
-          METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
-          SWEEP: "all" #Flag required to define the regions that the sweeper is to be ran in
-          SWEEP_ALLOW_FAILURES: "true" #Enable to allow Sweeper Tests to continue after failures
+  test-STS-creds:
```
ctreatma, line 209: The test sweeper step was accidentally deleted when the `test-STS-creds` job was introduced and it will need to be restored.

```suggestion
      - name: Sweeper PFCR
        if: ${{ always() }}
        env:
          EQUINIX_API_CLIENTID: ${{ secrets.EQUINIX_API_CLIENTID_PFCR }}
          EQUINIX_API_CLIENTSECRET: ${{ secrets.EQUINIX_API_CLIENTSECRET_PFCR }}
          METAL_AUTH_TOKEN: ${{ secrets.METAL_AUTH_TOKEN }}
          SWEEP: "all" #Flag required to define the regions that the sweeper is to be ran in
          SWEEP_ALLOW_FAILURES: "true" #Enable to allow Sweeper Tests to continue after failures

  test-STS-creds:
```

ctreatma:
@equinix/governor-digin-fabric you have been tagged for review on this as codeowners.  Please take a look when you are able.

PR #925, opened 2025-07-30
chore: update release commit to be used in github actions

sbhatnagar-equinix:
This is to update the sha for a github action instead of version to fix a vulnerability.


adammohammed (approved):

.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     - name: Release
       id: release
-      uses: cycjimmy/semantic-release-action@v4
+      uses: cycjimmy/semantic-release-action@9a24fc62fd6dbf0c523b81421b634ee3c2c2ad55
```
lucasmpb, line 23: Might be a good idea to add the version at the end of the hash, like the other actions, to make it easier to know which version we are using.
```suggestion
      uses: cycjimmy/semantic-release-action@9a24fc62fd6dbf0c523b81421b634ee3c2c2ad55 # v4
```

.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
...
       with:
         extra_plugins: |
           conventional-changelog-conventionalcommits@8.0.0
```
lucasmpb, line 26: Do we also need to update this reference?

lucasmpb (dismissed):
Looks good overall. Left a couple small comment :)

sbhatnagar-equinix:
> Renovate is already configured to replace GitHub Actions version tags with SHAs, although it doesn't seem to have gotten to this particular one yet (it's probably waiting on some of its open PRs to be merged first).
> 
> Can you provide more context as to the security concern for this specific action?

Hi @ctreatma , this relates to reference confusion vulnerability (https://nvd.nist.gov/vuln/detail/CVE-2024-34394). Solution is to use actual commit sha for the github actions where not used. I think once we update the sha renovate automatically tracks and updates as new commits for the respective version come in ? If yes, then we can go ahead with this change since renovate didn't update for this particular github action. Let me know your thoughts.

sbhatnagar-equinix:
> The CVE you referenced is what led to the updated Renovate settings; I was wondering if there was a new issue specific to this PR or if the security concern was the one we addressed with the Renovate change.
> 
> It looks like in this specific case, Renovate failed to convert from version to digest because `v4` is a branch, not a tag, for the semantic release action. Your change looks like it pulls in the latest commit sha for the `main` branch, which is likely not what we want; I'm not clear how Renovate will handle that going forward, but it _might_ just keep pulling in the latest from `main` which would not be good for stability. It would be better to adopt a tag--it looks like `v4.2.2` is the tag equivalent to what we've been using so far--and let Renovate do the SHA pinning afterwards so we can be more confident that Renovate will do what we expect it to do going forward.

Yes, this makes sense. I haven't dealt with renovate

> The CVE you referenced is what led to the updated Renovate settings; I was wondering if there was a new issue specific to this PR or if the security concern was the one we addressed with the Renovate change.
> 
> It looks like in this specific case, Renovate failed to convert from version to digest because `v4` is a branch, not a tag, for the semantic release action. Your change looks like it pulls in the latest commit sha for the `main` branch, which is likely not what we want; I'm not clear how Renovate will handle that going forward, but it _might_ just keep pulling in the latest from `main` which would not be good for stability. It would be better to adopt a tag--it looks like `v4.2.2` is the tag equivalent to what we've been using so far--and let Renovate do the SHA pinning afterwards so we can be more confident that Renovate will do what we expect it to do going forward.

I updated it to use tag instead of sha. I am not sure how to trigger renovate. Can you please review again?

ctreatma (approved):

PR #926, opened 2025-07-31
fix: Fix html report

tutkat:


.github/workflows/fabric_acctest.yml
```diff
@@ -198,6 +199,7 @@ jobs:
           go test ./... -v -coverprofile coverage_pfcr.txt -covermode=atomic -count 1 -parallel 8 -run "(PFCR)" -timeout 180m | tee pfcr_test_output.log
           
       - name: Upload PFCR Testing Log
+        if: ${{ always() }}
```
thogarty, line 202: Do we need these if statements?

.github/workflows/fabric_acctest.yml
```diff
@@ -119,6 +119,7 @@ jobs:
           go test ./... -v -coverprofile coverage_pnfv.txt -covermode=atomic -count 1 -parallel 8 -run "(PNFV)" -timeout 180m | tee pnfv_test_output.log
 
       - name: Upload PNFV Testing Log
+        if: ${{ always() }}
```
thogarty, line 122: Do we need these if statements?

.github/workflows/fabric_acctest.yml
```diff
@@ -259,22 +261,27 @@ jobs:
           name: pfcr_test_logs
 
       - name: Create HTML Testing Report
+        if: always()
```
thogarty, line 264: Do we need these if statements?

.github/workflows/fabric_acctest.yml
```diff
@@ -300,4 +307,4 @@ jobs:
           name: UAT Terraform Acceptance Test Reports
           path: |
             uat_test_report.html
-          compression-level: 0
+          compression-level: 0
```
thogarty, line 310: [nitpick] needs newline

thogarty (approved):
I love it!

.github/workflows/fabric_acctest.yml
tutkat: removed

.github/workflows/fabric_acctest.yml
tutkat: removed

scripts/check_tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
ctreatma, line 1: The acceptance test steps already fail if the tests fail, and that failure is reflected in both the summary view of the GitHub Actions workflow (there's a red circle with an x next to the step if it fails) and in the console output for the test step.  Rather than add this missing check script, why not remove the references to it instead?

ctreatma (commented):
I added a couple suggestions/comments on specific parts of this PR, but looking at it overall, I think this sort of thing might be better handled outside of this repository.  An alternative approach--one that would remove the need to get these changes through cross-team review--would be to set up [a `workflow_run` webhook on this repository](https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run).  In that case your workflow would upload the unmodified test logs as artifacts and the webhook receiver would filter for `action=completed` and `workflow.name=<your workflows>`, download the workflow artifacts, transform them as needed, and publish them somewhere (an S3 bucket, a GitHub Pages site, etc.).

.github/workflows/fabric_acctest.yml
```diff
@@ -259,22 +259,26 @@ jobs:
           name: pfcr_test_logs
 
       - name: Create HTML Testing Report
+        shell: bash
         run: |
           cat pfcr_test_output.log pnfv_test_output.log > uat_test_report.log
-          go-junit-report -in uat_test_report.log > uat_test_report.xml && python3 -m junit2htmlreport uat_test_report.xml uat_test_report.html
+          python3 generate_test_report.py uat_test_report.log uat_test_report.html
```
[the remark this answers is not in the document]
ctreatma: Coming back to this to suggest that you could also with with [JSON test output](https://pkg.go.dev/cmd/test2json#hdr-Output_Format) (from `go test -json` or `gotestsum --jsonfile example.json`).  Parsing JSON will be much more resilient than parsing plain text.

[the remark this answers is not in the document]
tutkat: ok I modified to have output as json

ctreatma (approved):

PR #935, opened 2025-08-07
CXF-11376: Update terraform provider docs and test with new sink types

d-bhola:
Updated test - `TestAccFabricStreamSubscription_PFCR`. Run locally: 
Pasted Graphic


internal/resources/fabric/stream_subscription/resource_test.go
```diff
@@ -142,6 +142,58 @@ func testAccFabricStreamSubscriptionConfig(streamTestData map[string]map[string]
 		  }
 		}
 
+		resource "equinix_fabric_stream_subscription" "servicenow" {
```
thogarty, line 145: For the tests, there's a maximum limit of subscriptions per stream. Have you tested this to make sure it's working with the number of new streams on it? I thought it was max 3 subscriptions per stream.

thogarty (commented):
See question about max subscriptions per stream.

internal/resources/fabric/stream_subscription/resource_test.go
d-bhola: I modified it now. You're correct. Just confirmed its 3 subscriptions only per stream.

thogarty (approved):
LGTM.

PR #1029, opened 2026-06-18
feat: implement searching for route filter rules datasource

tng1-equinix:


internal/resources/fabric/route_filter_rule/datasources.go
```diff
@@ -44,41 +45,95 @@ Additional Documentation:
...
+			sortItem.SetDirection(fabricv4.RouteFilterRuleSortDirection(direction.(string)))
+		}
+		if property, ok := sortInput["property"]; ok {
+			sortItem.SetProperty(fabricv4.RouteFilterRuleSortBy(property.(string)))
+		}
+
+		reqSortParam = append(reqSortParam, *sortItem)
+	}
+	searchRequest.SetSort(reqSortParam)
+
```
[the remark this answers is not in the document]
tng1-equinix: We already type cast it to `[]any`, so it shouldn't be empty.

internal/resources/fabric/route_filter_rule/datasources_schema.go
```diff
@@ -136,6 +174,28 @@ func dataSourceAllRulesForRouteFilterSchema() map[string]*schema.Schema {
 			Default:     20,
 			Description: "Number of elements to be requested per page. Number must be between 1 and 100. Default is 20",
 		},
+		"sort": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "Filters for the Data Source Search Request",
+			Elem: &schema.Resource{
```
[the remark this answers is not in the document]
tng1-equinix: This is inherited from connection.

Update this and the connections misnaming

shahbhoomi (dismissed):

tng1-equinix:
```
$ TF_ACC=1 TF_LOG=debug go test ./... -run=TestAccFabricRouteFilterRule_DataSources_PFCR
ok      github.com/equinix/terraform-provider-equinix/internal/resources/fabric/route_filter_rule       233.920s
```

internal/slice/slice.go
```diff
@@ -0,0 +1,11 @@
+package slice
+
+func Map[T any, U any](xs []T, mapFunc func(T) U) []U {
```
displague, line 4:
```suggestion
// Map returns a new slice containing the results of applying mapFunc to each element of xs.
func Map[T any, U any](xs []T, mapFunc func(T) U) []U {
	if xs == nil {
		return nil
	}
	
```

If you would prefer to preseve nil<->nil behavior.

Which ever behavior is desired should be covered by tests.

tng1-equinix: Ok, updated the code and added the tests

internal/slice/slice.go
```diff
@@ -0,0 +1,16 @@
+package slice
```
displague, line 1: We may want to move comparisons.SlicesMatch https://github.com/equinix/terraform-provider-equinix/blob/main/internal/comparisons/comparisons.go#L41-L56 into this package. (seperate PR)

internal/resources/fabric/route_filter_rule/datasources_test.go
```diff
@@ -1,79 +1,226 @@
...
 
 	"github.com/equinix/terraform-provider-equinix/internal/acceptance"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
+const config = `
```
displague, line 14: nit: This config block is rather large and could be split into relevant configs for setup (the resources, which could be, or may already be, defined in the adjacent resources_test.go) and the pieces tested seperately or incrementally if necessary.

tng1-equinix: Yea, we're considering moving it out to it's own `.tf` file, which can separate out the config

but the drawback is that the test case data becomes further from the assertions, which may make analyzing the test harder, we'll go forwards and see which is better, then we'll improve all the test cases 

we're also thinking if it makes sense for datasource and resource test to be together, since datasource test requires the creation of a resource anyways

displague: >  for datasource and resource test to be together

They are already in the same package, there's not a big reason to combine them from datasource_test -> resource_test or vice versa. Having a seperate templates_test configs_test could help, or a seperate package if the reuse of configs needs to extend beyond this package (something like https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/acctest/configs.go).

I wouldn't advise using a .tf file directly, this doesn't fit best practices over the past few years (unless that has changed).

displague (approved):

shahbhoomi (approved):

PR #1030, opened 2026-06-24
feat: Upgrade to sdk-go v0.66.0

tng1-equinix:


srushti-patl (approved):

internal/resources/fabric/port/resource.go
```diff
@@ -262,9 +262,8 @@ func buildCreateRequest(ctx context.Context, plan resourceModel) (fabricv4.PortR
 		request.SetConnectivitySourceType(fabricv4.PortConnectivitySourceType(plan.ConnectivitySourceType.ValueString()))
 	}
 	request.SetLagEnabled(plan.LagEnabled.ValueBool())
-	request.SetName(plan.Name.ValueString())
```
ctreatma, line 265: Are users no longer allowed to set the name when creating a port?

internal/resources/fabric/port/resource.go
```diff
@@ -275,24 +274,11 @@ func buildCreateRequest(ctx context.Context, plan resourceModel) (fabricv4.PortR
...
-		if diags.HasError() {
-			mDiags.Append(diags...)
-			return fabricv4.PortRequest{}, mDiags
-		}
-		portSettings := fabricv4.PortSettings{}
-		portSettings.SetPackageType(fabricv4.PortSettingsPackageType(settings.PackageType.ValueString()))
-		portSettings.SetSharedPortType(settings.SharedPortType.ValueBool())
-		request.SetSettings(portSettings)
-	}
-
```
ctreatma, line 295: Are these settings still available in the API or is the functionality completely gone?

go.sum
```diff
@@ -38,16 +38,8 @@ github.com/dnaeon/go-vcr v1.2.0 h1:zHCHvJYTMh1N7xnV7zf1m1GPBF9Ad0Jk/whtQ1663qI=
...
-github.com/equinix/equinix-sdk-go v0.64.0 h1:gGHSorW5lfXBjDZixWhYgJ8DSIv06DYm3U835yjGD1Y=
-github.com/equinix/equinix-sdk-go v0.64.0/go.mod h1:QokAmUtlYlD4gJ1s5UL1nZ4e6XALV0ftl5ZCwdPYp5M=
-github.com/equinix/equinix-sdk-go v0.64.1 h1:hCkctXS3gROmepRZkHT2M+jb0Unv6loY19G/dtaKmY4=
-github.com/equinix/equinix-sdk-go v0.64.1/go.mod h1:QokAmUtlYlD4gJ1s5UL1nZ4e6XALV0ftl5ZCwdPYp5M=
-github.com/equinix/equinix-sdk-go v0.64.2 h1:8XHgeQ3/BseK9kcyylCF3lfrYiQydHESw0oZtIuLjJY=
-github.com/equinix/equinix-sdk-go v0.64.2/go.mod h1:QokAmUtlYlD4gJ1s5UL1nZ4e6XALV0ftl5ZCwdPYp5M=
-github.com/equinix/equinix-sdk-go v0.65.0 h1:BHiWO7sYG5JO4jX+xLL3OmKBIa9aMd9Y+/Lsq4JxZoI=
-github.com/equinix/equinix-sdk-go v0.65.0/go.mod h1:QokAmUtlYlD4gJ1s5UL1nZ4e6XALV0ftl5ZCwdPYp5M=
+github.com/equinix/equinix-sdk-go v0.66.0 h1:2t0espzvXMKrskOEGCGILXYI+hU4DtWe20m55C1tIDI=
+github.com/equinix/equinix-sdk-go v0.66.0/go.mod h1:QokAmUtlYlD4gJ1s5UL1nZ4e6XALV0ftl5ZCwdPYp5M=
```
ctreatma, line 42: This should be https://github.com/equinix/equinix-sdk-go/releases/tag/v0.65.0

ctreatma (dismissed):

internal/resources/fabric/port/resource.go
tng1-equinix: The issue is that the Port API completely removed the `name `field.

https://docs.equinix.com/api-catalog/fabricv4/#tag/Ports/operation/createPort

From the Terraform perspective, there is no way to supply the name since the underlying API does not support it any longer.

internal/resources/fabric/port/resource.go
tng1-equinix: Same as above, the API completely removed this field

go.sum
tng1-equinix: Ah, yea, there was an internal discussion with the team about this.

The issue is that there was a bad release of v0.65.0 (evident by the fact that there is v0.65.0 in the `go.sum` of the current master.

So the go package mirror already has the checksum of this bad v0.65.0 release, which causes us to be unable to use v0.65.0

Thus, the bypass this, the team agreed to use v0.66.0 instead

internal/resources/fabric/port/resource.go
ctreatma: OK, thanks for confirming.  In the future, if you can include some details in the PR description to document why the changes are being made, it will help guide review.

go.sum
ctreatma: I see.  In the future, please keep in mind that the SDK represents multiple services and release issues should not be resolved by one team without coordination.  For now we can stick with v0.66.0 (though note that because the release did not follow standard procedures, traffic from v0.66.0 will identify itself as v0.65.0 via the User-Agent header).

ctreatma (approved):

PR #1036, opened 2026-06-30
fix: UAT acceptance test cases

tng1-equinix:
fix: add connections deps to resource sweepers which depends on connections
fix: make filter more specific to RA datasource to avoid retrieving RAs from other test
fix: skip trying to sweep `DO_NOT_DELETE` Fabric resources

### DIFF
--- a/internal/resources/fabric/cloud_router/sweeper.go
+++ b/internal/resources/fabric/cloud_router/sweeper.go
@@ -16,7 +16,7 @@ import (
 func AddTestSweeper() {
 	resource.AddTestSweepers("equinix_fabric_cloud_router", &resource.Sweeper{
 		Name:         "equinix_fabric_cloud_router",
-		Dependencies: []string{},
+		Dependencies: []string{"equinix_fabric_connection"},
 		F:            testSweepCloudRouters,
 	})
 }
--- a/internal/resources/fabric/routeaggregation/datasource_test.go
+++ b/internal/resources/fabric/routeaggregation/datasource_test.go
@@ -13,7 +13,7 @@ import (
 var datasourceConfig = `
 resource "equinix_fabric_route_aggregation" "new_ra_1" {
   type        = "BGP_IPv4_PREFIX_AGGREGATION"
-  name        = "route_agg_1_PFCR"
+  name        = "RA_datasource_1_PFCR"
   description = "route_agg_1_PFCR_description"
   project = {
     project_id = "33ec651f-cc99-48e0-94d3-47466899cdc7"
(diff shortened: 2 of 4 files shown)


internal/resources/fabric/cloud_router/sweeper.go
```diff
@@ -16,7 +16,7 @@ import (
 func AddTestSweeper() {
 	resource.AddTestSweepers("equinix_fabric_cloud_router", &resource.Sweeper{
 		Name:         "equinix_fabric_cloud_router",
-		Dependencies: []string{},
+		Dependencies: []string{"equinix_fabric_connection"},
```
tng1-equinix, line 19: Sweeper is failing to cleanup this resource because it is unable to delete it due to active connections

internal/resources/fabric/routeaggregation/datasource_test.go
```diff
@@ -40,14 +40,14 @@ data "equinix_fabric_route_aggregations" "data_ras" {
   filter = {
     property = "/name"
     operator = "LIKE"
-    values   = ["%_PFCR"]
```
tng1-equinix, line 43: This test sometimes fails because other Acceptance tests creates RA with the same `_PFCR` suffix, which gets picked up by the datasource.

Tightening the formatting to filter out the resources generated by the other tests

srushti-patl (dismissed):

internal/sweep/sweep.go
```diff
@@ -27,6 +27,10 @@ func IsSweepableTestResource(namePrefix string) bool {
 
 func IsSweepableFabricTestResource(resourceName string) bool {
 	for _, suffix := range FabricTestResourceSuffixes {
+		if strings.Contains(resourceName, "DO_NOT_DELETE") {
```
ctreatma: I would recommend adopting a single, test-specific naming convention for resources that are created during acceptance tests and can be swept.  The existing `IsSweepableTestResource` function above implements that approach: any resource whose name starts with `tfacc` is considered a resource that is fully managed by Terraform acceptance tests and can therefore be swept if needed after tests complete.

Resources that are created outside of the acceptance test suite should not use the `tfacc` prefix in the name, and they will then be ignored by the sweeper.

tng1-equinix: Fair enough, hmmm but when I search for usages, it seems like `IsSweepableTestResource` is not used at all

EDIT: oh seems like they are removed together with the metal side of things

srushti-patl (approved):

internal/sweep/sweep.go
tng1-equinix: Yep, fair enough, the team decided to rename this resource instead of modifying the sweep conditions. Thanks

PR #1039, opened 2026-07-01
fix: various UAT acceptance tests

tng1-equinix:

PR #1043, opened 2026-07-04
chore: Vlan no sweep

tng1-equinix:
This pull request introduces several improvements and refactorings to the acceptance tests and test helpers for Equinix Fabric resources. The most significant changes include the introduction of helper functions to dynamically allocate available VLANs during tests, updates to test assertions for greater flexibility and coverage, and workflow enhancements for targeted test sweeps. These changes aim to make the tests more robust, less brittle, and easier to maintain.

**Test infrastructure enhancements:**

* Added `RandomVlan` and supporting helpers in `internal/fabric/testing_helpers/vlans.go`, along with unit tests in `vlans_test.go`, to dynamically select available VLANs for test resources, reducing the likelihood of conflicts and test failures. [[1]](diffhunk://#diff-74439344f8f963d6b36f49b48e9fddf2a25d1b1f1c0c58a830a822944bdf263aR1-R76) [[2]](diffhunk://#diff-e3503b57a363444f94634b61f180bafe9ae495f45c901093e1152715afee97aeR1-R43)
* Updated multiple acceptance tests (e.g., `resource_fabric_routing_protocol_acc_test.go`, `connection/datasources_test.go`) to use dynamically allocated VLANs via the new helpers, replacing hardcoded VLAN values and improving test reliability. [[1]](diffhunk://#diff-3ee129d496ea086e9cf9dcd2bb9718ab2c8fbb0a635fe557546cfd8170e1a8dbR33-R45) [[2]](diffhunk://#diff-3ee129d496ea086e9cf9dcd2bb9718ab2c8fbb0a635fe557546cfd8170e1a8dbL94-R100) [[3]](diffhunk://#diff-3ee129d496ea086e9cf9dcd2bb9718ab2c8fbb0a635fe557546cfd8170e1a8dbL156-R162) [[4]](diffhunk://#diff-3ee129d496ea086e9cf9dcd2bb9718ab2c8fbb0a635fe557546cfd8170e1a8dbL209-R215) [[5]](diffhunk://#diff-8cd0af9cae8e9179e2385e53bd7a51e2bed78adaf0223a1c4e44faa176401dd9R22-R132)

**Test assertion improvements:**

* Introduced `ExpectKnownAttributes` and `ExpectKnownAttributesAt` helpers in `internal/fabric/testing_helpers/check.go` to simplify and standardize state assertions in tests.
* Refactored test assertions in several test files to use more expressive checks, including object partial and list exact matching, for better coverage and maintainability. [[1]](diffhunk://#diff-8cd0af9cae8e9179e2385e53bd7a51e2bed78adaf0223a1c4e44faa176401dd9R22-R132) [[2]](diffhunk://#diff-e4d73adb2d6c5564844736b6442b7fa15e7e1d234e92ce11b5b0744910f3aaacL128-R126) [[3]](diffhunk://#diff-e4d73adb2d6c5564844736b6442b7fa15e7e1d234e92ce11b5b0744910f3aaacR159-R175)

**Test configuration and workflow updates:**

* Updated `.github/workflows/fabric_acctest.yml` to add `SWEEP_TARGET` variables for PFNV and PFCR, enabling more targeted and flexible test sweeps in CI. [[1]](diffhunk://#diff-e0d73dde80a3d36f7c76d2e300b497b616f2790cef94c291106bb345e4a027e3R139) [[2]](diffhunk://#diff-e0d73dde80a3d36f7c76d2e300b497b616f2790cef94c291106bb345e4a027e3R214)
* Refactored test configuration blocks to use input variables for port UUIDs, VLANs, and other parameters, improving test reusability and clarity. [[1]](diffhunk://#diff-e4d73adb2d6c5564844736b6442b7fa15e7e1d234e92ce11b5b0744910f3aaacR16-L19) [[2]](diffhunk://#diff-e4d73adb2d6c5564844736b6442b7fa15e7e1d234e92ce11b5b0744910f3aaacL78-L90) [[3]](diffhunk://#diff-8cd0af9cae8e9179e2385e53bd7a51e2bed78adaf0223a1c4e44faa176401dd9R22-R132)

**General test cleanup:**

* Removed redundant or obsolete fields and expectations, such as `ExpectNonEmptyPlan: false`, from various test cases for clarity and to align with updated test logic. [[1]](diffhunk://#diff-aa844974d00b1ac118f5280b7bf46a13d9618d5eeb5f601fa2b50cb37298f9bbL60) [[2]](diffhunk://#diff-0885364531e7d96fc5f50fa418a30c7fcd974919e13db62ef9666852a9ca256fL57-L65) [[3]](diffhunk://#diff-66f26874e00c638bc1f76f9a264bba13d5e1c6102dc224d4fc5066d6e3458db3L42)

These changes collectively improve the reliability, maintainability, and clarity of the acceptance test suite for Equinix Fabric resources.


equinix/resource_fabric_cloud_router_acc_test.go
```diff
@@ -54,15 +54,13 @@ func TestAccCloudRouterCreateOnlyRequiredParameters_PFCR(t *testing.T) {
 					resource.TestCheckResourceAttrSet("equinix_fabric_cloud_router.test", "change_log.0.updated_date_time"),
 					resource.TestCheckResourceAttrSet("equinix_fabric_cloud_router.test", "change_log.0.deleted_date_time"),
 				),
-				ExpectNonEmptyPlan: false,
```
displague, line 57: The reason for these removals was not included in the PR description (there was none, I generated one, but that doesn't provide reason).

It would also be helpful to see these resources move from equinix/ to the internal/fabric folder, and in that move these checks would change https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks

displague: Also - why are we removing the parallel testing. 

From what I can gather there is some concern about sweepers at the end of one PR test sweeping resources held by currently in-use tests of another PR. The sweepers are intended to capture resources outside of their immediate testing, left behind by other runs. 

There are other ways we could approach protecting the resources of the current test, including GitHub workflow tweaks.

I would also consider if we could change the E2E test asset prefix to contain a 3+ random/unique(enough) character suffix to denote the resources created through this test run (`tfacc_xyz_`). We have to be careful with prefix length because many resources have name length constraints and we have bumbed into these in the past. 

The regular sweeper would only remove resources of this PR (`tfacc_xyz_*`), and a general sweeper workflow would run (perhaps via GH cron) to sweep under the shared prefix (`tfacc*`).

There also seems to be some removal of shared/static test assets (knownvalue datasources), but I may hae misread this. If the intent is to dynamically register more resource dependencies (create a VLAN for each PR run/step, rather than reusing a shared one across all PRs) this can help with validating the overall reliability of the provider against the API.

displague: This seems to be what is happening in #1044 , to an extent, but I belive a PR wide test suffix will behave better for your intention.

tng1-equinix: > The reason for these removals was not included in the PR description (there was none, I generated one, but that doesn't provide reason).

You mean the `ExpectNonEmptyPlan: false,`? These are not required, since it is defaulted to `false` anyways.

> It would also be helpful to see these resources move from equinix/ to the internal/fabric folder, and in that move these checks would change https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks

Yea, that is the ultimate goal, but that'll take quite a few steps, because that'll require a migration from SDKv2 to plugin framework

tng1-equinix: > Also - why are we removing the parallel testing.

Oh, this is a separate issue from the sweeper issue
https://github.com/equinix/terraform-provider-equinix/pull/1039/changes/68236331e8edca029318ad9100e0af66d5964d7d
The reason is in the commit message
`Deparallelize streams test, because concurrent streams creation can hit the 5 stream per project limit`


> I would also consider if we could change the E2E test asset prefix to contain a 3+ random/unique(enough) character suffix to denote the resources created through this test run (`tfacc_xyz_`). We have to be careful with prefix length because many resources have name length constraints and we have bumbed into these in the past.

I don't think there is much benefit for this random suffix.
From what I gather, the purpose of the suffix is to avoid name collision with other resources.
But our Fabric resources is not uniquely defined by name, there can be multiple resources with the same name, so we don't need the name to be unique.

> The regular sweeper would only remove resources of this PR (`tfacc_xyz_*`), and a general sweeper workflow would run (perhaps via GH cron) to sweep under the shared prefix (`tfacc*`).

Our side is using the `PFCR/PNFV` suffix to identify these resources.
We thought about the suggestion of using the standardized `tfacc` prefix, but that'll be a big change that requires us to change all our resource test.
For now, we're fixing all the tests before we make any big migrations

But running it in a cron job seems promising. We can update #1044 to use a cron job if that's agreeable


> If the intent is to dynamically register more resource dependencies (create a VLAN for each PR run/step, rather than reusing a shared one across all PRs) this can help with validating the overall reliability of the provider against the API.

The idea is that each test should use it's own available VLAN.
If multiple connection resource (in the acceptance test, many of them create a connection) is created on the same VLAN, it will error out

```
Error: 400 Bad Request Code: EQ-3142205, Message: This Connection either has a duplicate Z-side VLAN ID for the port or the VLAN ID is being deleted and should be available soon., Details: Given vlanSTag or vlanCTag is being used for existing connection, AdditionalInfo: [{Property: /zSide/accessPoint/linkProtocol, No Reason included in API Response}]
```

tng1-equinix: In any case, this was a "testing" PR that I created, mainly to consolidate all the changes in the other PR to verify that the (fixed) tests are passing

PR #1047, opened 2026-07-08
chore: remove unused sweeper code

tng1-equinix:
- **sweeper logic has been moved to internal/fabric, deleting this outdated file**


internal/fabric/sweep/sweep.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func IsSweepableFabricTestResource(resourceName string) bool {
+	for _, suffix := range FabricTestResourceSuffixes {
+		if strings.HasSuffix(resourceName, suffix) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
ctreatma: Are the old definitions in `internal/sweep/sweep.go` still used?  If not they should be cleaned up.  If they are, is there a reason they cannot be updated in place to implement these changes?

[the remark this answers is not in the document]
tng1-equinix: Nope, it's not being used. We have removed this file

internal/sweep/sweep.go
```diff
@@ -1,57 +0,0 @@
...
-	"strings"
-	"time"
-
-	"github.com/equinix/terraform-provider-equinix/internal/config"
-	"github.com/equinix/terraform-provider-equinix/internal/env"
-)
-
-const (
-	// duplicated from equinix_sweeoer_test.go
-	testResourcePrefix        = "tfacc"
```
displague, line 15: Following Metal's removal and pending #622, sweep.go is really unused, with the Fabric parts moved to https://github.com/equinix/terraform-provider-equinix/blob/main/internal/fabric/sweep/sweep.go#L7-L18


What I am noticing in this file and the moved file, is a dangerous pattern of deleting resources based on suffix. The prefix check (`tfacc*`) has saved the day when a developer (any user/contributor performing provider testing against their own environment) intentionally or accidentally runs the sweepers against an environment with non-test resources. 

Test resources are generally expected to match `tfacc*`. If you want to add a suffix as a secondary check, that is fine. I do wonder if that suffix check is passing for all of the Fabric test resources being created. I would imagine some test resources will not fall into one of these three categories ("_PFCR", "_PNFV", "_PPDS").

Assuming (and perhaps this is not the case today), that Fabric TF test resources are `tfacc` prefixed, I would prefer the standard prefix check over this custom suffix check. This standard prefix check can continue to be used and shared when NE is moved to internal (https://github.com/equinix/terraform-provider-equinix/pull/622#issuecomment-3632591251), and when additional resources are added.

internal/sweep/sweep.go
```diff
@@ -1,57 +0,0 @@
-package sweep
```
ctreatma, line 1: Sweeper client config and standards for identifying test resources is really a thing where we should be consistent across products rather than defining separate, product-specific baselines.

internal/sweep/sweep.go
```diff
@@ -1,57 +0,0 @@
...
-func IsSweepableFabricTestResource(resourceName string) bool {
-	for _, suffix := range FabricTestResourceSuffixes {
-		if strings.HasSuffix(resourceName, suffix) {
-			return true
-		}
-	}
-	return false
-}
-
-func GetConfigForFabric() (*config.Config, error) {
```
ctreatma, line 37: I see nothing Fabric-specific in here.  Client ID and client secret are the same things used for Network Edge, for example.  I suspect this name is what it is because `GetConfigForMetal` truly used a different configuration for Metal's non-standard auth, and Fabric just happened to be the other product that was implementing sweepers.

I would prefer for this to become `GetConfig()` and to be used for any sweeper that is making Equinix API calls.

internal/sweep/sweep.go
displague: While this is all dead-code today, the pattern and intention of this sweeper file is to be used with additional resources added to the provider. Only the Fabric definitions, redundant to internal/fabric/sweep and internal/fabric/config, should be removed from this file.

internal/sweep/sweep.go
```diff
@@ -1,57 +0,0 @@
...
-	cannotConvertTimeoutToInt = "cannot convert value of '%s' env variable to int"
-	missingFabricSecrets      = "missing fabric clientId - %s, and clientSecret - %s"
-)
-
-var (
-	FabricTestResourceSuffixes = []string{"_PFCR", "_PNFV", "_PPDS"}
-)
-
-func IsSweepableTestResource(namePrefix string) bool {
-	return strings.HasPrefix(namePrefix, testResourcePrefix)
```
ctreatma, line 25: Echoing Marques's comment above, using a consistent, unlikely-to-be-accidental prefix for test resource names is more reliable and less risky than using suffixes that might reasonably be used for non-test resources.

From what I recall, part of the reason for introducing suffixes was to prevent cross-talk between sweeper steps for the different acceptance test jobs.  It would be preferable to have a single sweeper job that runs after the parallel test jobs complete, and sweeps all Terraform test resources by prefix.

displague: To avoid a sweeper knocking out the resources of another test in progress, https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency may be of interest.

ctreatma: My understanding is that concurrency would be functionally equivalent to running the test suites in series in a single workflow.  If the goal is to run tests in parallel but avoid cross-talk between sweepers, that would likely involve a [matrix job](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#adding-a-matrix-strategy-to-your-workflow-job) to run the test suites and a sweeper job that depends on ([`needs`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#defining-prerequisite-jobs)) the matrix job.

PR #1055, opened 2026-08-17
chore: implement sweeper pipeline

tng1-equinix:
Migrate sweeper logic as pipeline from Go.

The reason is that we noticed multiple duplication of sweeper logic across repos.
We are exposing this action in this repo so other repos can refer and use it instead of re-implementing it themselves


ctreatma:
Is there a particular issue or use case that is addressed by reimplementing sweepers as shell scripts rather than using the built-in sweeper support in the plugin framework?

tng1-equinix:
> Is there a particular issue or use case that is addressed by reimplementing sweepers as shell scripts rather than using the built-in sweeper support in the plugin framework?

Hey, this is still in the discussion phase, but Panthers noticed that there is quite a lot of duplicated logic for the sweepers, between this repo, `terraform-equinix-fabric`, and now MCP and ENA.

So we thought it might make sense to have a single GHA that is reusable and referenced from those repos, instead of reimplementing and maintaining so many different sweeper logic in each of the repos

We're still not certain if this is the best repo to hold it, but we thought it is appropriate given that it is a public repo which would be able to expose this pipeline to the private repos, and that this repo is the first implementation of sweepers

tng1-equinix:
Hey @ctreatma 

Panthers have decided to pursue this solution.

The PR is now ready for review, really appreciate your help here.

ctreatma:
> there is quite a lot of duplicated logic for the sweepers, between this repo, terraform-equinix-fabric, and now MCP and ENA.

Sweepers are meant as a convenience for testing during Terraform provider development, not as a repeatable pattern to be used in real Terraform configurations.  The need to duplicate sweepers in many different places sounds less like a sweeper implementation issue and more like an issue with how those other repositories manage resources (people are making manual changes outside of terraform that prevent terraform from destroying resources) or an issue with the provider (the provider is failing to destroy resources that it should be able to destroy).  We should address these issues by improving the Terraform provider's ability to destroy the resources it manages, rather than by creating shared scripts to clean up Terraform resources outside of the Terraform lifecycle.

tng1-equinix:
> rather than by creating shared scripts to clean up Terraform resources outside of the Terraform lifecycle.

Actually, that is kinda exactly the kind of cases that sweeper is meant to handle.
From what I understand, the Terraform Acceptance test cases cleans up after itself by default.

However, there are situations where this cleanup might fail, be it Terraform level (panics in provider _etc_) or API level (ephemeral failures in DELETE API _etc_).
The sweeper would step in to identify and delete all these resources.

https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers

> Acceptance tests in Terraform provision and verify real infrastructure using [Terraform's testing framework](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests). Ideally all infrastructure created is then destroyed within the lifecycle of a test, however the reality is that there are several situations that can arise where resources created during a test are “leaked”. Leaked test resources are resources created by Terraform during a test, but Terraform either failed to destroy them as part of the test, or the test falsely reported all resources were destroyed after completing the test. Common causes are intermittent errors or failures in vendor APIs, or developer error in the resource code or test.


From my research, the Terraform Go framework for sweepers is just simply wraps the "search and destroy" functionality, which can be reproduced in `bash`.
It does not provider any other functionality from what I understand.
By re-implementing it in bash, it is more portable to the other repos, rather than requiring all the other pipeline to 1) require Go, 2) pull the Go source code.

displague:
@tng1-equinix Terraform Sweepers are the idiomatic and conventional mechanism for managing acceptance-test cleanup in Terraform providers. HashiCorp documents this pattern as part of the Terraform Plugin Testing framework, and it has been established practice across Terraform provider development for many years (https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/sweepers).

The important consideration is not simply that sweepers are conventional; it is that they establish an **upstream developer expectation**. A developer learning how to build and contribute to a Terraform provider should be able to follow the Terraform provider documentation and reasonably expect the repository to follow the patterns described there.

Introducing a separate cleanup mechanism, implemented in another language and operating outside the Terraform testing framework, creates a second model that contributors must understand and maintain. That adds onboarding and maintenance burden without changing the underlying requirement: reliably cleaning up resources created by acceptance tests.

More importantly, deviations from the established Terraform provider patterns weaken the relationship between the upstream documentation and the repository's actual development practices. Contributors should not have to learn that this provider follows a different set of conventions for fundamental provider-development concerns unless there is a concrete technical requirement that justifies the deviation.

We should  continue to use Terraform Sweepers for test cleanup and build on the upstream pattern rather than establish a provider-specific standard. If sweepers have a limitation that prevents them from meeting a particular cleanup requirement, that limitation should be identified explicitly and addressed directly. Sweepers can be run periodically if needed, outside the normal test events. Problems with the Terraform approach to sweeping tend to be indicative of problems users will experience directly when using the Terraform provider. They should be solved in a Terraform way, within the resource management.

An action that performs a similar task to Terraform sweepers, for use in different projects (Terraform Modules, Ansible configs, Pulumi modules, etc) could be managed in another repository. See https://github.com/equinix-labs/metal-sweeper-action as an example (and others: https://github.com/orgs/equinix-labs/repositories?q=action).

I'm closing this PR to prevent the removal of Terraform test harnessing. 

I recommend moving the action to another project. That action should not be called in the direct test path of this project, nor in a parallel path (GH cron) on this project.

PR #1056, opened 2026-08-17
chore: fix acceptance tests

tng1-equinix:


equinix/resource_fabric_routing_protocol_acc_test.go
```diff
@@ -42,7 +42,8 @@ func TestAccFabricCreateRoutingProtocols_PFCR(t *testing.T) {
 		CheckDestroy: checkRoutingProtocolDelete,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccFabricCreateRoutingProtocolConfig("RP_Conn_Test_PFCR", portUUID, targetVlan),
+				ExpectNonEmptyPlan: true,
```
tng1-equinix, line 45: Because we fixed a bug with the wrong diff logic in connections, now it surfaces the non-empty plans in these tests

internal/resources/fabric/servicetoken/datasources_test.go
```diff
@@ -71,7 +71,7 @@ func testAccFabricServiceTokenConfigDataSourceConfig(expiration string, virtualD
 			}
 			notifications {
     			type   = "ALL"
-    			emails = ["panthers_auto@equinix.com", "test1@equinix.com", "example@equinix.com"]
+    			emails = ["panthers_auto@equinix.com", "panthersne@equinix.com", "test1@equinix.com", "example@equinix.com"]
```
tng1-equinix, line 74: We swapped off the sunsetted account, so we need to add this new email to the list

srushti-patl (approved):

PR #1058, opened 2026-08-24
chore: fix tests

tng1-equinix:


internal/resources/fabric/precision_time/resource.go
```diff
@@ -69,7 +69,7 @@ func (r *Resource) Create(
 		return
 	}
 
-	createTimeout, diags := plan.Timeouts.Create(ctx, 10*time.Minute)
+	createTimeout, diags := plan.Timeouts.Create(ctx, 30*time.Minute)
```
tng1-equinix, line 72: EPT team says that 10 min is not sufficient to get EPT services provisioned. Extending the timeout for now, should be backwards compatible for users

internal/resources/fabric/stream_subscription/resource_test.go
```diff
@@ -37,200 +37,135 @@ func CheckStreamSubscriptionDelete(s *terraform.State) error {
 
 func testAccFabricStreamSubscriptionConfig(streamTestData map[string]map[string]string) string {
 	return fmt.Sprintf(`
-		resource "equinix_fabric_stream" "new_stream" {
```
tng1-equinix, line 40: Vision team said that they no longer have accounts for Grafana, servicenow and splunk.

They are also not testing it on their side, we can remove for now. 
We can always revert this once they obtain the new credentials, if they ever

internal/resources/fabric/streamalertrule/resource_test.go
```diff
@@ -98,33 +76,7 @@ resource "equinix_fabric_stream" "new_stream" {
   }
 }
 
-resource "equinix_fabric_stream_subscription" "SPLUNK" {
```
tng1-equinix, line 101: Not sure why this is here, this is a stream alert test so there's no need to create a subscription

srushti-patl (approved):

PR #1079, opened 2026-09-14
feat: Cxf 130433 ip block eia datasources and acceptance tests

tutkat:
Screenshot 2026-09-15 at 08 13 23


internal/resources/fabric/internet_access/api_helpers.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/equinix/equinix-sdk-go/services/fabricv4"
+)
+
+// searchInternetAccessServicesRaw makes the POST /fabric/v4/internetAccessServices/search call
+// without going through the SDK's strict UnmarshalJSON, which requires "order" to be present in
+// every InternetAccessService item even though the API does not always return it.
```
tng1-equinix, line 16: should we raise this to IA team to update their schema if this is the case?

tutkat: hmm maybe I shouldn't do that for this endpoint. I heard it will be update in the next release (endpoint), btw it doesn't make any issue I guess

tng1-equinix (approved):

internal/resources/fabric/ip_block/datasource_schema.go
```diff
@@ -0,0 +1,254 @@
...
+					},
+				},
+			},
+		},
+		"change": schema.SingleNestedAttribute{
+			Description: "Current state of the latest change on the IP block",
+			Computed:    true,
+			CustomType:  fwtypes.NewObjectTypeOf[ChangeModel](ctx),
+			Attributes: map[string]schema.Attribute{
+				"href": schema.StringAttribute{
```
tng1-equinix: Does Terraform consumers need this?

I'm not sure how Terraform users can use this href

